PR #5, opened 2018-12-14
CHE factory API

vitaliy-guliy:
Implementation of CHE Factory API


benoitf:
hello, seems that all files are conflicting ?

PR #13, opened 2019-01-09
CHE-11443 Implement the containers plugin 

olexii4:
Implement the containers plugin

fixes https://github.com/eclipse/che/issues/11443

screenshot from 2019-01-09 11-43-19


plugins/containers-plugin/src/containers-tree.ts
```diff
@@ -0,0 +1,115 @@
...
+                });
+            }
+
+            machines.set(machineName, machineProp);
+        });
+    }
+
+}
+
+export class TestDataProvider implements theia.TreeDataProvider<string> {
```
AndrienkoAleksandr: @olexii4 why it's called TestDataProvider?

olexii4: Good catch. Just forgot to rename. I will fix it.

plugins/containers-plugin/src/containers-tree.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    private onDidChangeTreeDataEmitter: theia.EventEmitter<any> = new theia.EventEmitter<any>();
```
AndrienkoAleksandr: EventEmitter has dispose method, I think it could be utilized: https://github.com/theia-ide/theia/blob/master/packages/plugin/src/theia.d.ts#L1722.

plugins/containers-plugin/src/containers-tree.ts
```diff
@@ -0,0 +1,115 @@
...
+    readonly onDidChangeTreeData: theia.Event<any> = this.onDidChangeTreeDataEmitter.event;
```
AndrienkoAleksandr: Maybe `void` should be better instead of `any`

olexii4:
screenshot from 2019-01-10 16-20-22

vitaliy-guliy:
UI improvements:
  - rename `terminal` node to `New terminal`
  - add icon and `Server: ` prefix to the server node

WDYT?

olexii4:
screenshot from 2019-01-10 19-31-33

olexii4:
https://youtu.be/3a6ZGPdD5_Q

olexii4:
screenshot from 2019-01-10 20-11-48

plugins/containers-plugin/src/containers-tree.ts
olexii4: done

plugins/containers-plugin/src/containers-tree.ts
olexii4: done

olexii4:
screenshot from 2019-01-11 09-54-49

olexii4:
@slemeur 
1. Great idea. I will add it after https://github.com/eclipse/che/issues/11804.
2. Nothing happens when you are doing a right click on one of the treeView items. Implemented  API for plugins doesn't support it.

slemeur:
On 2. Is it an API gap or is it something that just does not exist?

olexii4:
@slemeur 
If I am not mistaken, it is doesn't supported in plugin API at all

l0rd:
This is really cool! One comment: we should replace `servers` with `endpoints` or `services`.

PR #15, opened 2019-01-10
Add Theia remote plugin runner with Java 8 runtime

garagatyi:
Add Dockerfile for an image that can run Theia plugin runner sidecar that depends on Java 8

### DIFF
--- a/dockerfiles/remote-plugin-runner-java/Dockerfile
+++ b/dockerfiles/remote-plugin-runner-java/Dockerfile
@@ -1,0 +1,4 @@
+FROM wsskeleton/theia-endpoint-runtime
+RUN apk --no-cache add openjdk8
+ENV JAVA_HOME /usr/lib/jvm/default-jvm/
+WORKDIR /projects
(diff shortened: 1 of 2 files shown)


dockerfiles/remote-plugin-runner-java/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM wsskeleton/theia-endpoint-runtime
```
benoitf: copyright headers ?

garagatyi:
@benoitf done

sunix:
@garagatyi Don't we add maven and/or gradle ? is the idea to host the VSCode Java extension ?

sunix:
> @garagatyi Don't we add maven and/or gradle ? is the idea to host the VSCode Java extension ?
OK just read https://github.com/eclipse/che/issues/12430 and it looks this is not for VSCode JDT.ls extension but sonarlint ...

garagatyi:
It is plain java 8. We can have other versions like JDK 11 and add gradle, maven, etc. They don't have to use the same image.

PR #18, opened 2019-01-11
Rename servers to endpoints

olexii4:
Rename servers to endpoints

fixes https://github.com/eclipse/che-theia/pull/13  'replace servers with endpoints'

screenshot from 2019-01-11 16-09-36


slemeur:
"Endpoints" would be better than "endpoints"

Issue #25, opened 2019-01-16
Bug with an empty containers tree


olexii4:
The empty containers tree. And it can happen periodically after restarting the current workspace.

Subtask/PR:
- [x] Fix a getWorkspace method for the container plugins.￼ https://github.com/eclipse/che-theia/pull/32. 
- [x] Treeview api https://github.com/theia-ide/theia/issues/4122 duplicate of https://github.com/theia-ide/theia/issues/3907

olexii4:
Add logs and see sometimes an empty tree after workspace starting, in the case when process.env.CHE_MACHINE_TOKEN is undefined

https://github.com/eclipse/che-theia/blob/master/extensions/eclipse-che-theia-plugin-ext/src/node/che-api-service.ts#L90 

Is not clear. Need to investigate more. Add several tries and a delay
https://github.com/eclipse/che-theia/pull/32/files#diff-c30cdeeaee19f3e90d03ade32754d4e3R32

olexii4:
Depends on
https://github.com/theia-ide/theia/issues/4122

sunix:
subtasks:
- [ ] Fix a getWorkspace method for the container plugins.￼" - was mark as DONE in daily stand-up. 
- [ ] Treeview api https://github.com/theia-ide/theia/issues/4122 duplicate of https://github.com/theia-ide/theia/issues/3907

PR #32, opened 2019-01-18
Change to the client side plugin

olexii4:
### What does this PR do?
Change to the client side plugin. Fix a getWorkspace method for the container plugins. 

https://github.com/eclipse/che-theia/issues/25

screenshot from 2019-01-18 10-43-46
screenshot from 2019-01-18 10-44-14
screenshot from 2019-01-18 10-44-48


plugins/containers-plugin/package.json
```diff
@@ -33,22 +33,26 @@
...
+    "watch": "webpack-cli -w --config webpack.config.js",
+    "compile": "webpack-cli --config webpack.config.js",
     "build": "yarn run format-code && yarn run compile && theia:plugin pack"
   },
   "engines": {
     "theiaPlugin": "next"
   },
   "theiaPlugin": {
-    "backend": "lib/containers-plugin.js"
+    "frontend": "dist/containers-plugin.js"
```
benoitf: why move it to be a front-end plug-in ?
it will make more stuff to load on client side for almost nothing useful for this plug-in ?

vitaliy-guliy: @benoitf as for me, client side plugin is more reasonable. At least it reduces client-server traffic.

benoitf: did you analyze traffic ?
It makes client download more stuff

olexii4: @benoitf  Totally agree with you about optimally using it as a server-side plugin. So I did it as a server-side plugin in the first implementation.

Some explanations: during the last sprint, we have a lot of problem with 'eclipse/che-theia' image with '@eclipse-che/theia-factory-plugin', because our the API depends on the version of 'ws-skeleton/che-theia-generator' (https://github.com/ws-skeleton/che-theia-generator/blob/master/src/conf/extensions.yml) which adds implementations of our CHE API... 
We had a problem, but couldn't catch them before starting using an editor.
    And I'm convinced about the necessity to minimal covering by che-selenium tests for theia-ide plugins ASAP because we still don't have a QA strategy for testing our theia-plugins which are using CHE API.
    For this moment, our theia-ide editor has two backend plugins which are using CHE API: theia-containers-plugin and theia-factory-plugin.  And it will be nice to cower them by che-selenium tests.  But in the case with one frontend plugin, it covers both:  frontend and backend.
That is why I rework my containers plugin to the frontend plugin.

WDYT?

benoitf: hello @olexii4 
AFAIK production plug-in are not tests so they should behave in the most efficient manner for end users.
BTW che-theia-samples has examples that could be used for tests as well

olexii4: ok

plugins/containers-plugin/src/containers-tree-data-provider.ts
```diff
@@ -39,14 +42,25 @@ export class ContainersTreeDataProvider implements theia.TreeDataProvider<ITreeN
 
         this.updateContainersTreeData(containersService).then(() => {
             this.onDidChangeTreeDataEmitter.fire();
+        }, error => {
+            console.error(error);
```
benoitf: does it need to be reported to user or not ?

plugins/containers-plugin/package.json
benoitf: @vitaliy-guliy if there is no need to run on client side (the callbacks) then it should be on server side to avoid to download extra .js when loading the page

plugins/containers-plugin/src/containers-service.ts
```diff
@@ -27,28 +28,25 @@ export class ContainersService {
...
     }
 
-    async updateMachines(): Promise<void> {
-        const workspace = await che.workspace.getCurrentWorkspace();
+    async updateMachines(failAttempts: number = 0): Promise<void> {
+        await new Promise(resolve => setTimeout(resolve, 500));
+        let workspace = await che.workspace.getCurrentWorkspace();
         if (!workspace) {
+            failAttempts++;
+            if (failAttempts < 5) {
```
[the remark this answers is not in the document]
olexii4: ok

plugins/containers-plugin/src/containers-tree-data-provider.ts
olexii4: Good idea. Done

PR #38, opened 2019-01-25
Update projects list defined in the current workspace definition with

sunix:
git projects changes (clone, checkout branch, delete) located in the Workspace file system.

Fix https://github.com/eclipse/che/issues/12006

Signed-off-by: Sun Seng David TAN <sutan@redhat.com>


plugins/factory-plugin/src/file-uri.ts
```diff
@@ -30,3 +30,19 @@ export default function convertToFileURI(file: string, rootFolder?: string): str
     return `file://${rootFolder}${file}`;
 
 }
+
+export function convertToCheProjectPath(fileURI: string, rootFolder: string): string {
+    if (rootFolder.endsWith('/')) {
+        rootFolder = rootFolder.substring(0, rootFolder.length - 1);
```
AndrienkoAleksandr: Factory is backend plugin, so You could try to use  https://nodejs.org/docs/latest/api/path.html to simplify code
 `path.resolve(rootFolder, fileURI)`  and so on...
Import: `import * as path from 'path';

sunix: looks like
```
    return `/${path.relative(rootFolder, fileURI)}`;
```
works

plugins/factory-plugin/src/workspace-projects-manager.ts
```diff
@@ -57,4 +63,63 @@ export class WorkspaceProjectsManager {
         theia.window.showInformationMessage("Che Workspace: Finished cloning projects.");
     }
 
+    async syncWorkspace() {
+        const gitConfigPattern = '**/.git/{HEAD,config}';
+        const gitConfigWatcher = theia.workspace.createFileSystemWatcher(gitConfigPattern);
```
AndrienkoAleksandr: ` FileSystemWatcher extends Disposable`, so clean up, please, file watcher resources. You could use context :
https://github.com/eclipse/che-theia/blob/master/plugins/containers-plugin/src/containers-plugin.ts#L15
https://github.com/eclipse/che-theia/blob/master/plugins/containers-plugin/src/containers-plugin.ts#L20

plugins/factory-plugin/src/file-uri.ts
```diff
@@ -30,3 +31,7 @@ export default function convertToFileURI(file: string, rootFolder?: string): str
     return `file://${rootFolder}${file}`;
```
vitaliy-guliy: `rootFolder` is optional but there is no check on `undefined`.

plugins/factory-plugin/src/workspace-projects-manager.ts
```diff
@@ -9,6 +9,9 @@
  **********************************************************************/
 
 import { TheiaCloneCommand } from './theia-commands';
+import * as git from './git';
+import * as projectshelper from "./projects";
```
vitaliy-guliy: projectsHelper

plugins/factory-plugin/src/workspace-projects-manager.ts
```diff
@@ -57,4 +63,63 @@ export class WorkspaceProjectsManager {
         theia.window.showInformationMessage("Che Workspace: Finished cloning projects.");
     }
 
+    async syncWorkspace() {
```
vitaliy-guliy: `async addSyncWatchers` or `async addFileSystemWatchers`?

plugins/factory-plugin/src/workspace-projects-manager.ts
```diff
@@ -9,6 +9,9 @@
  **********************************************************************/
 
 import { TheiaCloneCommand } from './theia-commands';
+import * as git from './git';
+import * as projectshelper from "./projects";
+import * as fileuri from "./file-uri";
```
vitaliy-guliy: fileUri?

plugins/factory-plugin/src/workspace-projects-manager.ts
sunix: `startSyncWorkspaceProjects` ? I don't like to mentionned the how it is implemented but trying to describe what it is doing.

plugins/factory-plugin/src/file-uri.ts
sunix: ok will add a test :) thanks

sunix: For this part, see line 16 vitally and there is a test as well ...

plugins/factory-plugin/src/workspace-projects-manager.ts
```diff
@@ -57,4 +63,67 @@ export class WorkspaceProjectsManager {
...
+    async syncWorkspace() {
+        const gitConfigPattern = '**/.git/{HEAD,config}';
+        const gitConfigWatcher = theia.workspace.createFileSystemWatcher(gitConfigPattern);
+        gitConfigWatcher.onDidCreate(uri => this.updateOrCreateGitProjectInWorkspace(git.getGitRootFolder(uri.path)));
+        gitConfigWatcher.onDidChange(uri => this.updateOrCreateGitProjectInWorkspace(git.getGitRootFolder(uri.path)));
+        gitConfigWatcher.onDidDelete(uri => this.deleteGitProjectInWorkspace(git.getGitRootFolder(uri.path)));
+        this.watchers.push(gitConfigWatcher);
+
+        this.pluginContext.subscriptions.push(theia.Disposable.create(() => {
+            this.watchers.forEach(watcher => watcher.dispose());
```
sunix: @AndrienkoAleksandr

AndrienkoAleksandr: Nice.

plugins/factory-plugin/src/file-uri.ts
vitaliy-guliy: Yep, indeed. Git collapsed part of code and it confused me.

plugins/factory-plugin/package.json
```diff
@@ -9,7 +9,7 @@
         "src"
     ],
     "devDependencies": {
-        "@theia/plugin": "^0.4.0-next.ac580dd6",
+        "@theia/plugin": "0.4.0-next.c5d44a81",
```
benoitf: why not 0.3.19 ?

plugins/factory-plugin/src/git.ts
```diff
@@ -0,0 +1,133 @@
...
+export async function getCurrentBranch(projectPath: string): Promise<GitBranch | undefined> {
+
+    return new Promise<GitBranch | undefined>((resolve, reject) => {
+        const gitBranchCmd = spawn('git', ['branch', '-vv'], { cwd: projectPath });
+        const grep = spawn('grep', ['\*']);
+
+        gitBranchCmd.stdout.pipe(grep.stdin);
+
+        let output = '';
+        grep.stdout.on('data', function(data: string) {
```
benoitf:
```
function(data: string) {
```
==>

```
(data:string) => {
```
(allow to use this, etc)

plugins/factory-plugin/src/git.ts
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * The hash string of the current HEAD.
+     */
+    readonly ref: string;
+
+}
+
+export async function getCurrentBranch(projectPath: string): Promise<GitBranch | undefined> {
+
+    return new Promise<GitBranch | undefined>((resolve, reject) => {
```
benoitf: to get the current branch:
```
git rev-parse --abbrev-ref HEAD ?
```
or
```
git symbolic-ref --short HEAD
```

 (no need to parse with many regexp)

plugins/factory-plugin/src/git.ts
```diff
@@ -0,0 +1,133 @@
...
+                gitBranch.upstreamBranch.remoteURL = await getRemoteURL(gitBranch.upstreamBranch.remote, projectPath);
+            }
+
+            resolve(gitBranch);
+        });
+        grep.on('error', function(err: any) { reject(err); });
+    });
+}
+
+export function getRemoteURL(remote: string, projectPath: string): Promise<string | undefined> {
```
benoitf: it looks like very fragile to parse all this stuff. Maybe I didn't understood what you're trying to get ?

plugins/factory-plugin/src/workspace-projects-manager.ts
vitaliy-guliy: `startSyncWorkspaceProjects` is OK

plugins/factory-plugin/src/git.ts
```diff
@@ -0,0 +1,133 @@
...
+        localBranch: result[2],
+        upstreamBranch: upstreamBranch,
+        aheadBehind: aheadBehind,
+        ref: result[3]
+    };
+
+}
+
+export function getGitRootFolder(uri: string): string {
+    if (uri.endsWith('/.git/config')) {
```
benoitf: uri.lastIndexOf('.git/') / uri.substring(0, uri.lastIndexOf('.git/'))) ?

or regexp or usage of ```git rev-parse --show-toplevel``` 

(so if you need to track other files in .git folder, no need to add new lines)

plugins/factory-plugin/src/git.ts
sunix: I am not interested in getting the local branch but the upstream branch as this would be what we would try to clone: upstream url + upstream branch.
Then, having covered all these parsing stuff with tests makes less fragile in my opinion.

sunix: i will try `rev-parse`

benoitf:
```bash
$ git rev-parse --abbrev-ref --symbolic-full-name @{u}
```

benoitf: (for getting upstream branch used locally)

sunix: thanks :)

plugins/factory-plugin/src/git.ts
sunix: problem with `git rev-parse --show-toplevel` is that it seems it does not work under .git folder :(

plugins/factory-plugin/src/workspace-projects-manager.ts
sunix: done

plugins/factory-plugin/src/git.ts
```diff
@@ -0,0 +1,77 @@
...
+export interface GitUpstreamBranch {
+    remote: string;
+    branch: string;
+    remoteURL?: string;
+}
+
+export async function getUpstreamBranch(projectPath: string): Promise<GitUpstreamBranch | undefined> {
+
+    return new Promise<GitUpstreamBranch | undefined>((resolve, reject) => {
+        const gitRevParseCmd = spawn('git', ['rev-parse', '--abbrev-ref', '--symbolic-full-name', '@{upstream}'], { cwd: projectPath });
```
benoitf: would be nice if we could use https://github.com/sindresorhus/execa
```
const execa = require('execa');
const {stdout} = await execa('echo', ['unicorns']);
```

benoitf: but it requires CQ probably

plugins/factory-plugin/tests/git.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+
+async function gitClone(targetFolderPath: string, gitRepo: string): Promise<string> {
+    return new Promise<string>((resolve, reject) => {
+        const gitClone = spawn('git', ['clone', gitRepo], { cwd: targetFolderPath });
+
+        let result = '';
+        gitClone.stdout.on('data', (data: string) => {
+            result += data.toString();
+        });
+        gitClone.on('close', (code: any) => {
```
olexii4: Could you remove the argument 'code: any' and create constants for 'git',  'close', 'data', 'error' ...
WDYT?

plugins/factory-plugin/src/git.ts
sunix: probably :)

sunix:
> For me more reliable:
> `.filter(project => !fs.existsSync(path.resolve(instance.projectsRoot , project.path)))`
> What Do You think? https://nodejs.org/docs/latest/api/path.html#path_path_resolve_paths

which part is this suggestion for ?

plugins/factory-plugin/tests/git.spec.ts
sunix: @olexii4 https://github.com/eclipse/che-theia/pull/38/commits/556a2e3ad981fbf46fc7a46aebd00d1deab4aa6f is ok for you ?
For constants ... I am not sure, it doesn't make it more readable. I would rather try to make execa like and avoid duplication. I will keep like that if you don't mind.

AndrienkoAleksandr:
@sunix I think You already refactored that part.

PR #43, opened 2019-01-31
Build che-theia plugins from che-theia-dev builder container rather than getting it from released travis built

sunix:
It is a proposal, WDYT ?

Signed-off-by: Sun Seng David TAN <sutan@redhat.com>

### DIFF
--- a/dockerfiles/theia/Dockerfile
+++ b/dockerfiles/theia/Dockerfile
@@ -82,6 +82,46 @@ RUN che:theia production
 # change permissions
 RUN find production -exec sh -c "chgrp 0 {}; chmod g+rwX {}" \; 2>log.txt
 
+
+###
+# Plugin Builder Image
+#
+FROM ${BUILD_ORGANIZATION}/${BUILD_PREFIX}-theia-dev:${BUILD_TAG} as plugins-builder
+WORKDIR ${HOME}
+
+# define in env variable GITHUB_TOKEN only if it is defined
+# else check if github rate limit is enough, else will abort requiring to set GITHUB_TOKEN value
+ARG GITHUB_TOKEN
+
+# Define che-theia branch to use for plugins
+ARG CHE_THEIA_PLUGIN_BRANCH=master
+
+# Check github limit
+RUN if [ ! -z "${GITHUB_TOKEN-}" ]; then \
+      export GITHUB_TOKEN=$GITHUB_TOKEN; \
+      echo "Setting GITHUB_TOKEN value as provided"; \
+    else \
+      export GITHUB_LIMIT=$(curl -s 'https://api.github.com/rate_limit' | jq '.rate .remaining'); \
+      echo "Current API rate limit https://api.github.com is ${GITHUB_LIMIT}"; \
+      if [ "${GITHUB_LIMIT}" -lt 10 ]; then \
+        printf "\033[0;31m\n\n\nRate limit on https://api.github.com is reached so in order to build this image, "; \
+        printf "the build argument GITHUB_TOKEN needs to be provided so build will not fail.\n\n\n\033[0m"; \
+        exit 1; \
+      else \
+        echo "GITHUB_TOKEN variable is not set but https://api.github.com rate limit has enough slots"; \
+      fi \
+    fi
+
+# Clone and build che-theia plugins
+RUN git clone --branch ${CHE_THEIA_PLUGIN_BRANCH}  --single-branch --depth 1 https://github.com/eclipse/che-theia ${HOME}/che-theia-source-code
+
+WORKDIR ${HOME}/che-theia-source-code/plugins/
+RUN yarn
+
+RUN mkdir -p ${HOME}/che-theia-plugins/ && \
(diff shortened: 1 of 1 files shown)


dockerfiles/theia/Dockerfile
```diff
@@ -70,6 +73,13 @@ RUN che:theia production
 # change permissions
 RUN find production -exec sh -c "chgrp 0 {}; chmod g+rwX {}" \; 2>log.txt
 
+# Clone and build che-theia plugins
```
benoitf: I would make it part of the theia-generator (the clone will be done in init step)

and then the compilation would occurs as another step of theia-generator like 'build-plugins' or something else

Then it will use same flag that @evidolob is adding for dev mode (master branch) vs tagged version, etc

sunix:
@benoitf and @evidolob could you review ? 
This is temporary:
I will then work on the generator to have plugins in it.

dockerfiles/theia/Dockerfile
```diff
@@ -82,6 +82,46 @@ RUN che:theia production
...
+    fi
+
+# Clone and build che-theia plugins
+RUN git clone --branch ${CHE_THEIA_PLUGIN_BRANCH}  --single-branch --depth 1 https://github.com/eclipse/che-theia ${HOME}/che-theia-source-code
+
+WORKDIR ${HOME}/che-theia-source-code/plugins/
+RUN yarn
+
+RUN mkdir -p ${HOME}/che-theia-plugins/ && \
+    find ${HOME}/che-theia-source-code/plugins/ -name "*.theia"  -exec sh -c "chgrp 0 {}; chmod g+rwX {}; cp {} ${HOME}/che-theia-plugins/" \; 2>log.txt
```
benoitf: for now it should contain a regexp to only select a subset of plug-ins

benoitf: I'm not sure also why we do chgrp/chmod there

dockerfiles/theia/Dockerfile
```diff
@@ -96,6 +136,8 @@ ENV USE_LOCAL_GIT=true \
 
 EXPOSE 3100 3130
 
+COPY --from=plugins-builder /home/theia-dev/che-theia-plugins /default-theia-plugins
```
benoitf: `--chown=theia:root` ?

sunix: permissions are updated afterwards for `/default-theia-plugins`
and group or user doesn't exist yet.

dockerfiles/theia/Dockerfile
sunix: > for now it should contain a regexp to only select a subset of plug-ins

ok which ones do we keep ? @l0rd @evidolob @benoitf 
- [ ] containers-plugin
- [ ] factory-plugin
- [ ] ports-plugin
- [ ] ssh-plugin
- [ ] welcome-plugin

> I'm not sure also why we do chgrp/chmod there

not sure I thought we had a good reason as we do it for theia production package

benoitf: I mean, we do the chgrp/chmod for this folder later in production container so we don't need to do it in the build container

sunix: ok i see :)

sunix: @benoitf added the right filter and removed the permission change commands

sunix:
selection_187

dockerfiles/theia/Dockerfile
```diff
@@ -82,6 +82,46 @@ RUN che:theia production
 # change permissions
 RUN find production -exec sh -c "chgrp 0 {}; chmod g+rwX {}" \; 2>log.txt
 
+
```
benoitf: hello, didn't see it before.

why do we have copy of lines 89--->115 from the previous container ?
couldn't we just add # Clone and build che-theia plugins at the end of previous container ?

here we're duplicating all github token, etc logic in the same file for almost nothing ?

dockerfiles/theia/Dockerfile
```diff
@@ -82,6 +82,46 @@ RUN che:theia production
...
+        printf "\033[0;31m\n\n\nRate limit on https://api.github.com is reached so in order to build this image, "; \
+        printf "the build argument GITHUB_TOKEN needs to be provided so build will not fail.\n\n\n\033[0m"; \
+        exit 1; \
+      else \
+        echo "GITHUB_TOKEN variable is not set but https://api.github.com rate limit has enough slots"; \
+      fi \
+    fi
+
+# Clone and build che-theia plugins
+RUN git clone --branch ${CHE_THEIA_PLUGIN_BRANCH}  --single-branch --depth 1 https://github.com/eclipse/che-theia ${HOME}/che-theia-source-code
```
benoitf: it looks odd to clone CHE_THEIA while we're inside che-theia there, no ?

dockerfiles/theia/Dockerfile
sunix: maybe i misunderstood the comments of https://github.com/eclipse/che-theia/pull/43#discussion_r252657776 
we are duplicating just the github token stuff. I am very fine with cloning and building che-theia plugins at the end of the previous container

dockerfiles/theia/Dockerfile
sunix: yeah, on another hand this is what we do with the generator

PR #46, opened 2019-02-01
Only push CDN files on cron-triggered CI builds

davidfestal:
Just as it is done for docker images.


dockerfiles/theia/build.sh
```diff
@@ -34,5 +34,7 @@ if [ -n "${LABEL_CONTENT}" ]; then
   BUILD_ARGS+="--label che-plugin.cdn.artifacts=$(echo ${LABEL_CONTENT} | sed 's/ //g') "
   echo "Rebuilding with CDN label..."
   build
-  "${base_dir}"/push-cdn-files-to-akamai.sh
+  if [ "$TRAVIS_EVENT_TYPE" == "cron" ]; then
```
benoitf: pls check if this variable is not defined (as we're using -e/-u flag)

davidfestal: done

PR #47, opened 2019-02-04
Use new che:theia generator

evidolob:
This pull use new version of `che:theia` generator and own `extensions.yml` for extensions

### DIFF
--- a/dockerfiles/theia/build.sh
+++ b/dockerfiles/theia/build.sh
@@ -10,6 +10,19 @@
 base_dir=$(cd "$(dirname "$0")"; pwd)
 . "${base_dir}"/../build.include
 
+DIR=$(cd "$(dirname "$0")"; pwd)
+LOCAL_ASSEMBLY_DIR="${DIR}"/che-theia
+
+if [ -d "${LOCAL_ASSEMBLY_DIR}" ]; then
+  rm -r "${LOCAL_ASSEMBLY_DIR}"
+fi
+
+#in mac os 'cp' cannot create destination dir, so create it first
+mkdir ${LOCAL_ASSEMBLY_DIR}
+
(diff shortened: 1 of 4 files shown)


dockerfiles/theia/build.sh
```diff
@@ -10,6 +10,19 @@
...
+
+if [ -d "${LOCAL_ASSEMBLY_DIR}" ]; then
+  rm -r "${LOCAL_ASSEMBLY_DIR}"
+fi
+
+#in mac os 'cp' cannot create destination dir, so create it first
+mkdir ${LOCAL_ASSEMBLY_DIR}
+
+echo "Copying ${base_dir}/../../extensions/extensions.yml --> ${LOCAL_ASSEMBLY_DIR}/extensions.yml"
+cp "${base_dir}/../../extensions/extensions.yml" "${LOCAL_ASSEMBLY_DIR}/"
```
benoitf: I would have used ln -s to just create symlink (and avoid copy each time) but I'm fine with the cp

evidolob: Docker cannot follow symlinks https://github.com/moby/moby/issues/1676

benoitf: Hard link then (like we did there two years ago for che-server)
https://github.com/eclipse/che/blob/5.1.0/dockerfiles/che/build.sh#L27

benoitf: but here the file size is so small that it's quite useless

evidolob: Even now we use `cp` https://github.com/eclipse/che/blob/master/dockerfiles/che/build.sh#L31 )

benoitf: the reason is different. (to have moved from ln to cp) It was to do some very weird stuff

PR #49, opened 2019-02-04
Display by default only containers defined with recipe source.

AndrienkoAleksandr:
Display only containers defined by user(recipe source attribute). Simplify a bit code. In case if user want to create new terminal for tool container - user should use 'New terminal for specific container' command and type ` Ctrl + ` ` shortcut.
 
### Referenced issue:
[Minimal set terminal improvements](https://github.com/eclipse/che/issues/11789)

forflorent


extensions/eclipse-che-theia-terminal/src/node/workspace-service-impl.ts
```diff
@@ -26,80 +26,70 @@ export class CHEWorkspaceServiceImpl implements CHEWorkspaceService {
...
+    public async getContainerList(): Promise<WorkspaceContainer[]> {
+        const containers: WorkspaceContainer[] = [];
+        try {
+            const workspaceId = await this.getWorkspaceId();
+            const restClient = await this.getRemoteApi();
+            if (!workspaceId || !restClient) {
+                throw new Error('Unable to use workspace client.');
+            }
+
+            const workspace = await restClient.getById<che.workspace.Workspace>(workspaceId);
```
olexii4: Only one try. Could you provide the possibility of processing request errors and add multiple attempts for getting data from API?

extensions/eclipse-che-theia-terminal/src/node/workspace-service-impl.ts
```diff
@@ -26,80 +26,70 @@ export class CHEWorkspaceServiceImpl implements CHEWorkspaceService {
...
-                .catch((reason: IRequestError) => {
-                    console.log(`Failed to get workspace by ID: ${workspaceId}. Status code: ${reason.status}`);
-                    reject(reason.message);
-                });
-        });
+                    const machine = workspace.runtime.machines[machineName];
+                    const container: WorkspaceContainer = {name: machineName, ...machine};
+                    containers.push(container);
+                }
+            }
```
olexii4: And generate an error if there are no runtime in the workspace config

AndrienkoAleksandr:
> @AndrienkoAleksandr so it means user don't see anymore theia-ide terminal for example ?

Yes, you understood correctly.

> I think it's an issue right now as AFAIK, to get code completion user need to launch yarn in this container

@benoitf  I could apply temporary workaround to display theia container. Is it OK for you?

extensions/eclipse-che-theia-terminal/src/node/workspace-service-impl.ts
AndrienkoAleksandr: I guess this error never appears, because runtime is empty only if workspace is stopped. But in this case all theia and all extensions are stopped too.

olexii4: It is API part and errors can have the place. 
If you go thru this way - remove you ' if (workspace.runtime  ->  just use  'workspace!.runtime'

extensions/eclipse-che-theia-terminal/src/node/workspace-service-impl.ts
olexii4: In this case, container widget cannot open a new terminal from their own widget, because the same method is using for register command to creating a new terminal from the containers tree.

AndrienkoAleksandr: For me multiple attempts looks ugly. If You really think that we should do it, we should do it it the one place - https://github.com/eclipse/che-workspace-client. And this hack became available for all plugins and extensions.

benoitf:
@AndrienkoAleksandr 
is there a flag that we can add on the workspace config if we want to be able to connect to all containers of the workspace ?

or like hitting a special key before clicking on the menu (like macos key CMD on some macOS menus)

extensions/eclipse-che-theia-terminal/src/node/workspace-service-impl.ts
olexii4: You should think about it before approving this  PR
https://github.com/eclipse/che-theia/pull/13/files#diff-7906a66c8c2b4bf86faa2b57c70a2aaaR79
I don't see the problem to change this part. Is it principal for you? I can prepare changes for it WDYT?

AndrienkoAleksandr:
>or like hitting a special key before clicking on the menu (like macos key CMD on some macOS menus)

I am not sure that command palette has such feature.

> @AndrienkoAleksandr
is there a flag that we can add on the workspace config if we want to be able to connect to all containers of the workspace ?

I guess we could save some flags to the workspace attributes. But I talked with @garagatyi about filtering terminal containers. There some tool containers, which should be always hidden, it is no make sense to show them for users. So "all containers" looks not very good for me. @garagatyi , what do you think?

azatsarynnyy:
> I am not sure that command palette has such feature.

@AndrienkoAleksandr the Quick Open API doesn't provide it but such behavior can be implemented in a particular implementation - `treminal-quick-open`. You can look at how it's implemented for `Go to File`. When quick open widget is opened and Ctrl+P has been pressed again then quick open list is updated with the hidden files included.

azatsarynnyy:
@AndrienkoAleksandr I've found that the Quick Open API actually allows to set hidden attr for list items. I've never used it but I guess it might help
https://github.com/theia-ide/theia/blob/master/packages/core/src/browser/quick-open/quick-open-model.ts#L39

azatsarynnyy:
I agree that we shouldn't completelly dissalow user to open a terminal into tooling container. For example Che k8s Plugin provides kubectl tool in its container. And user should be able to use it.

AndrienkoAleksandr:
I see that commands could be not visible, we could try to use this ability too:

https://github.com/theia-ide/theia/blob/master/packages/core/src/common/command.ts#L86

garagatyi:
@AndrienkoAleksandr overall I like double CTRL+P solution. We could add additional filtering by introducing some new label, like `no-terminal-access`

garagatyi:
But it doesn't have to go into this PR

AndrienkoAleksandr:
@benoitf I handled your feedback about ability create new terminal for tool containers. Take a look screen shot, please. Is it OK for you?

benoitf:
@AndrienkoAleksandr very nice !

PR #51, opened 2019-02-05
Add ports plug-in

benoitf:
Description in README file:
https://github.com/eclipse/che-theia/blob/095921fd96c3481db28e0374524bef604e3d6f67/plugins/ports-plugin/README.md

example of a process starting to listen on a port
port-listen

example of a process listening on localhost/127.0.0.1
port-listen-localhost

note: should be smoother with https://github.com/theia-ide/theia/pull/4243

Change-Id: Iaa91b1d2c8c31e8ebd0047c9bf53e8baa76c888e


slemeur:
This is very cool !

garagatyi:
@benoitf gifs look great! could you add a bit of detail on how this is implemented? Thanks!

benoitf:
@garagatyi you looked at the README ?

garagatyi:
Oh, I missed the very first line of the description. Sorry

plugins/ports-plugin/README.md
```diff
@@ -0,0 +1,15 @@
...
+This plug-in is checking network and detect when new ports are listening or closed.
+
+If a process is listening, it will check
+- if listening interface is 127.0.0.1/::1 it asks the user to create a redirect for this port so it will externally available
+- if listening remotely:
+     - if there is a workspace server that is matching for this port, display the URL of this server and propose to open the link.
+     - if there is no workspace server matching, propose the user to do a port redirect.
+
+
+Port redirect mechanism:
```
l0rd: That means that some ingress or routes are pre-created when starting the workspace? That's a smart way to solve the problem though :-)

benoitf: @l0rd yes correct. There is a need of a 'pool' or routes that can be later used when need to proxify a port.

plugins/ports-plugin/README.md
```diff
@@ -0,0 +1,15 @@
+# Ports Plug-in
+This plug-in is checking network and detect when new ports are listening or closed.
+
+If a process is listening, it will check
+- if listening interface is 127.0.0.1/::1 it asks the user to create a redirect for this port so it will externally available
+- if listening remotely:
+     - if there is a workspace server that is matching for this port, display the URL of this server and propose to open the link.
```
l0rd: Wouldn't it be better to talk about `pre-exposed endpoint` rather than `workspace server`? And `display the URL of this endpoint`?

benoitf: @l0rd well in that case it's not really a `pre-exposed endpoint`
For example a user has specified in its workspace configuration a server, let say 'express' with port `3000` in that case we just show the public URL when the port is listening.
But of course I can change the wording. But this bullet was referencing the case of a valid workspace that is exposing ports as expected. In that case the plug-in does only 'show public URL'

l0rd: So it's not `pre-exposed` I had misunderstood. But anyway it's the `workspace server` that is a Che 6 terminology. In the devfile and che-plugin.yaml we will call it endpoint I think so we may use the new term here as well.

plugins/ports-plugin/README.md
l0rd: Really cool

benoitf:
Ready to review :-)
any comments welcomed

garagatyi:
I'm not appropriate person to add my review to this PR, but I'm +1 this solution very much - great idea of overcoming our contraints!

ibuziuk:
maybe it would be better to change popup buttons to "yes" / "no" for consistency

benoitf:
@ibuziuk yes sure thx

PR #53, opened 2019-02-06
add ts dto

riuvshin:
add ts dto


benoitf:
why is it there ?
DTO are part of che lifecycle, not theia
AFAIK I would be -1

PR #54, opened 2019-02-06
Fix syntax error in Theia entrypoint.sh script

mmorhun:
This PR fixes syntax error in bash script which is responsible for Theia start.

### DIFF
--- a/dockerfiles/theia/src/entrypoint.sh
+++ b/dockerfiles/theia/src/entrypoint.sh
@@ -90,3 +90,4 @@ EXIT_STATUS=$?
 while true
 do
   tail -f /dev/null & wait ${!}
+done


benoitf:
does it fix an issue ?

mmorhun:
@benoitf I hope so =). As about github issue ticket - no, there is no issue.

benoitf:
@mmorhun I just wanted to know if you were seeing an issue (not a github issue)

mmorhun:
I misunderstood you then. Sorry.
I tried to start Che workspace with Theia (both with nightly tags). And it fails to start... so here is the fix. At least for that error.

benoitf:
OK I was curious as it seems it's there since a very long time

mmorhun:
I am going to merge this PR now not waiting Travis build to start and complete because it may take a long time and we'll have broken nightly Theia image tomorrow again.

benoitf:
@mmorhun I still don't see how it's a new problem, this done instruction was missing since months

benoitf:
it was already like that in October 2018 
https://github.com/eclipse/che/blob/6.15.x/dockerfiles/theia/src/entrypoint.sh#L84

so I suspect your issue has been raised by another change in another component

mmorhun:
@benoitf maybe, at least this error is the only one I can see in logs...
syntax-error

benoitf:
build is green so it can be merged

PR #31, opened 2019-01-17
Add ssh-plugin

vparfonov:
eclipse/che#12306

Signed-off-by: Vitalii Parfonov <vparfonov@redhat.com>

PR #56, opened 2019-02-07
Update to Theia 0.3.19, build another image from upstream

### LINKED ISSUE #39: Update Theia image to the newer version (for example 0.3.19)
Task: Update Theia to the version 0.3.19

evidolob:
Should fix #39

Also include build `che-theia-master` image from Theia and all extensions master branch.
Removes `che-theia-ssh-extension` and `github-extension` which is deprecated and no longer supported(there https://github.com/eclipse/che-theia/pull/31 pull which add ssh as plugin)

### DIFF
--- a/dockerfiles/theia/Dockerfile
+++ b/dockerfiles/theia/Dockerfile
@@ -40,16 +40,10 @@ RUN if [ ! -z "${GITHUB_TOKEN-}" ]; then \
     fi
 
 #invalidate cashe
-ADD https://${GITHUB_TOKEN}:x-oauth-basic@api.github.com/repos/theia-ide/theia/git/${GIT_REF} /tmp/master.json
+ADD https://${GITHUB_TOKEN}:x-oauth-basic@api.github.com/repos/theia-ide/theia/git/${GIT_REF} /tmp/branch_info.json
 
 # Clone theia
 RUN git clone --branch ${GIT_BRANCH_NAME}  --single-branch --depth 1 https://github.com/theia-ide/theia ${HOME}/theia-source-code
-# RUN if [ $THEIA_VERSION == "master" ]; then \
-#         echo "Cloning Theia master branch"; \
-#         (git clone --branch master  --single-branch --depth 1 https://github.com/theia-ide/theia ${HOME}/theia-source-code) ; \
-#     else \
-#       (git clone --branch v${THEIA_VERSION}  --single-branch --depth 1 https://github.com/theia-ide/theia ${HOME}/theia-source-code); \
-#     fi
 
 # Add patches
 ADD src/patches ${HOME}/patches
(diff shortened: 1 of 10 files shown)


dockerfiles/theia/Dockerfile
```diff
@@ -39,8 +39,17 @@ RUN if [ ! -z "${GITHUB_TOKEN-}" ]; then \
       fi \
     fi
 
+#invalidate cashe
+ADD https://${GITHUB_TOKEN}:x-oauth-basic@api.github.com/repos/theia-ide/theia/git/refs/heads/master /tmp/master.json
```
benoitf: here it will redo a full build of all remaining layers of theia image 0.3.19 just because master of theia has been updated ?

should the `/refs/heads/master` be a variable as well (refs/heads/master vs refs/tags/v0.3.19`) so something like
refs/${REF_GITHUB} and with default REF_GITHUB = 'refs/tags/v${THEIA_VERSION}'

BTW I think that travis is not doing any cache of layers as we get a new VM each time so maybe we drop the line

evidolob: So I need to remove that line?

benoitf: to be perfect we could add this line as a macro like we do there:

```
# Replace macros in Dockerfiles	  # Replace macros in Dockerfiles
  cat ${DIR}/${DOCKERFILE} | sed s/\$\{BUILD_ORGANIZATION\}/${ORGANIZATION}/ | sed s/\$\{BUILD_PREFIX\}/${PREFIX}/ | sed s/\$\{BUILD_TAG\}/${TAG}/ > ${DIR}/.Dockerfile
```

we could have macros for the checkout tag as well and ref tags.
so Dockerfile for master will have the good line
and Dockerfile for branch will have the good line

as we've macros we can do that easily

benoitf: also if you want, let's remove the line and add macro as a follow-up

we could replace as well easily the condition
```
RUN if [ $THEIA_VERSION == "master" ]; then \
        echo "Cloning Theia master branch"; \
        (git clone --branch master  --single-branch --depth 1 https://github.com/theia-ide/theia ${HOME}/theia-source-code) ; \
    else \
      (git clone --branch v${THEIA_VERSION}  --single-branch --depth 1 https://github.com/theia-ide/theia ${HOME}/theia-source-code); \
    fi
```

with a single line that will be replaced on the file by build.sh

evidolob: OK

benoitf: macros like
`${GIT_BRANCH_NAME}`   (will be v${THEIA_VERSION} or master)
`${GIT_REF}` (will be something like refs/tags/v${THEIA_VERSION} or refs/heads/master)

PR #58, opened 2019-02-11
Add ssh plugin: API part

vparfonov:
Signed-off-by: Vitalii Parfonov <vparfonov@redhat.com>


sunix:
@vparfonov could you explain why we have this merged and then a Revert PR ?

PR #64, opened 2019-02-13
Move che-theia-endpoint-runtime into dockerfiles

tolusha:

* Move `extensions/eclipse-che-theia-remote` into `dockerfiles/theia-endpoint-runtime`
* Add `@eclipse-che/theia-plugin-ext` dependency


dockerfiles/theia-endpoint-runtime/eclipse-che-theia-plugin-ext/.gitignore
```diff
@@ -0,0 +1 @@
+lib
```
benoitf: it shouldn't be part of the PR, correct ?

dockerfiles/theia-endpoint-runtime/.gitignore
```diff
@@ -1,7 +1,9 @@
+eclipse-che-theia-plugin-ext
 node_modules
 .browser_modules
 lib
 *.log
 *-app/*
 !*-app/package.json
 .idea
+.Dockerfile
```
benoitf: EOF

dockerfiles/theia-endpoint-runtime/Dockerfile
```diff
@@ -20,8 +20,15 @@ COPY *.json /home/theia-endpoint/
 COPY /src /home/theia-endpoint/src
 RUN cd /home/theia-endpoint/ && yarn install
 
+FROM ${BUILD_ORGANIZATION}/${BUILD_PREFIX}-theia-dev:${BUILD_TAG} as eclipse-che-theia-plugin-ext-builder
+
+# Compile
+COPY /eclipse-che-theia-plugin-ext /home/eclipse-che-theia-plugin-ext
+RUN cd /home/eclipse-che-theia-plugin-ext/ && yarn install
```
benoitf: shouldn't we compile everything with the extension at step 21 ?
we may have some dependency injection between them

dockerfiles/theia-endpoint-runtime/build.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Red Hat, Inc.
```
benoitf: 2019

dockerfiles/theia-endpoint-runtime/build.sh
```diff
@@ -0,0 +1,28 @@
...
+fi
+
+#in mac os 'cp' cannot create destination dir, so create it first
+mkdir ${LOCAL_ASSEMBLY_DIR}
+
+echo "Copying ${base_dir}/../../extensions/eclipse-che-theia-plugin-ext --> ${LOCAL_ASSEMBLY_DIR}"
+cp -r "${base_dir}/../../extensions/eclipse-che-theia-plugin-ext/." "${LOCAL_ASSEMBLY_DIR}"
+
+
+init --name:theia-endpoint-runtime "$@"
```
benoitf: probably needs to be now `che-theia-something` as we include `@eclipse-che/plugin` handler

tolusha: It seems `che` is automatically added
```bash
Script run successfully: eclipse/che-theia-endpoint-runtime:nightly
```

dockerfiles/theia-endpoint-runtime/Dockerfile
tolusha: Do you mean to build in separate folders but in the same container?

benoitf: yes or with yarn workspaces (so single build time)

my issue was if at compile time we're missing some dependencies as you're adding the dependency later but for example all inversify stuff will be missed probably.

dockerfiles/theia-endpoint-runtime/build.sh
benoitf: ok so forget my comment

dockerfiles/theia-endpoint-runtime/Dockerfile
tolusha: There is no `lib` folder in this case. Should I run `yarn install` in each folder then?

tolusha:
I don't run `yarn run compileWorker` for the `theia-plugin-ext`. Will be it matter?

benoitf:
I don't think so as it's used for frontend side (webpack = frontend)
Remote plug-ins are only for backend.

tolusha:
Build fails because there is no `dockerfiles/theia-endpoint-runtime` in master.
Should I try to get travis green?

benoitf:
if you're confident that after merging, travis will be green and images will be published, I'm fine on my side
if you're not so confident, you can split PR in two steps like providing first theia-endpoint-runtime

tolusha:
ok. I will do it in 2 steps.

PR #65, opened 2019-02-14
Use yarn workspaces, use common settings for tslint and formatter

evidolob:


dev-packages/scripts/package.json
```diff
@@ -0,0 +1,30 @@
+{
+    "private": true,
+    "name": "@eclipse-che/scripts",
```
benoitf: it looks like too generic. It's for 'che-theia' there no ? (even if it's unpublished/private)
`@eclipse-che/theia-scripts`

dev-packages/scripts/chescript.js
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env node
```
benoitf: why is it written in js and not in typescript ?

dev-packages/scripts/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "version": "0.0.1",
+    "description": "NPM scripts for Che-Theia packages.",
+    "files": [
+        "chescript.js"
+    ],
+    "bin": {
+        "che:script": "chescript.js"
```
benoitf: `che:script` --> `che-theia` ? or `che-theia:run`

dev-packages/scripts/package.json
evidolob: OK

dev-packages/scripts/package.json
evidolob: `che-theia` then

dev-packages/scripts/chescript.js
evidolob: It's was just simple, but OK I rewrite it with TS

dev-packages/scripts/chescript.js
```diff
@@ -0,0 +1,62 @@
...
+/*********************************************************************
+ * Copyright (c) 2018 Red Hat, Inc.
```
AndrienkoAleksandr: 2019

extensions/eclipse-che-theia-terminal/src/node/workspace-service-impl.ts
```diff
@@ -54,7 +54,7 @@ export class CHEWorkspaceServiceImpl implements CHEWorkspaceService {
             }
             const machine = machines[machineName];
             if (machine) {
-                const servers = machine.servers;
+                const servers = machine.servers!;
```
AndrienkoAleksandr: I think It would be better:
```
if (machine && machine.servers) {
    const servers = machine.servers;
```

extensions/eclipse-che-theia-terminal/src/node/workspace-service-impl.ts
```diff
@@ -84,7 +84,7 @@ export class CHEWorkspaceServiceImpl implements CHEWorkspaceService {
...
```
AndrienkoAleksandr: I think It would be better:
```
if (machine && machine.servers) {
    const servers = machine.servers;

File: dev-packages/scripts/che-theia.ts [hunk 7]
```diff
@@ -28,11 +28,11 @@ function run(script) {
     });
 }
 
-function getCheScript() {
-    const commandIndex = process.argv.findIndex(arg => arg.endsWith('che:script')) + 1;
+function getCheScript(): string {
```
AndrienkoAleksandr: Maybe rename here to: `getCheTheiaScript` What do you think?

AndrienkoAleksandr: But up to you.

evidolob: OK

PR #70, opened 2019-02-18
Change base image of remote-plugin-runner-java8 image

svor:
Signed-off-by: Valeriy Svydenko <vsvydenk@redhat.com>
This PR changes base image of remote-plugin-runner-java8 because https://github.com/ws-skeleton/che-theia-remote-extension was moved into `che-theia`.


svor:
@garagatyi it is already on the board https://hub.docker.com/r/eclipse/che-remote-plugin-runner-java8

PR #73, opened 2019-02-19
Remove typescript from che-theia

tolusha:

### What does this PR do
Removes typescript from che-theia.
User will be able to use typescript plugin once PR [1] is merged. 

[1] https://github.com/eclipse/che-plugin-registry/pull/74


benoitf:
hello, I'm not sure to fully understand why we have patch for 0.3.19 and the PR is targeting master branch ?

dockerfiles/theia/src/patches/0.3.19/remove-typescript-extension.patch
```diff
@@ -0,0 +1,36 @@
...
++// import { LanguageServerContribution } from '@theia/languages/lib/node';
++// import { TypeScriptContribution } from './typescript-contribution';
++// import { typescriptVersionPath, TypescriptVersionService } from '../common/typescript-version-service';
++// import { TypescriptVersionServiceImpl } from './typescript-version-service-impl';
+ 
+ export default new ContainerModule(bind => {
+-    bind(LanguageServerContribution).to(TypeScriptContribution).inSingletonScope();
++    // bind(LanguageServerContribution).to(TypeScriptContribution).inSingletonScope();
+ 
+-    bind(TypescriptVersionService).to(TypescriptVersionServiceImpl).inSingletonScope();
```
benoitf: why not remove package there https://github.com/ws-skeleton/che-theia-generator/blob/master/src/templates/assembly-package.mst#L39 ?

svor:
@benoitf @tolusha does it make sense to remove `typescript` extension from `theia/examples/browser/package.json` and `theia/examples/electron/package.json`

benoitf:
@svor AFAIK we don't touch the other examples, we only bring a new example (which is che-theia)

PR #80, opened 2019-02-21
Add Theia user preferences synchronizer

mmorhun:
Theia user preferences synchronizer is designed to handle Theia user settings synchronization between workspaces. It retrieves Theia user preferences from Che user preferences on Theia server start and watches changes, so when any update is done it will be reflected back to Che storage.

Depends on [Preferences API PR](https://github.com/eclipse/che-theia/pull/79)

Signed-off-by: Mykola Morhun <mmorhun@redhat.com>

PR #82, opened 2019-02-22
Create Theia config directory if it doesn't exist

mmorhun:
This PR brings fix for https://github.com/eclipse/che-theia/pull/80.
It handles the situation when Theia user directory hasn't been created yet, but the `eclipse-che-theia-user-preferences` extension tries to create the settings.json file.


benoitf:
maybe use `"fs-extra": "7.0.1"` instead ?

no need to use sync methods but can use async with await

benoitf:
https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/ensureDir.md

benoitf:
I've approved but better to use async calls if possible

PR #89, opened 2019-02-28
Show status bar message if workspace is in ephemeral mode

ashumilova:
Signed-off-by: Anna Shumilova <ashumilo@redhat.com>

Show warning to user in Che Theia status bar - "Ephemeral mode" and tooltip "All changes to the source code will be lost when the workspace is stopped unless they are pushed to a source code repository."

Issue - https://github.com/eclipse/che/issues/12565

screenshot from 2019-02-28 13-34-00

_A note: background of status bar item cannot be set by plugin API, thus only font color._

### DIFF
--- a/plugins/factory-plugin/src/factory-plugin.ts
+++ b/plugins/factory-plugin/src/factory-plugin.ts
@@ -11,7 +11,7 @@
 import * as theia from '@theia/plugin';
 import { FactoryInitializer } from './factory-initializer';
 import { WorkspaceProjectsManager } from './workspace-projects-manager';
-import { WorkspaceManager } from './workspace-manager';
+import { EphemeralWorkspaceChecker } from './ephemeral-workspace-checker';
 
 export async function start(context: theia.PluginContext) {
     let projectsRoot = '/projects';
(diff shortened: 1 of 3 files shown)


plugins/factory-plugin/src/workspace-manager.ts
```diff
@@ -0,0 +1,48 @@
...
+ **********************************************************************/
+
+import * as theia from '@theia/plugin';
+import * as che from '@eclipse-che/plugin';
+import { che as cheApi } from '@eclipse-che/api';
+
+/**
+ * Make checks on workspace configuration and shows dedicated information to user.
+ */
+export class WorkspaceManager {
```
sunix: I wouldn't call it `WorkspaceManager` because it is too generic. `Manager` usually is bad naming. Maybe `WorkspaceEphemeralChecker` ? or do you have other suggestion ?

plugins/factory-plugin/src/factory-plugin.ts
```diff
@@ -19,6 +20,7 @@ export async function start(context: theia.PluginContext) {
         projectsRoot = projectsRootEnvVar;
     }
 
+    await new WorkspaceManager().run();
```
sunix: i think that run should be run asynchrounously: we don't need to wait for that to be finished to start the `FactoryInitializer` right ?

ashumilova: yeah, you are right, no need.

plugins/factory-plugin/src/workspace-manager.ts
ashumilova: Maybe then EphemeralWorkspaceChecker?

vitaliy-guliy: +1 on `EphemeralWorkspaceChecker`

PR #116, opened 2019-03-18
Add cwd to the task plugin

AndrienkoAleksandr:
https://github.com/eclipse/che-theia/pull/116#issue-261948491


plugins/task-plugin/src/task/che-task-provider.ts
```diff
@@ -53,6 +53,10 @@ export class CheTaskProvider {
             resultTarget.machineName = await this.machinePicker.pick();
         }
 
+        if (target && target.workingDir) {
+            resultTarget.workingDir = cheTaskDefinition.target.workingDir;
```
RomanNikitenko: You already have `target` here, so can take `workingDir` from `target` directly, like  `resultTarget.workingDir = target.workingDir`

plugins/task-plugin/src/task/che-task-runner.ts
```diff
@@ -59,13 +59,16 @@ export class CheTaskRunner {
             throw new Error("Che task config must have 'target.machineName' property specified");
         }
 
+        const workingDir = target.workingDir;
```
RomanNikitenko: Do we need this const here? As I can see we don't use it in another places, so you can just pass `target.workingDir`. But up to you.

olexii4: so many men, so many minds
Make sense   
```
const cwd = target.workingDir;
...
          tty: true,
          cwd
};
```

PR #122, opened 2019-03-20
Add /projects to gopath. Install gcc and g++. Fix permissions

eivantsov:
This PR fixes golang docker image:

1. `/.cache` is created when debugger starts. On OpenShift this dir isn't writable, so it needs to be fixed.
2. No gcc and g++ installed, as a result go build fails (maybe hello world succeeds but a real golang project fails)
3. Add /projects to gopath, so that language server and tools it's using to provide intellisense work correctly.

With these changes I was able to import che-operator and debug it, test auto completion, code navigation etc:

image

@AndrienkoAleksandr I think you will be interested as well.


tsmaeder:
This is a fix for https://github.com/eclipse/che/issues/12878

dockerfiles/remote-plugin-go-1.10.7/Dockerfile
```diff
@@ -67,10 +67,11 @@ RUN set -eux; \
     ; \
 	apk del .build-deps \
     ; \
-	mkdir ${HOME}/.cache && chmod -R 777 ${HOME}/.cache \
+	mkdir /.cache && chmod -R 777 /.cache \
```
tsmaeder: Why does this need to be in root?

dockerfiles/remote-plugin-go-1.10.7/Dockerfile
```diff
@@ -67,10 +67,11 @@ RUN set -eux; \
...
 	; \
-    go version
+    go version && \
+    apk add gcc g++
```
tsmaeder: What do we need gcc?

dockerfiles/remote-plugin-go-1.10.7/Dockerfile
eivantsov: When launching debugger, it attempted to create .cache directory in the root. Not sure why HOME wasn't resolving properly.

```
/.cache/go-build $ pwd
/.cache/go-build $ ls -la
total 1164
drwxr-xr-x  258 10002400 root          4096 Mar 20 11:28 .
drwxrwxrwx    1 root     root          4096 Mar 20 11:27 ..
drwxr-xr-x    2 10002400 root          4096 Mar 20 11:27 00
drwxr-xr-x    2 10002400 root          4096 Mar 20 11:27 01
drwxr-xr-x    2 10002400 root          4096 Mar 20 11:27 02
```

dockerfiles/remote-plugin-go-1.10.7/Dockerfile
eivantsov: image

Probably not every project but it is not uncommon for a go build to depend on gcc and g++

AndrienkoAleksandr: I think these dependencies are not a problem. There are some projects which needed it. So for me it's looks ok.

AndrienkoAleksandr:
I tested it with . Works amazing https://github.com/eclipse/che-machine-exec
 @tsmaeder is it ok for you if this pr will be merged?

tsmaeder:
@AndrienkoAleksandr why not, I did not "request changes".

dockerfiles/remote-plugin-go-1.10.7/Dockerfile
benoitf: HOME was not set (now it is) so we should put it back to $HOME/.cache

eivantsov:
@AndrienkoAleksandr @tsmaeder thanks for review and questions. Merging.

It should not be possible to work with real life golang projects. The only thing golang devs need to do is to go get rather than git clone the project. So, go get will clone the project into `/projects/src/your/project` and this way it is in GOPATH for a language server.

dockerfiles/remote-plugin-go-1.10.7/Dockerfile
benoitf: @eivantsov ?

eivantsov: @benoitf sorry haven't seen your comment. Yes, I'll do that in a new PR then.

PR #130, opened 2019-03-22
[docker] [remote] Fix permissions on openshift

benoitf:
Some processes may want to create folders or files in the remote sidecar and if random uid is used, then permissions are denied to create folders in the HOME folder (now set to /home/theia)


benoitf:
ci-build

benoitf:
ci-build

benoitf:
ci-build

PR #132, opened 2019-03-24
Fix the workspace URL for UD

olexii4:
Existing solution, which depends on user ''che", doesn't work with che-multiuser.

In this PR regexp and replacement were fixed.

### DIFF
--- a/extensions/eclipse-che-theia-dashboard/src/browser/theia-dashboard-client.ts
+++ b/extensions/eclipse-che-theia-dashboard/src/browser/theia-dashboard-client.ts
@@ -92,7 +92,7 @@ export class TheiaDashboardClient implements FrontendApplicationContribution {
         }
         const ideWorkspaceUrl = workspace!.links!.ide!;
 
-        return ideWorkspaceUrl.replace('/che/', '/dashboard/#/ide/che/');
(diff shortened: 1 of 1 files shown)


extensions/eclipse-che-theia-dashboard/src/browser/theia-dashboard-client.ts
```diff
@@ -92,7 +92,7 @@ export class TheiaDashboardClient implements FrontendApplicationContribution {
         }
         const ideWorkspaceUrl = workspace!.links!.ide!;
 
-        return ideWorkspaceUrl.replace('/che/', '/dashboard/#/ide/che/');
+        return ideWorkspaceUrl.replace(/^https?:\/\/[^\/]+/, match => `${match}/dashboard/#/ide`);
```
AndrienkoAleksandr: So, it should work only for https? What about http?

akurinnoy: The question mark in a regexp makes preceding token optional. Hence, `/^https?:\/\//` matches both `https://` and `http://`.

AndrienkoAleksandr: Oh, sorry I see.

PR #133, opened 2019-03-25
Workspace from devfile

vitaliy-guliy:
Makes it possible to create a workspace from a devfile using a factory service

Add necessary functionality to the factory plugin.

create workspace from devfile

Solves issue https://github.com/eclipse/che/issues/12707


extensions/eclipse-che-theia-plugin-ext/src/browser/che-devfile-main.ts
```diff
@@ -0,0 +1,117 @@
+/*********************************************************************
+ * Copyright (c) 2018 Red Hat, Inc.
```
olexii4: 2019

extensions/eclipse-che-theia-plugin-ext/src/plugin/che-devfile.ts
```diff
@@ -0,0 +1,30 @@
+/*********************************************************************
+ * Copyright (c) 2018 Red Hat, Inc.
```
olexii4: 2019

extensions/eclipse-che-theia-plugin-ext/src/browser/che-devfile-main.ts
```diff
@@ -0,0 +1,117 @@
...
+            resolve();
+        });
+    }
+
+    /**
+     * Opens window with URL to the factory.
+     */
+    protected async openFactoryWindow(uri: string): Promise<void> {
+        // do nothing for electron browser
+        if (isNative) {
```
olexii4: We don't need to check it

extensions/eclipse-che-theia-plugin-ext/src/browser/che-devfile-main.ts
vitaliy-guliy: Done

extensions/eclipse-che-theia-plugin-ext/src/plugin/che-devfile.ts
vitaliy-guliy: Done

extensions/eclipse-che-theia-plugin-ext/src/browser/che-devfile-main.ts
vitaliy-guliy: Yep, indeed. Thanks

plugins/factory-plugin/src/devfile.ts
```diff
@@ -0,0 +1,44 @@
...
+    constructor(private context: theia.PluginContext) {
+    }
+
+    init() {
+        this.context.subscriptions.push(theia.commands.registerCommand(
+            CREATE_WORKSPACE_COMMAND, uri => this.createWorkspace(uri))
+        );
+    }
+
+    async createWorkspace(uri: theia.Uri) {
```
benoitf: missing return type ?

plugins/factory-plugin/package.json
```diff
@@ -53,5 +53,27 @@
             "json"
         ],
         "testURL": "http://localhost/"
+    },
+    "contributes": {
+        "menus": {
+            "editor/context": [
+                {
+                    "command": "factory-plugin.create-workspace",
+                    "when": "resourceFilename =~ /devfile/"
```
benoitf: AFAIK now the devfile is named `.devfile.yml` no ?

vitaliy-guliy: All names containing 'devfile' are suitable.
E.g. `devfile`, `.devfile`, `devfile.yaml` and others.
But this mask is not case sensitive, so `Devfile` will not be handled.

plugins/factory-plugin/src/devfile.ts
vitaliy-guliy: void

vitaliy-guliy: It's a command handler. I suppose we can omit return type here. Isn't it?

benoitf: I don't know which tslint rules we're using but I was thinking it was mandatory

vitaliy-guliy: It's not a problem to add return type after adding corresponding rule for tslint.

Issue #111, opened 2019-03-15
Terminal API: sending text to the newly created terminal doesn't work as expected


azatsarynnyy:
### Description
[`window.createTerminal`](https://github.com/theia-ide/theia/blob/b21023ef38060a127b0c152bdf71344de8950f59/packages/plugin/src/theia.d.ts#L3126) followed by [`terminal.sendText`](https://github.com/theia-ide/theia/blob/b21023ef38060a127b0c152bdf71344de8950f59/packages/plugin/src/theia.d.ts#L2340) leads to opening an empty terminal.

### Reproduction Steps
1/ Run the [terminal-plugin sample](https://github.com/eclipse/che-theia-samples/tree/master/samples/terminal-plugin) in Che Theia
2/ Call the `Send text to the terminal` command from the Command Palette.

#### Expected behavior:
New Terminal is created and text is printed to the terminal
![image](https://user-images.githubusercontent.com/1636395/54398444-be311c00-46c2-11e9-96dd-c38e9cb460a7.png)
**Note:** it works as expected in standalone Theia

#### Actual behavior:
New Terminal is created but the text isn't printed to the terminal
![image](https://user-images.githubusercontent.com/1636395/54398413-95a92200-46c2-11e9-862b-b96e70e5c32d.png)

azatsarynnyy:
It's important for running the Kubernetes Tooling Che Plugin since a lot of the plugin's commands depend on these API calls.

ashumilova:
Added to the next sprint with High priority. Thank you for notifying.

sunix:
@AndrienkoAleksandr can we close this issue ?

Issue #112, opened 2019-03-15
Terminal API doesn't work for a plugin running in a sidecar container


azatsarynnyy:
### Description
[Theia Plugin Terminal API](https://github.com/theia-ide/theia/blob/b21023ef38060a127b0c152bdf71344de8950f59/packages/plugin/src/theia.d.ts#L3126) doesn't work for a plugin running in a sidecar container.

### Reproduction Steps
1/ Start Che Workspace with the [terminal-plugin sample](https://github.com/eclipse/che-theia-samples/tree/master/samples/terminal-plugin) running in its own container.
Use the DevFile:
```yaml
specVersion: 0.0.1
name: terminal-sample
tools:
  - name: theia-ide
    type: cheEditor
    id: org.eclipse.che.editor.theia:1.0.0
  - name: exec-plugin
    type: chePlugin
    id: che-machine-exec-plugin:0.0.1
  - name: terminal-sample-plugin
    type: chePlugin
    id: https://raw.githubusercontent.com/azatsarynnyy/che7-plugin-registry/master/terminal-sample:0.0.1

```
2/ Call the `Send text to the terminal` command from Theia Command Palette.

#### Expected behavior:
New Terminal is created and text is printed to the terminal
![image](https://user-images.githubusercontent.com/1636395/54398444-be311c00-46c2-11e9-96dd-c38e9cb460a7.png)

#### Diagnostics:
from browser's console:
```
Uncaught (in promise) TypeError: Cannot read property 'proxy' of undefined
    at TerminalServiceExtImpl.createTerminal (/home/theia/node_modules/@theia/plugin-ext/lib/plugin/terminal-ext.js:77)
    at TerminalServiceExtImpl.createTerminal (/home/theia/lib/node/terminal-container-aware.js:43)
    at Object.createTerminal (/home/theia/node_modules/@theia/plugin-ext/lib/plugin/plugin-context.js:299)
    at createTerminalWithOptions (/plugins/terminal-sample.0.0.1/lib/new-terminal-backend.js:262)
    at /plugins/terminal-sample.0.0.1/lib/new-terminal-backend.js:106
    at /home/theia/node_modules/@theia/plugin-ext/lib/plugin/command-registry.js:113
    at CommandRegistryImpl.executeLocalCommand (/home/theia/node_modules/@theia/plugin-ext/lib/plugin/command-registry.js:165)
[...]

PR #145, opened 2019-03-27
Fix send text for plugin api.

AndrienkoAleksandr:
### Referenced issue:
https://github.com/eclipse/che-theia/issues/111


AndrienkoAleksandr:
> I wonder does it also fix this one #112 or that's another problem?
@azatsarynnyy I think it's separated one. We have it in the current sprint.

PR #146, opened 2019-03-28
Invalidate docker cache during plugin code clone

evidolob:


dockerfiles/theia/Dockerfile
```diff
@@ -112,6 +112,9 @@ RUN if [ ! -z "${GITHUB_TOKEN-}" ]; then \
       fi \
     fi
 
+#invalidate cashe
```
benoitf: typo: cache (there is same typo later in the file)

PR #156, opened 2019-04-03
[ports plug-in] Limit the number of pre-defined routes to #3

benoitf:
### What does this PR do?

Limit the number of pre-defined routes to #3
Also do not propose redirect for ephemeral ports

#139 

Change-Id: I98570624346a5a63cfd55da66677cced634aa775


benoitf:
ci-build

slemeur:
Should this be configurable?

benoitf:
@slemeur what part configurable ? list of exposed servers is not 'configurable' in workspace definition. Either we declare something or we don't declare it but we can't provide a flag like 'create me CHE_DUMMY_ROUTES routes'

PR #178, opened 2019-04-15
Move user containers from root to 'User Runtimes' node. 

mmorhun:
### What does this PR do?
Renames `Containers` panel into `My Workspace`.
Moves user defined containers from root node to `User Runtimes` node, so root has only two nodes: `User Runtimes` and `Plugins`

https://github.com/eclipse/che/issues/12984

#### Release Notes
`Containers` panel is renamed into `My workspace`

#### Screenshots
che-12984

For case, when no runtime containers are defined:
che-12984-no-runtime

### DIFF
--- a/plugins/containers-plugin/package.json
+++ b/plugins/containers-plugin/package.json
@@ -14,16 +14,16 @@
     "viewsContainers": {
       "right": [
         {
-          "id": "containers",
-          "title": "Containers",
+          "id": "my-workspace",
+          "title": "My Workspace",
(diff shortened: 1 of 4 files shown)


mmorhun:
@slemeur is it looks good to you?

slemeur:
Yep good for me !

sunix:
@slemeur @mmorhun coming back after the battle. Why having a single node `WORKSPACE` that has no sibling ?

PR #179, opened 2019-04-15
Add example docker run command to readme

tetchel:
### What does this PR do?
Adds example `docker run` command to run `che-theia` locally. This shows developers:
1) It runs on port 3100, not 3000 like regular Theia.
2) How to mount the desired workspace.


benoitf:
Hello @tetchel I'm not sure we can run Che-theia like that.
It expects to have several environment variables to be set to be properly working and these variables are provided by Che Workspace Master.
So I would say it could mislead users to have such instructions

tetchel:
yes, https://github.com/eclipse/che-theia/issues/180#issuecomment-483522408 made me realize that.

PR #181, opened 2019-04-16
Introduce Che Theia about menu instead of Theia one

benoitf:
### What does this PR do?

It provides Eclipse Che logo and sha1 of che theia and theia

image

N/A

Change-Id: I0baa118b7273226ecc479a15b840568fcc35cfaa


extensions/eclipse-che-theia-about/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@eclipse-che/theia-about",
```
mmorhun: It is up to you, but to me `@eclipse-che/che-theia-about` looks more consistent.

benoitf: @mmorhun I tried to be consistent with existing extensions
https://github.com/eclipse/che-theia/blob/master/extensions/eclipse-che-theia-plugin-ext/package.json#L2

https://github.com/eclipse/che-theia/blob/master/extensions/eclipse-che-theia-dashboard/package.json#L2

it seems they're using the pattern `@eclipse-che/theia-something`

benoitf:
ci-build

extensions/eclipse-che-theia-about/package.json
mmorhun: I see. My point was that `@eclipse-che` is namespace and then we need to have a component which is Che Theia About, like you have in class and file names: `AboutCheTheiaDialog`, `about-che-theia-dialog.ts`

Issue #155, opened 2019-04-03
running a task/che-commands is showing error: `Request {taskname} failed with error: Information not found`


sunix:
### Description
When trying to run a che-command, `Request {taskname} failed with error: Information not found` message is display and the command is not executed

### Reproduction Steps
Run this devfile:

```yaml
specVersion: 0.0.1
name: sunix-devfiles
projects:
  - name: devfiles
    source:
      type: git
      location: 'https://github.com/sunix/devfiles.git'
components:
  - name: theia-editor
    type: cheEditor
    id: org.eclipse.che.editor.theia:next
  - name: exec-plugin
    type: chePlugin
    id: che-machine-exec-plugin:0.0.1
  - name: git
    type: dockerimage
    image: sunix/git-devtools
    mountSources: true
    memoryLimit: 256M
  - name: theia-endpoint-runtime
    type: dockerimage
    image: eclipse/che-theia-endpoint-runtime:nightly
    endpoints:
      - name: theia-remote-endpoint
        port: 10000
        attributes:
          protocol: ws
          public: 'true'
    memoryLimit: 512M
    mountSources: true
  - name: vscode-yaml
    type: chePlugin
    id: redhat.vscode-yaml:0.3.0
commands:
  - name: run
    actions:
      - type: exec
        component: git
        command: pwd

```
run the command `run` it is supposed to execute the command `pwd` but it is not and showing in a dialog box: `Error launching task 'run': Request run failed with message: Information not found`

and in theia ide logs:
```
root ERROR Request run failed with error: Information not found Error: Information not found
    at CheTask.<anonymous> (/home/theia/node_modules/@eclipse-che/theia-plugin-ext/lib/node/che-task-service.js:216:35)
    at step (/home/theia/node_modules/@eclipse-che/theia-plugin-ext/lib/node/che-task-service.js:54:23)
    at Object.next (/home/theia/node_modules/@eclipse-che/theia-plugin-ext/lib/node/che-task-service.js:35:53)
    at fulfilled (/home/theia/node_modules/@eclipse-che/theia-plugin-ext/lib/node/che-task-service.js:26:58)
    at <anonymous>
task ERROR Error launching task 'init ... che-theia': Request run failed with message: Information not found
```

**OS and version:**    
Minikube - che7 started with chectl

tolusha:
It is impossible to test new stacks with plugins ^(
https://github.com/eclipse/che/issues/12944

tsmaeder:
[...]

PR #183, opened 2019-04-16
Do not override task plugin handlers

RomanNikitenko:
### What does this PR do?
At the moment remote plugins can override task plugin handlers.
This way is described [here](https://github.com/eclipse/che-theia/issues/155#issuecomment-482510278).
The PR allows to handle all set subscribed handlers.  

https://github.com/eclipse/che-theia/issues/155

### DIFF
--- a/extensions/eclipse-che-theia-plugin-ext/src/browser/che-task-client.ts
+++ b/extensions/eclipse-che-theia-plugin-ext/src/browser/che-task-client.ts
@@ -15,22 +15,27 @@ import { TaskConfiguration, TaskInfo } from '@eclipse-che/plugin';
 @injectable()
 export class CheTaskClientImpl implements CheTaskClient {
     private readonly onKillEventEmitter: Emitter<number>;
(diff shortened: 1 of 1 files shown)


extensions/eclipse-che-theia-plugin-ext/src/browser/che-task-client.ts
```diff
@@ -15,22 +15,27 @@ import { TaskConfiguration, TaskInfo } from '@eclipse-che/plugin';
...
         }
+        return undefined;
     }
 
     async getTaskInfo(id: number): Promise<TaskInfo | undefined> {
-        if (this.taskInfoHandler) {
-            return await this.taskInfoHandler(id);
+        for (const taskInfoHandler of this.taskInfoHandlers) {
+            const taskInfo = await taskInfoHandler(id);
+            if (taskInfo) {
```
vinokurig: The method returns Promise which may be `undefined`, so the `if` check is not necessary

Issue #45, opened 2019-02-01
Build and package che-theia plugins during che-theia image build.


sunix:
The same way we are building extensions, we would build che-theia plugins from sources and package them so they are available right away.

1. skate board step: https://github.com/eclipse/che-theia/pull/43 
2. moto cycle step: include che-theia plugins build into the `che:theia` generator
    - `che:theia init` would also clone plugins projects
        ```
        ${theia}/che/<clone ext/plugins repos>
        ${theia}/packages/<ln -s extensions>
        ${theia}/plugins/<ln -s plugins>
        ```

- [x] it would use a single `extensions.yml` file. we would specify which project is plugin:
         -  move `extensions/extensions.yml` to `che-theia-init-sources.yml`
         -  renamed `folders` to `extensions` , root `extensions` to `sources` and adding new folders for plugins: `plugins`
- [x] build all the plugins `${theia}/plugins` with a single command: provides a script located in plugins folder that would perform `yarn` on each plugins.
- [x] packaging the future content of `/default-theia-plugins/`: the script `foreach_yarn` is also copying the .theia files in production/plugins folder.
- [x] update CI https://github.com/eclipse/che-theia/pull/201

benoitf:
I like the proposal

probably rename `extensions.yml` to a more generic name (as for now extensions.yml was only for extensions)

I think we can avoid default (no need of compatibility AFAIK)

AndrienkoAleksandr:
+1 For generic name.

sunix:
- `che-theia-customizations.yml` ?
- `customizations.yml` ?

sunix:
currently testing to build theia (after che:theia init) with symb links that i added manually to plugins but got this issue:
```
[compile] yarn compile exited with code 0
Packaging of plugin
🔍 Validating...✔️
🗂  Getting dependencies...❌️
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

    at Promise.all.then.arr (/projects/theia/node_modules/lerna/node_modules/execa/index.js:236:11)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
  code: 1,
  killed: false,
[...]

PR #197, opened 2019-04-29
Preparing next che:theia init generator model. Will move to the root …folder of the repo, rename root element and folders

sunix:
### What does this PR do?
Propose the next format of `extensions.yml`, following the discussion https://github.com/ws-skeleton/che-theia-generator/pull/20#issuecomment-483263490:

- moving `extensions/extensions.yml` to `che-theia-init-sources.yml`
- renamed root `extensions` item to `sources`
- rename `pluginFolders` to `plugins`
- rename `extensionFolders` to `extensions`

https://github.com/eclipse/che-theia/issues/45

### DIFF
--- a/che-theia-init-sources.yml
+++ b/che-theia-init-sources.yml
@@ -1,6 +1,6 @@
-extensions:
+sources:
 - source: https://github.com/eclipse/che-theia
-  folders:
+  extensions:
     - dockerfiles/theia-endpoint-runtime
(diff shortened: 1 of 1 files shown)


benoitf:
It will require the update of generator as well no ?

sunix:
yes this is keeping the old file with the old format. We will keep it until the generator and builder would be updated.

PR #201, opened 2019-05-02
Using latest che theia generator. Introducing che-theia-init-sources.…

sunix:
…yml files that handles also plugins. For build and package che-theia plugins during che-theia image build.

### What does this PR do?
Use the new version of the che:theia cli. Now only one clone is done for extensions and plugins.
plugins that have to be included are done in che-theia-init-sources.yml file (replacing extensions/extensions.yml).

https://github.com/eclipse/che-theia/issues/45

### DIFF
--- a/dockerfiles/theia-dev/Dockerfile
+++ b/dockerfiles/theia-dev/Dockerfile
@@ -38,7 +38,7 @@ ENV HOME=/home/theia-dev \
     GIT_EXEC_PATH=/usr/libexec/git-core
 
 # Define package of the theia generator to use
-ARG THEIA_GENERATOR_PACKAGE=@eclipse-che/theia-generator@0.0.1-1553159770
+ARG THEIA_GENERATOR_PACKAGE=@eclipse-che/theia-generator@0.0.1-1556612396
 
 WORKDIR ${HOME}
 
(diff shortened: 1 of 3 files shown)


dockerfiles/theia/Dockerfile
```diff
@@ -77,52 +77,13 @@ RUN yarn
 # Run into production mode
 RUN che:theia production
 
+# Compile plugins
+RUN cd plugins && ./foreach_yarn
```
AndrienkoAleksandr: Sorry, @sun where is located script `foreach_yarn` or maybe you forgot to push it? I can not find it.

sunix: @AndrienkoAleksandr  it is provided by the generator when you perform a che:theia init

dockerfiles/theia-dev/Dockerfile
```diff
@@ -38,7 +38,7 @@ ENV HOME=/home/theia-dev \
     GIT_EXEC_PATH=/usr/libexec/git-core
 
 # Define package of the theia generator to use
-ARG THEIA_GENERATOR_PACKAGE=@eclipse-che/theia-generator@0.0.1-1553159770
+ARG THEIA_GENERATOR_PACKAGE=@eclipse-che/theia-generator@0.0.1-1556612396
```
sunix: @AndrienkoAleksandr here the new version of the generator

PR #207, opened 2019-05-07
Introducing devfiles for che-theia in che.

sunix:
### What does this PR do?
Moving devfiles from https://github.com/sunix/devfiles. These devfile generates workspace environment to work on cheTheia

### DIFF
--- a/devfiles/che-theia-generator-cli.devfile.yaml
+++ b/devfiles/che-theia-generator-cli.devfile.yaml
@@ -1,0 +1,98 @@
+# A devfile to work on the che-theia generator cli: che:theia init, che:theia production, etc ...
+
+specVersion: 0.0.1
+name: cheTheiaGenerator-in-che
+projects:
+  - name: che-theia
+    source:
+      type: git
+      location: 'https://github.com/eclipse/che-theia.git'
+  - name: theia
+    source:
+      type: git
+      location: 'https://github.com/theia-ide/theia.git'
+  - name: che-theia-generator
+    source:
+      type: git
+      location: 'https://github.com/ws-skeleton/che-theia-generator.git'
+components:
+  - 
+    alias: git-devtools
+    type: dockerimage
+    image: sunix/git-devtools
+    mountSources: true
+    memoryLimit: 256M
+  - 
+    alias: che-dev
+    type: dockerimage
+    image: eclipse/che-theia-dev:nightly
+    mountSources: true
+    endpoints:
+      - name: "theia-dev-flow"
+        port: 3010
+        attributes:
+          protocol: http
+          public: 'true'
+    memoryLimit: "2Gi"
+  - 
+    alias: theia-editor
+    id: eclipse/che-theia/next
+    type: cheEditor
+  - 
+    id: eclipse/che-machine-exec-plugin/0.0.1
+    type: chePlugin
+  - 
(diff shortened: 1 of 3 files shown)


devfiles/che-theia-generator-cli.devfile.yaml
```diff
@@ -0,0 +1,98 @@
...
+      location: 'https://github.com/theia-ide/theia.git'
+  - name: che-theia-generator
+    source:
+      type: git
+      location: 'https://github.com/ws-skeleton/che-theia-generator.git'
+components:
+  - 
+    alias: git-devtools
+    type: dockerimage
+    image: sunix/git-devtools
```
mmorhun: We mustn't use personal images in Eclipse projects. Please move the image to Eclipse organization.

sunix: Why not ? These are not binaries, what about source code like spring petclinic you clone, what about all the components that are vscode extension and not from Eclipse ?

AndrienkoAleksandr: I am not sure that everyone need to use this container. Maybe don't include this container, and it will save a bit ram.

mmorhun: @sunix it is not appropriate for organization like Eclipse.

> These are not binaries

No, with docker image you bring binaries, which is build and stored in your personal account.

sunix: 1. This is not bringing third party dependencies to the the project.
2. It means that you can not use yarn or npm or git or maven or docker because these are not Eclipse. What if you have a script in che-theia repo that invoking docker build. Docker is not a Eclipse project so it can't be in a script in che-theia repo ?
3. You don't use git from the terminal ?

but I am ok to move it. Just your reasons don't make sense.

mmorhun: 1. How is this related to binaries inside personal docker image?
2. You didn't understand me. I am not talking about using tools, my concern is about image origin.
3. I use git from terminal, but so far haven't use `sunix/git-devtools` because git included in other containers.

About reasons. Think about it from another side: why do we have all images in Eclipse organization and not in personal accounts? Why do we need Eclipse namespace on dockerhub? I hope you'll understand the problem, so your arguments doesn't make sense.

evidolob: For me, we should not depend on personal github repositories in our public project.
And if we put some devfile which depends on your docker image we should put dockerfile for that image in this repository. It's a matter of maintainability of that devfile and docker image.

sunix: so in our devfiles we should not use any extensions from vscode marketplace because of maintainability, etc ... let's close this issue for the time being because we have plenty of other things to deal with then debating ...

AndrienkoAleksandr:
@sun You could create che plugin with this one tool sunix/git-devtools in the separated pull request, and contribute dockerfile and so on for this tool and maintain it in the che-theia project. 
By default we could create dev workspaces from your devfiles without this plugin, but if somebody need it, he could simple activate this tool like plugin with help che-theia ui or dashboard. I think your devfiles are good and it make sense to merge this pr.
For git-devtools you could even create on more advanced devfile.

PR #215, opened 2019-05-10
Disable replacing ffmpeg for electron

evidolob:
To restore build on CI we need to disable replacing ffmpeg for electron

### DIFF
--- a/dockerfiles/theia-dev/Dockerfile
+++ b/dockerfiles/theia-dev/Dockerfile
@@ -35,7 +35,8 @@ ENV HOME=/home/theia-dev \
     # Specify the directory of git (avoid to search at init of Theia)
     USE_LOCAL_GIT=true \
     LOCAL_GIT_DIRECTORY=/usr \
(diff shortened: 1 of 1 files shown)


dockerfiles/theia-dev/Dockerfile
```diff
@@ -66,8 +67,8 @@ RUN npm config set prefix "${HOME}/.npm-global" && \
     echo '{"optOut": true}' > ${HOME}/.config/insight-nodejs/insight-yo.json && \
     # Change permissions to let any arbitrary user
     for f in "${HOME}" "/etc/passwd" "/etc/group" "/projects"; do \
-        echo "Changing permissions on ${f}" && chgrp -R 0 ${f} && \
-        chmod -R g+rwX ${f}; \
+    echo "Changing permissions on ${f}" && chgrp -R 0 ${f} && \
```
sunix: why this change ?

mmorhun: I think we should have an indent here.

PR #220, opened 2019-05-15
Allow Kubernetes Tooling Che Plugin use Buildah as an image builder instead of Docker daemon

azatsarynnyy:

### What does this PR do?
Allows the [Kubernetes Tooling Che Plugin](https://github.com/eclipse/che-plugin-registry/tree/master/v3/plugins/ms-kubernetes-tools/vscode-kubernetes-tools) use [Buildah](https://github.com/containers/buildah) as an image builder instead of Docker daemon.
Currently, it's implemented with the `docker.sh` wrapper script that calls `buildah`.
In the future, we can add support for Buildah by [VS Code Kubernetes extension](https://github.com/Azure/vscode-kubernetes-tools/) natively and the wrapper script is not needed.

closes https://github.com/eclipse/che/issues/13316

### DIFF
--- a/dockerfiles/remote-plugin-kubernetes-tooling-1.0.0/Dockerfile
+++ b/dockerfiles/remote-plugin-kubernetes-tooling-1.0.0/Dockerfile
@@ -11,15 +11,15 @@
 FROM ${BUILD_ORGANIZATION}/${BUILD_PREFIX}-theia-endpoint-runtime:next as endpoint
 FROM quay.io/buildah/upstream:v1.8.2
 
+ENV KUBECTL_VERSION v1.14.1
+ENV HELM_VERSION v2.13.1
 ENV HOME=/home/theia
+
 COPY --from=endpoint /home/theia /home/theia
 COPY --from=endpoint /projects /projects
 COPY --from=endpoint /etc/passwd /etc/passwd
 COPY --from=endpoint /etc/group /etc/group
 COPY --from=endpoint /entrypoint.sh /entrypoint.sh
-
-ENV KUBECTL_VERSION v1.14.1
-ENV HELM_VERSION v2.13.1
 
 ADD etc/docker.sh /usr/local/bin/docker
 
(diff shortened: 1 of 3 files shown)


azatsarynnyy:
@benoitf could you take a look, please

dockerfiles/remote-plugin-kubernetes-tooling-1.0.0/etc/docker.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+CMD=$1
+if [[ $1 = build ]]; then
+    CMD="bud"
+fi
+
+buildah $CMD $2 $3 $4 $5 $6 $7 $8 $9
```
riuvshin: maybe it is better to use just `$@` ?

azatsarynnyy: It will require using `sed` to replace the argument in `$@` if I understand correctly.
Since I'm not good at `sed` I chose that approach.

benoitf: you should use `shift` 
```
shift
buildah $CMD $@
```

so the first argument is removed ($1)

dockerfiles/remote-plugin-kubernetes-tooling-1.0.0/etc/docker.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
benoitf: it is missing copyright headers

dockerfiles/remote-plugin-kubernetes-tooling-1.0.0/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+FROM quay.io/buildah/upstream:v1.8.2
+
+ENV HOME=/home/theia
+COPY --from=endpoint /home/theia /home/theia
+COPY --from=endpoint /projects /projects
+COPY --from=endpoint /etc/passwd /etc/passwd
+COPY --from=endpoint /etc/group /etc/group
+COPY --from=endpoint /entrypoint.sh /entrypoint.sh
+
+ENV KUBECTL_VERSION v1.14.1
```
benoitf: you should move these instructions before the COPY instructions to have better cache of layers

azatsarynnyy:
@benoitf I've fixed all your notes. Thanks for the review!

PR #221, opened 2019-05-16
CHE-13112: Respect workspaces based on devfile in Che Theia IDE UI

mmorhun:

### What does this PR do?
Adds functionality to support devfile configuration of workspace on Theia UI. So now it supports both: devfile and workspace config.
The changes is mostly about projects sync between Theia and devfile projects section.
Also adds small changed to workspace panel to be able to show appropriate info from devfile. But with current state we cannot get information about volumes, so in case of devfile config the information about volumes is missing for now. See https://github.com/eclipse/che/issues/13290.

https://github.com/eclipse/che/issues/13112

### DIFF
--- a/plugins/containers-plugin/src/containers-service.ts
+++ b/plugins/containers-plugin/src/containers-service.ts
@@ -28,6 +28,11 @@ export interface IContainer {
     commands?: string[]
 }
 
+enum ContainerType {
+    TOOL = 'tool',
+    USER_DEFINED = 'recipe'
+}
+
 const MAX_FAILED_ATTEMPTS = 5;
 
 export class ContainersService {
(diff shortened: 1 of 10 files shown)


olexii4:
https://github.com/eclipse/che-theia/blob/374f5b2dfeabf540d6377e51c905541e9e0e033a/plugins/containers-plugin/src/containers-tree-data-provider.ts#L109

command: { id: 'task:run', arguments: ['che', commandName] }  -> command: { id: 'task:run', arguments: ['che', commandName, container.name] }

WDYT?

plugins/containers-plugin/src/containers-service.ts
```diff
@@ -47,23 +52,46 @@ export class ContainersService {
...
+                    // should never happen
+                    throw new Error('Invalid workspace structure');
+                }
             }
-            if (workspace.runtime!.commands) {
+            if (runtime.commands) {
                 container.commands = [];
-                workspace.runtime.commands.forEach(command => {
+                runtime.commands.forEach(command => {
                     if (command.attributes && command.attributes.machineName && command.attributes.machineName !== name) {
```
AndrienkoAleksandr: So, if command without attributes  at all, we push this command to the container? Maybe for such commands we should apply attribute with machineName manually and than push to the container command?

plugins/factory-plugin/src/projects.ts
```diff
@@ -7,13 +7,88 @@
...
+    projectPath: string,
+    projectGitLocation: string,
+    projectGitRemoteBranch: string
+): cheApi.workspace.devfile.Project[] {
+
+    if (!projects) {
+        projects = [];
+    }
+
+    for (let i = 0; i < projects.length; i++) {
```
AndrienkoAleksandr: Maybe `for...of` will be better? But up to you.

plugins/containers-plugin/src/containers-service.ts
mmorhun: I'll handle it in https://github.com/eclipse/che-theia/issues/222

plugins/factory-plugin/src/workspace-projects-manager.ts
```diff
@@ -66,70 +70,150 @@ export class WorkspaceProjectsManager {
     async startSyncWorkspaceProjects() {
         const gitConfigPattern = '**/.git/{HEAD,config}';
         const gitConfigWatcher = theia.workspace.createFileSystemWatcher(gitConfigPattern);
-        gitConfigWatcher.onDidCreate(uri => this.updateOrCreateGitProjectInWorkspace(git.getGitRootFolder(uri.path)));
-        gitConfigWatcher.onDidChange(uri => this.updateOrCreateGitProjectInWorkspace(git.getGitRootFolder(uri.path)));
-        gitConfigWatcher.onDidDelete(uri => this.deleteGitProjectInWorkspace(git.getGitRootFolder(uri.path)));
+        gitConfigWatcher.onDidCreate(uri => this.updateOrCreateProjectInWorkspace(git.getGitRootFolder(uri.path)));
```
AndrienkoAleksandr: onDidCreate, onDidChange, onDidDelete returns Disposable, push them to the plugin context please.

mmorhun: thanks!

mmorhun: They will be disposed when the watcher gets disposed

Issue #192, opened 2019-04-25
tasks: output of a command is not rendered correctly


sunix:
### Description
Executing che commands in che-theia with a che-theia task (from the container view or command palette `run tasks`), output is not looking good. special characters bash code for colors seems not to be taken in account. Also progress output are generating new lines when updated.
See screenshot:

![output](https://user-images.githubusercontent.com/5676062/57067089-b52df580-6cd6-11e9-956f-8214a89549e2.png)

### Reproduction Steps
- Open this devfile https://github.com/sunix/devfiles/blob/41ef94d8cff4964d99ebec0b5da0d3136d69e938/che-in-che/che-theia.devfile.yaml
- Start any che task like che:theia init

- Output is display but no colors and weird special characters are display

**OS and version:**    
Minikube and che7 nightly built version

vparfonov:
Actually we wait on fix this epic https://github.com/eclipse/che/issues/11458 for replacing current output widget with terminal widget

RomanNikitenko:
It was temporary solution to display output of a command to output channel. Within this [issue](https://github.com/eclipse/che/issues/11458) we depend on https://github.com/eclipse/che/issues/11442
I can provide a small improvement to fix this issue for output channel.
![output](https://user-images.githubusercontent.com/5676062/56742181-ea898f00-677c-11e9-9c66-a12f6bea12a5.png)

If it's OK that colors will not to be taken in account, I'll provide the PR
Then we switch task's output from output channel to terminal widget when https://github.com/eclipse/che/issues/11442 is resolved.

AndrienkoAleksandr:
+1 for small improvements.

sunix:
If it is a quick workaround, could you create the PR so we know what it is about ?

sunix:
sorry I am reopening it as the workaround does not solve everything

RomanNikitenko:
@sunix 
may be we should edit it like "switch task's output from output channel to terminal widget"
WDYT?

sunix:
@RomanNikitenko I think we should not change the title, 'output of a command' is STILL not rendered correctly (no colors)
Maybe one of the PR to fix this would have the title you are suggesting but it is the difference between a bug/feature and tasks to fix/implement that.

RomanNikitenko:
@sunix 
I think output channel is not designed for task output :-)
Are you sure a month ago we used output channel for task output?

btw: is screenshot actual at the moment?

sunix:
[...]

PR #223, opened 2019-05-16
Use terminal widget for output rendering in the task plugin 

AndrienkoAleksandr:
### What does this PR do?
Proposed changes:
- Use terminal widget for output rendering in the task plugin.
- Fix terminal onDidClosed method for terminal plugin api.

https://github.com/eclipse/che-theia/issues/192


plugins/task-plugin/src/task/che-task-runner.ts
```diff
@@ -62,30 +55,26 @@ export class CheTaskRunner {
...
-
         try {
-            const execId = await this.machineExecClient.getExecId(machineExec);
-            const terminalWidget = await this.terminalWidgetFactory.createWidget({ title: taskConfig.label, terminalId: execId });
-            terminalWidget.connectTerminalProcess();
+            const terminalOptions: theia.TerminalOptions = {
+                cwd: target.workingDir,
+                name: taskConfig.label,
+                shellPath: 'sh',
+                shellArgs: ['-c', "'" + taskConfig.command + "'"],
```
benoitf: shouldn't we use string literals ?

RomanNikitenko:
@AndrienkoAleksandr 
build_task

Is it possible to reuse widget for the same task if task was completed?
If so and it makes sense, we could improve this behavior, maybe in another issue

AndrienkoAleksandr:
> @AndrienkoAleksandr
I tested your changes and tried to run a task in ws/dev container, but it always was run in theia-ide container.
Could you test this case?
Please see video https://youtu.be/qooQnIPSKNI

Thanks a lot. Fixed!

Issue #190, opened 2019-04-23
[port plug-in] Filter out all ephemeral ports


svor:
### Description
 
It would be nice to filter out all ephemeral ports to avoid annoying error message about exposing ports:
![screenshot-che-mini-che 192 168 99 102 nip io-2019 04 23-12-48-13](https://user-images.githubusercontent.com/1271546/56571661-16aedf80-65c6-11e9-9f98-979292e56781.png)
 
### Reproduction Steps
To see this error window you need to run Minishift with remote Java plugin and try to debug something:
![output](https://user-images.githubusercontent.com/1271546/56572009-d13ee200-65c6-11e9-9d4c-c5c637ee5ddb.gif)

**OS and version:**    

**Diagnostics:**

benoitf:
![ephemeral](https://user-images.githubusercontent.com/436777/57788188-f393cd80-7736-11e9-9c4d-2c1f42c34418.gif)

Is it ok ?

svor:
It's ok for me, thanks!

PR #224, opened 2019-05-16
[Ports Plugin] excluding ports

benoitf:
### What does this PR do?

Allow to exclude some ports for Ports Plugin

It will exclude for now 3130 port used in hosted mode of Theia
It fixes https://github.com/eclipse/che/issues/13300

Also move trace about ephemeral port to Ports Plugin output channel
It fixes https://github.com/eclipse/che-theia/issues/190

Change-Id: I50e4891028a7bacf60d7376fe05a99a835e429e9

### DIFF
--- a/plugins/ports-plugin/src/ports-plugin.ts
+++ b/plugins/ports-plugin/src/ports-plugin.ts
@@ -25,11 +25,14 @@ import { WorkspacePort } from './workspace-port';
 const LISTEN_ALL_IPV4 = '0.0.0.0';
 const LISTEN_ALL_IPV6 = '::';
 const SERVER_REDIRECT_PATTERN = 'theia-redirect-';
+const PORT_EXCLUDE_ENV_VAR_PREFIX: string = 'PORT_PLUGIN_EXCLUDE_';
 
 // variables
 let workspacePorts: WorkspacePort[];
 let redirectPorts: WorkspacePort[];
 let redirectListeners: Map<number, BusyPort>;
+const excludedPorts: number[] = [];
+let outputChannel: theia.OutputChannel;
 
 // map a listener and the workspace port used
 export interface BusyPort {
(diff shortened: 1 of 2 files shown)


plugins/ports-plugin/src/ports-plugin.ts
```diff
@@ -151,13 +161,24 @@ function onClosedPort(port: Port) {
...
     workspacePorts = await workspaceHandler.getWorkspacePorts();
 
     redirectListeners = new Map<number, BusyPort>();
     redirectPorts = workspacePorts.filter(port => port.serverName.startsWith(SERVER_REDIRECT_PATTERN));
 
+    // initiate excluded ports
+    const excludedPortProperties: string[] = Object.keys(process.env).filter(key => key.startsWith(PORT_EXCLUDE_ENV_VAR_PREFIX));
+    excludedPortProperties.forEach(key => {
+        const value = process.env[key].toLocaleLowerCase() || '';
+        if (value !== 'no') {
```
AndrienkoAleksandr: Sorry, why do we need this check?

benoitf: hello @AndrienkoAleksandr 

it's because you can do
```
export PORT_PLUGIN_EXCLUDE_3130=YES
```

but
```
export PORT_PLUGIN_EXCLUDE_3130=NO
```
will not exclude the port

AndrienkoAleksandr: Oh, ok you want support ['yes', 'no'] values too. But in dockerfile you defined value in the boolean format:
> PORT_PLUGIN_EXCLUDE_3130=TRUE

How Do you want to cover false value?

benoitf: @AndrienkoAleksandr I added a fixup, is it ok for you ?

AndrienkoAleksandr: OK, absolutely.

PR #228, opened 2019-05-17
Added  preferences provider extension.

akurinnoy:

### What does this PR do?

Preferences Provider extension synchronizes Theia workspace settings with a preferences retrieved from a Che workspace configuration.

An extension preferences should be defined in the following way:

in `devfile`:

```yaml
- name: ${pluginName}
  type: chePlugin
  id: ${pluginId}
  preferences:
    ${propertyNameOne}: '${propertyValueOne}'
    ${propertyNameTwo}: '${propertyValueTwo}'
    ...
```

in `json` configuration:
```json
"attributes": {
  "plugin.${pluginId}.preference.${preferenceNameOne}": "${preferenceValueOne}",
  "plugin.${pluginId}.preference.${preferenceNameTwo}": "${preferenceValueTwo}"
  ...
}
```

resolve https://github.com/eclipse/che/issues/12992


extensions/eclipse-che-theia-prefs-provider/tests/prefs-provider-extension.test.ts
```diff
@@ -0,0 +1,202 @@
...
+
+        test('should retrieve preferences', async () => {
+            await prefsProvider.restorePluginProperties();
+
+            expect((<any>prefsProvider).setPluginProperties.mock.calls[0][0]).toEqual(prefsExpectation);
+        });
+
+    });
+
+});
```
AndrienkoAleksandr: New line, please.

extensions/eclipse-che-theia-prefs-provider/tsconfig.json
```diff
@@ -0,0 +1,30 @@
...
+            "dom"
+        ],
+        "sourceMap": true,
+        "rootDir": "src",
+        "outDir": "lib"
+    },
+    "include": [
+        "src"
+    ]
+}
```
AndrienkoAleksandr: Extends please from base config https://github.com/eclipse/che-theia/blob/master/extensions/eclipse-che-theia-user-preferences/tsconfig.json#L2

extensions/eclipse-che-theia-prefs-provider/jest.config.js
```diff
@@ -0,0 +1,29 @@
+/********************************************************************************
```
benoitf: hello, do we need this file ?
AFAIK until now all jest config was set in package.json files
https://github.com/eclipse/che-theia/blob/466ddd6ec38e3b73e5e6ce93b673ad4c3f316741/plugins/factory-plugin/package.json#L35-L56

extensions/eclipse-che-theia-prefs-provider/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@eclipse-che/theia-prefereneces-provider-extension",
```
benoitf: there is a typo on preferences (in the name)

extensions/eclipse-che-theia-prefs-provider/src/browser/prefs-provider-extension-frontend-module.ts
```diff
@@ -0,0 +1,25 @@
...
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { ContainerModule } from 'inversify';
+
+import { FrontendApplicationContribution } from '@theia/core/lib/browser';
+import { PrefsProvider } from './prefs-provider';
```
benoitf: I would opt for PreferencesProvider as name

extensions/eclipse-che-theia-prefs-provider/src/browser/prefs-provider.ts
```diff
@@ -0,0 +1,130 @@
...
+    PreferenceServiceImpl,
+    PreferenceScope,
+    FrontendApplicationContribution,
+    FrontendApplication
+} from '@theia/core/lib/browser';
+import { WorkspaceService } from '@theia/workspace/lib/browser';
+import WorkspaceClient, { IRemoteAPI } from '@eclipse-che/workspace-client';
+import { che } from '@eclipse-che/api';
+
+export interface WorkspaceWithDevfile extends che.workspace.Workspace {
```
benoitf: I'm curious, why do we need to extend che.workspace.Workspace ? devFile should already be there no ?

extensions/eclipse-che-theia-prefs-provider/package.json
benoitf: also I don't see where this extension is used ?

extensions/eclipse-che-theia-prefs-provider/src/browser/prefs-provider.ts
akurinnoy: I've just forgot to remove this. There was a type guard for a workspace with devfile, but later I got rid of it.

extensions/eclipse-che-theia-prefs-provider/jest.config.js
akurinnoy: hello, 
I'll move this config in package.json.

extensions/eclipse-che-theia-prefs-provider/src/browser/prefs-provider.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    private async setPluginProperties(props: [string, string][]): Promise<void> {
+        await this.workspaceService.roots;
+        const workspace = this.workspaceService.workspace;
+        if (!workspace) {
+            throw new Error('Failed to get Theia workspace.');
+        }
+        for (const [key, value] of props) {
+            await this.preferenceService.set(key, value, PreferenceScope.Workspace, workspace.uri);
```
tsmaeder: Shouldn't the scope here be "default"? Otherwise, how can the user override this preference?

extensions/eclipse-che-theia-prefs-provider/src/browser/prefs-provider.ts
```diff
@@ -0,0 +1,130 @@
...
+        });
+    }
+
+    async restorePluginProperties(): Promise<void> {
+        try {
+            const workspace = await this.getWorkspace();
+            const propsTuples = this.getPluginsProperties(workspace);
+            return this.setPluginProperties(propsTuples);
+        } catch (e) {
+            throw new Error('Preferences Provider Extension: ' + e.message);
```
tsmaeder: Why not just let the exception propagate here? This loses all information about where the error occurred.

akurinnoy: Good point, thanks, I'll fix

extensions/eclipse-che-theia-prefs-provider/src/browser/prefs-provider.ts
akurinnoy: Good catch!
Actually, default scope preference provider is not designed to set preferences ([preference-contribution.ts#L255](https://github.com/theia-ide/theia/blob/master/packages/core/src/browser/preferences/preference-contribution.ts#L255)). So "user" scope is the only option here, isn't it?

tsmaeder: Maybe an approach based on PreferenceProvider would work? I would take the discussion to the Theia chat, tbh.

akurinnoy: I'm not sure about overriding preferences in default scope, so let's consider other options first.

Setting recommended preferences in user scope is not an option here because user scope preferences are synchronizing between Che workspaces.

With workspace scope, we can make the extension working in the following way:
- set preferences if workspace scope doesn't contain them
- otherwise, ask user to override preferences with recommended ones.

What do you think?

AndrienkoAleksandr: > set preferences if workspace scope doesn't contain them

+1

tsmaeder: I still think the right thing to do would be to try an approach on PreferenceProvider. I'm sure Anton Kosyakov could provide feedback if that makes sense. Not an expert here.

evidolob:
ci-build

extensions/eclipse-che-theia-prefs-provider/src/browser/prefs-provider.ts
akurinnoy: I'm merging the PR as is, and [here](https://github.com/eclipse/che-theia/issues/243) is the issue to continue the discussion

PR #230, opened 2019-05-20
Cache Che commands

RomanNikitenko:
### What does this PR do?
Che commands are not modified in runtime at the moment, so I think we can cache them.

### DIFF
--- a/plugins/task-plugin/src/che-workspace-client.ts
+++ b/plugins/task-plugin/src/che-workspace-client.ts
@@ -16,6 +16,7 @@ const TERMINAL_SERVER_TYPE = 'terminal';
 
 @injectable()
 export class CheWorkspaceClient {
+    private cachedCommands: cheApi.workspace.Command[] = [];
 
     async getMachines(): Promise<{ [attrName: string]: cheApi.workspace.Machine }> {
         const workspace = await this.getCurrentWorkspace();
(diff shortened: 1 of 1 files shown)


sunix:
But we what if we want to be able to modify them at runtime eventually ?
What is the benefit at the moment of caching ?

plugins/task-plugin/src/che-workspace-client.ts
```diff
@@ -32,6 +33,10 @@ export class CheWorkspaceClient {
     }
 
     async getCommands(): Promise<cheApi.workspace.Command[]> {
+        if (this.cachedCommands.length > 0) {
+            return this.cachedCommands;
+        }
+
```
benoitf: I would avoid to cache it there.
Probably better to cache getCurrentWorkspace if config is not changed

RomanNikitenko: Agree, what about caching on che task provider level?

benoitf: cache is good but we need to invalidate cache when workspace configuration is changed

RomanNikitenko: We don't have events about changing workspace config, but workspace and plugins will be restarted if configuration is changed. Is it correct?

PR #242, opened 2019-05-23
Make sure clone path is always relative 

mmorhun:

### What does this PR do?
Make sure clone path is always relative and does not starts from slash as Che 6 project path.

https://github.com/eclipse/che/issues/13112

### DIFF
--- a/plugins/factory-plugin/src/workspace-projects-manager.ts
+++ b/plugins/factory-plugin/src/workspace-projects-manager.ts
@@ -145,8 +145,6 @@ export class DevfileProjectsManager extends WorkspaceProjectsManager {
             fileUri.convertToCheProjectPath(projectFolderURI, this.projectsRoot),
             projectUpstreamBranch.remoteURL,
             projectUpstreamBranch.branch);
-
-        if (workspace.devfile) { delete workspace.config; } // TODO delete this temporary workaround
     }
 
     deleteProject(workspace: cheApi.workspace.Workspace, projectFolderURI: string): void {
@@ -154,8 +152,6 @@ export class DevfileProjectsManager extends WorkspaceProjectsManager {
             workspace.devfile.projects,
             fileUri.convertToCheProjectPath(projectFolderURI, this.projectsRoot)
         );
-
-        if (workspace.devfile) { delete workspace.config; } // TODO delete this temporary workaround
     }
 
 }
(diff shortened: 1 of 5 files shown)


plugins/factory-plugin/src/workspace-projects-manager.ts
```diff
@@ -212,7 +208,7 @@ export class WorkspaceConfigProjectsManager extends WorkspaceProjectsManager {
 
         projectsHelper.deleteProjectFromWorkspaceConfig(
             workspace.config.projects,
-            fileUri.convertToCheProjectPath(projectFolderURI, this.projectsRoot)
+            '/' + fileUri.convertToCheProjectPath(projectFolderURI, this.projectsRoot)
```
AndrienkoAleksandr: Maybe  `path.sep` here instead of '/'?  But up to you.

plugins/factory-plugin/src/workspace-projects-manager.ts
```diff
@@ -195,7 +191,7 @@ export class WorkspaceConfigProjectsManager extends WorkspaceProjectsManager {
 
         projectsHelper.updateOrCreateGitProjectInWorkspaceConfig(
             workspace.config.projects,
-            fileUri.convertToCheProjectPath(projectFolderURI, this.projectsRoot),
+            '/' + fileUri.convertToCheProjectPath(projectFolderURI, this.projectsRoot),
```
AndrienkoAleksandr: Maybe  `path.sep` here instead of '/'?  But up to you.

plugins/factory-plugin/src/workspace-projects-manager.ts
mmorhun: According to [specification](https://github.com/eclipse/che/blob/e3281718a71db9f4de913d597dea0107f6c4b20f/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/devfile/Project.java#L23), only `/` is allowed.

plugins/factory-plugin/src/workspace-projects-manager.ts
mmorhun: The same as above

plugins/factory-plugin/tests/file-uri.spec.ts
```diff
@@ -26,12 +26,12 @@ describe("Test exec commands", () => {
...
 });
 
-describe("Testing convertion of project paths to be stored in the workspace config", () => {
-    test("Converting fs project path to che project path", async () => {
-        expect(convertToCheProjectPath('/projects/che-factory-extension/', '/projects')).toBe('/che-factory-extension');
-        expect(convertToCheProjectPath('/projects/che-factory-extension', '/projects')).toBe('/che-factory-extension');
-        expect(convertToCheProjectPath('/projects/che/che-factory-extension/', '/projects')).toBe('/che/che-factory-extension');
-        expect(convertToCheProjectPath('/projects/theiadev_projects/blog.sunix.org/', '/projects/theiadev_projects')).toBe('/blog.sunix.org');
-        expect(convertToCheProjectPath('/projects/che/che-factory-extension/', undefined)).toBe('/che/che-factory-extension');
+describe('Testing convertion of project paths to be stored in the workspace config', () => {
```
sunix: This is breaking projects cloning from workspace based on `workspace config`. The clonePath is for workspaces based on `devfile`.

PR #259, opened 2019-06-04
Fix build according to Theia refactoring

evidolob:
Update che-theia generator, and fix all compilation error


benoitf:
FYI ci job is not so happy

evidolob:
Unfortunately I don't found better (quicker) way than reseting changes in `che-theia` with
`RUN cd che/che-theia && git reset --hard`  
to restore build. We can discuss better way(che-theia repo reorganization) during F2F

PR #274, opened 2019-06-11
Run Che command in selected container if container doesn't specified in configuration

mmorhun:
#### Description

If container in which command should be run cannot be detected from config, run the command in selected container.
To reach that we need to send task modification which should be applied to persistent task configuration for only this one run and define the currently used container as a target container.

#### Resolves

https://github.com/eclipse/che-theia/issues/222

#### Depends on

https://github.com/theia-ide/theia/pull/5312

### DIFF
--- a/plugins/containers-plugin/src/containers-tree-data-provider.ts
+++ b/plugins/containers-plugin/src/containers-tree-data-provider.ts
@@ -11,6 +11,8 @@
 import * as theia from '@theia/plugin';
 import { IContainer } from './containers-service';
 
+const CHE_TASK_TYPE = 'che';
+
 interface ITreeNodeItem {
     id: string;
     name: string;
@@ -19,7 +21,7 @@ interface ITreeNodeItem {
     parentId?: string;
     command?: {
         id: string;
-        arguments?: string[]
+        arguments?: (string | Object)[]
     },
     isExpanded?: boolean;
 }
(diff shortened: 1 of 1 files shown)


plugins/containers-plugin/src/containers-tree-data-provider.ts
```diff
@@ -106,7 +108,7 @@ export class ContainersTreeDataProvider implements theia.TreeDataProvider<ITreeN
                         name: commandName,
                         tooltip: 'execute the command',
                         iconPath: 'fa-cogs medium-yellow',
-                        command: { id: 'task:run', arguments: ['che', commandName] }
+                        command: { id: 'task:run', arguments: [CHE_TASK_TYPE, commandName, this.overrideContainerName(container.name)] }
```
benoitf: hello, it couldn't be another field ? looks like we're now having complex types for command

vitaliy-guliy: I would not use dedicated method and constant.
Just inline
```
{
    type: 'che',
    target: {
        machineName: containerName
    }
}
```

AndrienkoAleksandr: @vitaliy-guliy, I think @mmorhun can not do that in a simple way. To do this stuff we need align this field in the che-machine-exec server side. @vitaliy-guliy create separated issue for that.

vitaliy-guliy:
@mmorhun could we wait until the https://github.com/eclipse/che-theia/pull/295 is merged and then will test the case again. I have a feeling that PR https://github.com/eclipse/che-theia/pull/295 will fix the problem.

mmorhun:
Closing this because the solution idea was changed due to changes in upstream.
There is a new PR that address the issue: https://github.com/eclipse/che-theia/pull/440

Issue #244, opened 2019-05-24
Display built-in plugins


vitaliy-guliy:
### Description
 

Display built-in (default) plugins in the list of installed plugins. We have to indicate somehow that plugin is default ( provided by Theia editor docker image ).
Default plugins are always present on the file system and cannot be removed.

### Reproduction Steps

**OS and version:**    

**Diagnostics:**

vitaliy-guliy:
I have already did some work on this issue and included plugins in the list. But there is still a lot of work on plugin decorations.
![Screenshot from 2019-05-24 10-30-59](https://user-images.githubusercontent.com/1655894/58310380-2d3b9700-7e0f-11e9-8846-225cabf98bb8.png)

sunix:
@vitaliy-guliy i don't think we want to display these plugins ... not critical for customers

benoitf:
I think it's useful to see them and to not have a way to install/uninstall them

it's like default apps on Android or Extensions in VS Code, there are there but you can't uninstall them

PR #276, opened 2019-06-12
Display list of built-in plugins

vitaliy-guliy:
### What does this PR do?
Makes it possible to view the list of built-in plugins

https://github.com/eclipse/che-theia/issues/244

Screenshot from 2019-06-12 16-00-38

### DIFF
--- a/extensions/eclipse-che-theia-plugin-ext/src/browser/plugin/che-plugin-frontend-service.ts
+++ b/extensions/eclipse-che-theia-plugin-ext/src/browser/plugin/che-plugin-frontend-service.ts
@@ -49,6 +49,8 @@ export class ChePluginFrontentService {
 
             const description = meta.source.description;
 
+            // Temporary disabled.
+            // We don't have an ability for now to display icons from the file system.
             // tslint:disable-next-line:no-any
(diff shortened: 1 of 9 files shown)


extensions/eclipse-che-theia-plugin-ext/src/browser/plugin/che-plugin-frontend-service.ts
```diff
@@ -25,14 +25,11 @@ export class ChePluginFrontentService {
...
-    async getDeployedPlugins(filter: string): Promise<ChePluginMetadata[]> {
-        if (PluginFilter.hasType(filter, '@installed')) {
-            let pluginList = await this.getAllDeployedPlugins();
-            pluginList = PluginFilter.filterPlugins(pluginList, filter);
-            return pluginList;
-        }
-
-        return [];
+    // returns a list of built-in plugins when filter contains '@builtin'
+    async getBuiltInPlugins(filter: string): Promise<ChePluginMetadata[]> {
```
AndrienkoAleksandr: Maybe method of the name should be more generic, because it woks not only for `builtin` filter value, I guess?

PR #279, opened 2019-06-14
Add dockerfile for Che Theia runtime image based on UBI 8

mmorhun:
#### Description

Add alterantive dockerfiles for `theia` and `theia-dev` which builds Che-Theia and Che-Theia-dev images based on `ubi8/nodejs-10` and `centos7` correspondingly.
The new dockerfiles are not included in any automated builds.

#### Related issues

Resolves partially https://github.com/eclipse/che/issues/13765

### DIFF
--- a/dockerfiles/theia-dev/centos7.Dockerfile
+++ b/dockerfiles/theia-dev/centos7.Dockerfile
@@ -1,0 +1,82 @@
+# Copyright (c) 2018-2019 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+
+FROM centos:7
+
+RUN \
+    # Add yum repository for nodejs 10.x
+    curl -sL https://rpm.nodesource.com/setup_10.x | bash - && \
+    yum install -y nodejs && \
+    # Add EPEL repository (jq)
+    yum install -y epel-release && \
+    yum install -y \
+    # To play with json in shell
+    jq \
+    # compile some javascript native stuff (node-gyp)
+    make gcc gcc-c++ python \
+    # clone repositories
+    git \
+    # Handle git diff properly
+    less \
+    # some lib to compile 'native-keymap' npm mpdule
+    libX11-devel libxkbfile-devel && \
+    yum clean all && \
+    # Install yarn
+    npm install -g yarn
+
+# Add npm global bin directory to the path
+ENV HOME=/home/theia-dev \
+    PATH=/home/theia-dev/.npm-global/bin:${PATH} \
+    # Specify the directory of git (avoid to search at init of Theia)
+    USE_LOCAL_GIT=true \
+    LOCAL_GIT_DIRECTORY=/usr \
+    GIT_EXEC_PATH=/usr/libexec/git-core \
+    THEIA_ELECTRON_SKIP_REPLACE_FFMPEG=true
+
+# Define package of the theia generator to use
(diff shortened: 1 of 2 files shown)


mmorhun:
Tested on openshift.io:
Theia-ubi-openshift io

dockerfiles/theia-dev/centos7.Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+
+FROM centos:7
```
benoitf: hello, where is UBI here ?
AFAIK it's not useful to have centos theia-dev

mmorhun: Unfortunately we cannot use UBI for builder image, because we lack some packages: `jq` and `libxkbfile-devel`

dockerfiles/theia/ubi8-nodejs10.Dockerfile
```diff
@@ -0,0 +1,152 @@
+# Copyright (c) 2018 Red Hat, Inc.
```
benoitf: it seems it's a big copy/paste difficult to maintain file

I would opt it for a given stage inside the existing Dockerfile

and just use one given target against that docker build
docker build --target <ubi>

dockerfiles/theia/ubi8-nodejs10.Dockerfile
```diff
@@ -0,0 +1,152 @@
...
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+
+###
+# Builder Image
+#
+# Should be based on CentOS
```
benoitf: I see no reason to have it based on centos

mmorhun: Thanks for pointing out. The comment is not clear enough. I meant, that if we build Che Theia on UBI we have to use CentOS based Che-Theia-dev

mmorhun: Otherwise Theia might not work

benoitf: I still don't see why

we're able to use debian in e2e images (which is not alpine based)
https://github.com/eclipse/che-theia/blob/master/dockerfiles/theia/e2e/Dockerfile#L10

dockerfiles/theia-dev/centos7.Dockerfile
benoitf: I mean, it's a dev container so ubi is not helpful there. It's more useful into production container (eclipse/che-theia image)

dockerfiles/theia/ubi8-nodejs10.Dockerfile
benoitf: we're doing typescript so it should be portable across all systems

mmorhun: No, unfortunately it isn't. We have some native code in node side dependencies. That's the reason why I've chosen CentOS but not Debian. Generally speaking, it should have the same base distro for builder and runner container (well, it may work, but no guarantee on that).

benoitf: @mmorhun what are these dependencies ? it's not the node-gyp dependencies ?
AFAIK we're rebuilding them there https://github.com/eclipse/che-theia/blob/master/dockerfiles/theia/e2e/Dockerfile#L31 for example

mmorhun: Thanks, we need to try

mmorhun:
Closing this, because we need better way to solve this task.

PR #284, opened 2019-06-16
Run `node-gyp install` before run `node-gyp rebuild`.

### LINKED ISSUE #283: Build failed sporadically
### Description

I got build failures on my CI server. This is reproduced sporadically. In some cases, failed. But it may succeeded even if same base image and same git HEAD commit.

The example of failure is [like this](https://dev.azure.com/pizzafactory/camino/_build/results?buildId=840).

This issue is not reported to `node-gyp` but I found the similar issue on [the another project](https://github.com/phusion/node-sha3/issues/42)

I believe this issue is not caused by Che-theia itself. So the goal is to find our workaround.

### Reproduction Steps

Just clean build with CI.

**Diagnostics:**

monaka:

### What does this PR do?

Runs `node-gyp install` before running `node-gyp rebuild`.
Referring to logs on my CI builds, node-gyp may start builds before extracting required header files.
As the issue #283 is less reproductive, I set this PR to draft for now.
I'm going to check this by some more CI builds.

fixes #283

### DIFF
--- a/dockerfiles/theia-dev/Dockerfile
+++ b/dockerfiles/theia-dev/Dockerfile
@@ -49,6 +49,8 @@ EXPOSE 3000 3030
 # Configure npm and yarn to use home folder for global dependencies
 RUN npm config set prefix "${HOME}/.npm-global" && \
     echo "--global-folder \"${HOME}/.yarn-global\"" > ${HOME}/.yarnrc && \
+    # Setup node-gyp. Yarn will be failed unless installing required files here.
+    yarn global add node-gyp && \
     # add eclipse che theia generator
     yarn global add yo @theia/generator-plugin@0.0.1-1540209403 ${THEIA_GENERATOR_PACKAGE} && \
     # Generate .passwd.template \
--- a/dockerfiles/theia/Dockerfile
+++ b/dockerfiles/theia/Dockerfile
@@ -72,7 +72,7 @@ RUN che:theia init -c ${HOME}/che-theia-init-sources.yml
 RUN che:theia cdn --theia="${CDN_PREFIX}" --monaco="${MONACO_CDN_PREFIX}"
 
 # Compile Theia
-RUN yarn
+RUN yarn global add node-gyp && node-gyp install && yarn
 
 # Run into production mode
 RUN che:theia production
--- a/dockerfiles/remote-plugin-dotnet-2.2.105/Dockerfile
+++ b/dockerfiles/remote-plugin-dotnet-2.2.105/Dockerfile
@@ -54,7 +54,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
     # Install typescript@2.9.2 and node-gyp
 
-    &&  yarn global add typescript@2.9.2 node-gyp
+    &&  yarn global add typescript@2.9.2 node-gyp \
+    &&  node-gyp install
 
 ENV HOME=/home/theia
 COPY --from=endpoint /home/theia /home/theia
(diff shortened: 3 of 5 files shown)


monaka:
See also: https://github.com/nodejs/node-gyp/issues/1780
https://github.com/nodejs/node-gyp/issues/1054
https://github.com/yarnpkg/yarn/issues/1874

monaka:
https://dev.azure.com/pizzafactory/camino/_build/results?buildId=873

dockerfiles/theia/Dockerfile
```diff
@@ -72,7 +72,7 @@ RUN che:theia init -c ${HOME}/che-theia-init-sources.yml
 RUN che:theia cdn --theia="${CDN_PREFIX}" --monaco="${MONACO_CDN_PREFIX}"
 
 # Compile Theia
-RUN yarn
+RUN yarn global add node-gyp && node-gyp install & yarn
```
benoitf: why adding node-gyp there ? it's not used

dockerfiles/theia-dev/Dockerfile
```diff
@@ -49,6 +49,8 @@ EXPOSE 3000 3030
 # Configure npm and yarn to use home folder for global dependencies
 RUN npm config set prefix "${HOME}/.npm-global" && \
     echo "--global-folder \"${HOME}/.yarn-global\"" > ${HOME}/.yarnrc && \
+    # Setup node-gyp. Yarn will be failed unless installing required files here.
```
benoitf: AFAIK there is no need to install node-gyp there

dockerfiles/theia/Dockerfile
monaka:
```
RUN yarn global add node-gyp && node-gyp install & yarn
```

is same as

```
RUN yarn global add node-gyp && node-gyp install
RUN yarn
```

It's safer to call `node-gyp install` before all `yarn` calls.

monaka: ... and I found a typo there. I'll fix it and re-push.

dockerfiles/theia-dev/Dockerfile
monaka: Yes, here is a considerable point.
As you say, there is no `yarn` call in `theia-dev/Dockerfile`.
But e2e uses `theia-dev` as the base image and calls `yarn`.

And end users that use this image will call `yarn` in the container.
They may encounter build failures.

Just IMO but it's better to preset header files provided node-gyp.

benoitf:
is that calling `node-gyp configure` before `node-gyp build` is not enough ?

monaka:
@benoitf Referring to issues on node-gyp and yarn, the root issue is that more than two packages try to download/extract node development files at the same. So `node-gyp configure` won't fix this.

dockerfiles/theia-dev/Dockerfile
monaka: I moved `node-gyp install` into `entrypoint.sh` because the home directory will be moved on non-root environments.
The new patch has already pushed.

`node-gyp install` deploys files into `${HOME}/.node-gyp`.
And, as you know, `${HOME}` is ephemeral directory.
So it should be called in every bootstrap.

monaka:
@benoitf Looks organize this issue.

The link is a typical log when the build was failed.

https://dev.azure.com/pizzafactory/camino/_build/results?buildId=840&view=logs&jobId=ca395085-040a-526b-2ce8-bdc85f692774&taskId=c310a698-906e-53a5-78f7-614142e7c740&lineStart=243&lineEnd=247&colStart=1&colEnd=4

The file `/home/theia-dev/.node-gyp/10.15.3/include/node/common.gypi` was broken.
This situation will be caused like below:

0. (node-gyp calls `node-gyp install` implicitly when `node-gyp rebuild` is called and `~/.node-gyp/{node-version}` is empty.) 

1. Yarn spawns multiple `node-gyp rebuild` at the same time.
2. builders try `node-gyp install` in asynchronous with no lock. After done `install`, each builder will do `rebuild`.
3. When one builder executing `rebuild` will refer content in `~/.node-gyp`, the another builder may overwrite content in `~/.node-gyp`.
4. Of course, the build will be failed if content was broken by the another builder.

`configure` and `rebuild` will touch content in `node_module/{module-name}`.
Each `node-gyp` builder don't touch another modules. So they commands don't cause this issue.

Touching `~/.node-gyp/{node-version}` by `install` is the single pain point.

monaka:
https://dev.azure.com/pizzafactory/camino/_build/results?buildId=901 The build is passed on my CI.

monaka:
After tons of try and error, I noticed the another approach. I reject this and will open the another PR later.

PR #290, opened 2019-06-18
Fix connection failure in the e2e test for `che-theia`

monaka:
### What does this PR do?

Some improvements to the e2e test for `che-theia`
The current test script wait booting Theia up by `sleep`.
But we can't estimate how long. So it's better to watch the status of TCP port.

None.

### DIFF
--- a/dockerfiles/theia/e2e/Dockerfile
+++ b/dockerfiles/theia/e2e/Dockerfile
@@ -15,7 +15,7 @@ ENV NOCDN=true
 
 RUN printf "deb http://archive.debian.org/debian/ jessie main\ndeb-src http://archive.debian.org/debian/ jessie main\ndeb http://security.debian.org jessie/updates main\ndeb-src http://security.debian.org jessie/updates main" > /etc/apt/sources.list
 RUN apt-get update && \
-    apt-get install -y libx11-dev libxkbfile-dev sudo
+    apt-get install -y libx11-dev libxkbfile-dev sudo iproute2
 CMD /root/docker-run.sh
 RUN yarn global add typescript@2.9.2 node-gyp
 
--- a/dockerfiles/theia/e2e/src/docker-run.sh
+++ b/dockerfiles/theia/e2e/src/docker-run.sh
@@ -6,13 +6,32 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 
+# (5 + 2 * 30 = 65 seconds) is the default timeout.
+: "${WAIT_COUNT:=30}"
+
 echo "Starting Theia..."
 rm -rf /root/logs/*
 HOME=/home/theia /entrypoint.sh > /root/logs/theia.log 2>/root/logs/theia-error.log&
-sleep 10s
+
 echo "Cleaning videos folder..."
 # Cleanup previous videos
-rm -rf /root/cypress/videos
+rm -rf /root/cypress/videos/*
+
+# Find TCP 0.0.0.0:3100 that will be opened by Theia.
+sleep 5s
+while [ $WAIT_COUNT -gt 0 ]; do
+    # Check the listening port
+    ss -nlt | grep -Fq ':3100' && break
+    # not found
+    WAIT_COUNT=$((WAIT_COUNT-1));
+    echo "Waiting for booting up Theia..."
+    sleep 2s
+done
+
+if [ $WAIT_COUNT -eq 0 ]; then
+    echo "Timeout. Theia is dead?"
+    exit 1
+fi
 
 # Run tests
 echo "Run the tests"


monaka:
BTW, The CI server is set up well?

Screenshot 2019-06-18 at 11 13 53

monaka:
https://dev.azure.com/pizzafactory/camino/_build/results?buildId=872

benoitf:
@monaka the CI is failing as it's unable to build from user branches :-/
I think we need to be logged in to have access to these build results

dockerfiles/theia/e2e/src/docker-run.sh
```diff
@@ -9,10 +9,16 @@
...
+
 echo "Cleaning videos folder..."
 # Cleanup previous videos
-rm -rf /root/cypress/videos
+rm -rf /root/cypress/videos/*
+
+# Find TCP 0.0.0.0:3100 that will be opened by Theia.
+until `ss -nlt | grep -Fq ':3100'`; do
+    echo "Waiting for booting up Theia..."
+    sleep 10s
```
benoitf: I think the wait increment is too long no ?

dockerfiles/theia/e2e/Dockerfile
```diff
@@ -15,7 +15,7 @@ ENV NOCDN=true
 
 RUN printf "deb http://archive.debian.org/debian/ jessie main\ndeb-src http://archive.debian.org/debian/ jessie main\ndeb http://security.debian.org jessie/updates main\ndeb-src http://security.debian.org jessie/updates main" > /etc/apt/sources.list
 RUN apt-get update && \
-    apt-get install -y libx11-dev libxkbfile-dev
+    apt-get install -y libx11-dev libxkbfile-dev sudo iproute2
```
benoitf: why sudo? AFAIK it's not used

dockerfiles/theia/e2e/src/docker-run.sh
monaka: It requires 40 minutes to run each CI build (on my server. Ubuntu 16.10 / Azure D2 VM).
IMO, it won't effective to check more frequently.

dockerfiles/theia/e2e/Dockerfile
monaka: e2e image copies /entrypoint.sh from `theia` image. `sudo` is used there.

https://github.com/PizzaFactory/che-theia/blob/4f92c61fde7b9c09757fcd83cec40d9694b14ed2/dockerfiles/theia/src/entrypoint.sh#L33-L34

The script checks if `sudo` is usable. So it won't make bad effects to Openshift based instances.

monaka: And it's better to apply https://github.com/eclipse/che-theia/pull/292 after merged this patch.
`che-theia` should be checked to run both root and non-root env.

dockerfiles/theia/e2e/src/docker-run.sh
benoitf: If you already have the image pre-built, and that you launch the script it should be fast. Here it's waiting 10s on each increment which is too long as you enhance the script to not wait so long.
IMHO should be around 1-2s

Also I think we should add a MAX_WAIT parameter to avoid infinite loop (let say theia is failing then it won't start then it won't listen and e2e will wait for eternity)

monaka: I'm easy about the value of interval.
Since it's hard to boot Theia up within 1 or 2 second, I'll fix to 5 sec as the initial and 2 second in the loop.

I agree the limit of the loop is required.

dockerfiles/theia/e2e/Dockerfile
benoitf: I would separate the PRs

One for the 'short wait time' upon boot
one around permissions which is partially handled in another PR

dockerfiles/theia/e2e/src/docker-run.sh
```diff
@@ -6,13 +6,32 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 
+# (5 + 2 * 30 = 65 seconds) is the default timeout.
+: ${WAIT_COUNT:=30}
```
benoitf: FYI there is warning on https://www.shellcheck.net/

Line 10:
: ${WAIT_COUNT:=30}
  ^-- SC2223: This default assignment may cause DoS due to globbing. Quote it.
 
Line 26:
    WAIT_COUNT=$(($WAIT_COUNT-1));
                  ^-- SC2004: $/${} is unnecessary on arithmetic variables.

dockerfiles/theia/e2e/Dockerfile
monaka: I see. I left the timeout issue here. And the sudo issue is moved to #293

dockerfiles/theia/e2e/src/docker-run.sh
monaka: Hmm... At least `$(())` is needed, AFAIK...

```
$ sh -c 'a=3; a=$(($a-1)); echo $a'
2
$ sh -c 'a=3; a=$($a-1); echo $a'
sh: 1: 3-1: not found
$ sh -c 'a=3; a=(($a-1)); echo $a'
sh: 1: Syntax error: "(" unexpected
```

benoitf:
```
 sh -c 'a=3; a=$((a-1)); echo $a'
```

(no $ sign before a)

benoitf: https://github.com/koalaman/shellcheck/wiki/SC2004

monaka: Thanks for your information. I see and pushed some fixes. I'm going to check my fixes by my CI server.

monaka:
The build passed on my CI server. https://dev.azure.com/pizzafactory/camino/_build/results?buildId=881

PR #291, opened 2019-06-18
Use `-z` for empty checks.

monaka:
### What does this PR do?

Fix the error in case lack of parameters.

```
$ sh -c '[ "${image}" == "" ] && echo $?'
sh: 1: [: unexpected operator
```

```
$ sh -c '[ -z "${image}" ] && echo $?'
0
```

None.

### DIFF
--- a/dockerfiles/theia/extract-for-cdn.sh
+++ b/dockerfiles/theia/extract-for-cdn.sh
@@ -13,7 +13,7 @@ help() {
 
 image="$1"
 destination="$2"
-if [ "${image}" == "" -o "${destination}" == "" ]; then
+if [ -z "${image}" -o -z "${destination}" ]; then
   help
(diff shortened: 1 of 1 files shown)


monaka:
https://dev.azure.com/pizzafactory/camino/_build/results?buildId=871

benoitf:
Hello, I'm curious to know in which case it's empty

the only call is there 
https://github.com/eclipse/che-theia/blob/983351d5ac7545fd76fe035d35131fc23cc0b9e8/dockerfiles/theia/build.sh#L44

you're not using the provided build.sh ?

monaka:
@benoitf Yes this issue is not realized on the normal build path. And the code is for the protection for abnormal calls. So it should be worked well with unexpected arguments, IMO.

PR #301, opened 2019-06-24
[node] Use version 10.16.

monaka:

### What does this PR do?

Upgrades to newer `node`.

None.


monaka:
I checked to build this and done with no error. ( #302 #303 also required )

PR #303, opened 2019-06-24
Set executable bits to `build.sh`.

monaka:

### What does this PR do?

Sets executable bits to `build.sh`.

None.


benoitf:
@monaka do you think it's possible to push to che-theia repository your branches so the CI would work ?

Issue #281, opened 2019-06-16
Exception on the  non-root environment.


monaka:
### Description

It looks the-theia can't access `/proc/net/tcp6` in the instance running as non-root.

### Reproduction Steps

1. Run as the non-root instance
2. Start workspace
3. Check logs from `che-theia` container.

**OS and version:**    

7.0.0-RC-2.0-SNAPSHOT

**Diagnostics:** 

```
root ERROR [hosted-plugin: 54] (node:54) UnhandledPromiseRejectionWarning: Error: Unable to execut
e the command cat /proc/net/tcp6: Error: Command failed: cat /proc/net/tcp6
cat: can't open '/proc/net/tcp6': No such file or directory

    at new CliError (/tmp/theia-unpacked/eclipse_che_ports_plugin.theia/lib/cli-error.js:32:28)
    at /tmp/theia-unpacked/eclipse_che_ports_plugin.theia/lib/command.js:69:40
    at ChildProcess.exithandler (child_process.js:301:5)
    at ChildProcess.emit (events.js:189:13)
    at maybeClose (internal/child_process.js:970:16)
    at Socket.stream.socket.on (internal/child_process.js:389:11)
    at Socket.emit (events.js:189:13)
    at Pipe._handle.close (net.js:597:12)
```

monaka:
I misunderstood about this issue. Indeed there have some issues related on root/non-root environment.
But this `/proc/net/tcp6` issue is an individual.  

In some environments (at least Azure AKS), IPv6 is disable and `/proc/net/tcp6` isn't created.

```
$ kubectl exec -it --namespace=banco workspacewvws39a5loq3c0oa.shell-5556494f8c-tkkdg ls /proc/net
Defaulting container name to shell.
Use 'kubectl describe pod/workspacewvws39a5loq3c0oa.shell-5556494f8c-tkkdg -n banco' to see all of the containers in this pod.
arp                 ip6_tables_names    netfilter  route         tcp
dev                 ip6_tables_targets  netlink    rpc           udp
dev_mcast           ip_mr_cache         netstat    rt_acct       udplite
fib_trie            ip_mr_vif           packet     rt_cache      unix
fib_triestat        ip_tables_matches   protocols  snmp          wireless
icmp                ip_tables_names     psched     sockstat      xfrm_stat
igmp                ip_tables_targets   ptype      softnet_stat
ip6_tables_matches  mcfilter            raw        stat
```

PR #315, opened 2019-07-01
Fix unhandled exceptions in port-scanner. (eclipse#281)

monaka:

### What does this PR do?

Fixes a bug in ports-plugin caused on specific K8s clusters. 

In case IPv6 was disabled, `/proc/net/tcp6` will not exist.
(Probably IPv4 will be similar.)

#281


monaka:
I checked this patch on my environment in Azure AKS. It works fine.

plugins/ports-plugin/src/port-scanner.ts
```diff
@@ -29,8 +29,12 @@ export class PortScanner {
         const ipConverter = new IpConverter();
         // connect to /proc/net/tcp and /proc/net/tcp6
         const command = new Command(__dirname);
-        const outputv4 = await command.exec(PortScanner.GRAB_PORTS_IPV4);
-        const outputv6 = await command.exec(PortScanner.GRAB_PORTS_IPV6);
+        const outputv4 = await command
```
benoitf: hello, wouldn't it be better to check if file is there instead of running the command and failing and display the error as it will be displayed tons of time

monaka: Hmm. An another side question. What is the reason why `fs` package isn't used here?  There is a room for consideration to stop using `command` if my current patch should be more integrated.

monaka:
@benoitf I pushed the patch based on [my comment](https://github.com/eclipse/che-theia/pull/315#discussion_r300169955) above.
Are there special reasons to call `cat`?
Generally it's better if scripts were no dependency to any external commands, IMO.

benoitf:
@monaka you're right, no reason to use cat at all and I agree with you on using fs

plugins/ports-plugin/src/port-scanner.ts
monaka: Code was totally refactored. So I mark this tread as resolved.

monaka:
@benoitf Thanks for your comment. I made new patch using `fs` package and force-pushed here. I tested it by Jest in `plugins/ports-plugin`.

musienko-maxim:

plugins/ports-plugin/package.json
```diff
@@ -16,6 +16,7 @@
     "@eclipse-che/plugin": "latest",
     "@theia/plugin": "latest",
     "@theia/plugin-packager": "latest",
+    "@types/jest": "24.0.3",
```
olexii4: We are using "jest": "22.1.4" for the factory-plugin.  
https://github.com/eclipse/che-theia/blob/master/plugins/factory-plugin/package.json#L17
Let's use the same version.

monaka:
@benoitf I pushed some more fixes here and I believe it becomes reviewable quality.

sunix:
@monaka, @evidolob is going to review this PR. Sorry for delay :)

evidolob:
This PR also should fix https://github.com/eclipse/che/issues/13620

PR #318, opened 2019-07-01
Fix preview URL panel to show the resolved URL instead of a server name.

akurinnoy:
### What does this PR do?

Currently, Preview URL panel's view is a bit confusing because it doesn't show any URL at all. This PR modifies it in the way to replace a server name with a resolved URL. In same way it updates a task preview notification which appears on top of Theia's page once a task is started.

#### Screenshots

 
Screenshot 2019-07-01 16 42 49

Screenshot 2019-07-01 16 44 39

fix https://github.com/eclipse/che/issues/13357

### DIFF
--- a/plugins/task-plugin/src/preview/previews-widget.ts
+++ b/plugins/task-plugin/src/preview/previews-widget.ts
@@ -84,8 +84,8 @@ export class PreviewUrlsWidget {
 
                     const server = `<a class="task-link" href="${url}" target="_blank">${url}</a>`;
                     const taskLabel = `<label>${cheTask.name}</label>`;
-                    const previewButton = `<button class='button' type="button" onclick="preview('${INTERNALLY_CHOICE}', '${previewUrl}')">${PREVIEW_BUTTON_NAME}</button>`;
-                    const goToButton = `<button class='button' type="button" onclick="preview('${EXTERNALLY_CHOICE}', '${previewUrl}')">${GO_TO_BUTTON_NAME}</button>`;
+                    const previewButton = `<button class='button' type="button" onclick="preview('${INTERNALLY_CHOICE}', '${url}')">${PREVIEW_BUTTON_NAME}</button>`;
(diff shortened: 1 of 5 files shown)


plugins/task-plugin/src/preview/previews-widget.ts
```diff
@@ -62,26 +66,31 @@ export class PreviewUrlsWidget {
...
+        return await Promise.all(
+            this.options.tasks.map(
+                async (cheTask: theia.Task) => {
+                    const previewUrl = cheTask.definition.previewUrl;
+                    const url = await this.previewUrlOpenService.resolve(previewUrl);
+
+                    const server = `<a class="task-link" href="${url}" target="_blank">${url}</a>`;
+                    const taskLabel = `<label>${cheTask.name}</label>`;
+                    const previewButton = `<button class='button' type="button" onclick="preview('${INTERNALLY_CHOICE}', '${previewUrl}')">${PREVIEW_BUTTON_NAME}</button>`;
+                    const goToButton = `<button class='button' type="button" onclick="preview('${EXTERNALLY_CHOICE}', '${previewUrl}')">${GO_TO_BUTTON_NAME}</button>`;
```
RomanNikitenko: If I understand  correctly, at this step you already have a resolved url, but unresolved url is passed to 'preview' function.
Do we have some reasons for it?
Maybe it makes sense to pass the resolved url, otherwise a url is resolved twice: [at rendering](https://github.com/eclipse/che-theia/pull/318/files#diff-76b3d37b7996f3b0708dbfc19bf23f41R83) and [when user click on 'Preview' or 'Go to' button](https://github.com/eclipse/che-theia/pull/318/files#diff-51367edb0f101b505665cc8cdd29af24R48) 
WDYT?

akurinnoy: Thank you @RomanNikitenko for pointing that out. That makes sense, I'll fix it.

PR #324, opened 2019-07-03
Updating codeowners for Che 7 endgame code reviews

l0rd:
Adding @l0rd @rhopp as global code reviewers as required by eclipse/che#13637


benoitf:
ci-build

PR #333, opened 2019-07-08
Fix plugins view

vitaliy-guliy:
### What does this PR do?

Should fix problem with Che Plugins view which sometimes appeared from time to time and difficult to debug. When I tried to find for the message, I received only one place in the code from which it could be printed.

https://github.com/theia-ide/theia/blob/master/packages/core/src/common/messaging/proxy-factory.ts#L168

The error occurred when calling `getWorkspacePlugins` method and it's not clear whether something wrong with the method or with the accessing `getWorkspacePlugins` method of ChePluginService through RPC.

This PR
 - aligns binding of `ChePluginService` with other services
 - checks the workspace configuration when getting a list of installed plugins

https://github.com/eclipse/che/issues/13747
https://github.com/eclipse/che/issues/13669

### DIFF
--- a/extensions/eclipse-che-theia-plugin-ext/src/node/che-backend-module.ts
+++ b/extensions/eclipse-che-theia-plugin-ext/src/node/che-backend-module.ts
@@ -51,7 +51,8 @@ export default new ContainerModule(bind => {
         })
     ).inSingletonScope();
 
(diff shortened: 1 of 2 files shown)


vitaliy-guliy:
Is not actual anymore

PR #334, opened 2019-07-08
Fix synchronization of the existing projects

olexii4:

### What does this PR do?
Add new tests for the project synchronization. Fix implementation.

https://github.com/eclipse/che/issues/13621


sunix:
@olexii4 I have tested and made few changes:
- fix one issue with the clonePath that shouldn't be here if it is the name == clonePath
- remove all the global variable from the test. Problem with these variables is that they make the test not easy to read for a human.

PR #339, opened 2019-07-10
Use native git instead of git plugin in the factory plugin

vinokurig:
### What does this PR do?
Use native git instead of git plugin in the factory plugin

fixes https://github.com/eclipse/che/issues/13796


sunix:
Hey, I don't know why and if there something wrong else where but i got this issue with `clonePath` not being taken in account when it is defined.
Selection_335

sunix:
@vinokurig I pushed the fix in this branch

PR #349, opened 2019-07-16
Change docker images version to '7.0.0-next'

evidolob:
### What does this PR do?
Changes version of che-theia docker images to `7.0.0-next` as we discuss there: https://github.com/eclipse/che/issues/13782
che-theia docker images updated on each commit in `7.0.0` branch

https://github.com/eclipse/che/issues/13782


benoitf:
IMHO it would be better to have channels than versions but it seems I'm the only one to like that

evidolob:
Problem with channels is that we don't know on which exactly version user have a problem, 
it want to be a problem if we switch all che to channels instead of fixed versions

benoitf:
@evidolob if you go inside theia help/about menu it displays each sha1 of upstream theia and che-theia used so user can report the exact version he's using (or some tools)

PR #352, opened 2019-07-16
Store SSH private key on the file system from SSH plugin

vinokurig:
### What does this PR do?
Store SSH private key file on the file system when key is generated from the SSH Plugin

fixes https://github.com/eclipse/che/issues/13866

### DIFF
--- a/plugins/ssh-plugin/src/ssh-plugin-backend.ts
+++ b/plugins/ssh-plugin/src/ssh-plugin-backend.ts
@@ -46,6 +46,10 @@ export async function start() {
         viewPublicKey(sshKeyManager);
     });
 }
+const getHostName = async function (): Promise<string> {
+    const hostName = await theia.window.showInputBox({ placeHolder: 'Please provide a Host name e.g. github.com' });
+    return hostName ? hostName : '';
+};
 
 const getSshServiceName = async function (): Promise<string> {
     /**
@@ -83,7 +87,7 @@ const getSshServiceName = async function (): Promise<string> {
     }
 };
 
-const updateConfig = function (host: string): void {
+const updateConfig = function (hostName: string): void {
     const sshDir = os.homedir() + '/.ssh';
     const configFile = sshDir + '/config';
     if (!fs.existsSync(configFile)) {
@@ -93,7 +97,7 @@ const updateConfig = function (host: string): void {
         fs.appendFileSync(configFile, '');
         fs.chmodSync(configFile, '644');
     }
-    const keyConfig = `\nHost ${host}\nHostName ${host}\nIdentityFile ${sshDir}/${host.replace('.', '_')}\n`;
+    const keyConfig = `\nHost ${hostName}\nHostName ${hostName}\nIdentityFile ${sshDir}/${hostName.replace('.', '_')}\n`;
     const configContent = fs.readFileSync(configFile).toString();
     if (configContent.indexOf(keyConfig) >= 0) {
         const newConfigContent = configContent.replace(keyConfig, '');
@@ -110,12 +114,12 @@ const writeKey = function (name: string, key: string): void {
 };
 
 const generateKeyPair = async function (sshkeyManager: SshKeyManager): Promise<void> {
-    const keyName = await theia.window.showInputBox({ placeHolder: 'Please provide a key pair name (Host name e.g. github.com)' });
-    const key = await sshkeyManager.generate(await getSshServiceName(), keyName ? keyName : '');
-    updateConfig(keyName);
-    writeKey(keyName, key.privateKey);
+    const hostName = await getHostName();
+    const key = await sshkeyManager.generate(await getSshServiceName(), hostName);
+    updateConfig(hostName);
+    writeKey(hostName, key.privateKey);
     const viewAction = 'View';
-    const action = await theia.window.showInformationMessage('Do you want to view the generated public key?', viewAction);
+    const action = await theia.window.showInformationMessage(`Key for ${hostName} successfully generated, do you want to view the public key?`, viewAction);
     if (action === viewAction && key.privateKey) {
         const document = await theia.workspace.openTextDocument({ content: key.publicKey });
         await theia.window.showTextDocument(document);
@@ -123,17 +127,17 @@ const generateKeyPair = async function (sshkeyManager: SshKeyManager): Promise<v
 };
 
 const createKeyPair = async function (sshkeyManager: SshKeyManager): Promise<void> {
-    const keyName = await theia.window.showInputBox({ placeHolder: 'Please provide a key pair name' });
+    const hostName = await getHostName();
     const publicKey = await theia.window.showInputBox({ placeHolder: 'Enter public key' });
     const privateKey = await theia.window.showInputBox({ placeHolder: 'Enter private key' });
     const service = await getSshServiceName();
 
     await sshkeyManager
-        .create({ name: keyName ? keyName : '', service, publicKey: publicKey, privateKey })
+        .create({ name: hostName, service, publicKey: publicKey, privateKey })
         .then(async () => {
-            theia.window.showInformationMessage('Key "' + `${keyName}` + '" successfully created');
-            updateConfig(keyName);
-            writeKey(keyName, privateKey);
+            theia.window.showInformationMessage('Key "' + `${hostName}` + '" successfully created');
+            updateConfig(hostName);
+            writeKey(hostName, privateKey);
         })
         .catch(error => {
             theia.window.showErrorMessage(error);
--- a/plugins/ssh-plugin/src/ssh-plugin-backend.ts
+++ b/plugins/ssh-plugin/src/ssh-plugin-backend.ts
@@ -9,13 +9,17 @@
  **********************************************************************/
 
 import * as theia from '@theia/plugin';
-import { RemoteSshKeyManager, SshKeyManager, CheService } from './node/ssh-key-manager';
+import { RemoteSshKeyManager, SshKeyManager } from './node/ssh-key-manager';
 import { che as cheApi } from '@eclipse-che/api';
 import * as fs from 'fs';
 import * as os from 'os';
 
 export async function start() {
     const sshKeyManager = new RemoteSshKeyManager();
+    const GENERATE_FOR_HOST: theia.CommandDescription = {
+        id: 'ssh:generate_for_host',
+        label: 'SSH: generate key pair for particular host...'
+    };
     const GENERATE: theia.CommandDescription = {
         id: 'ssh:generate',
         label: 'SSH: generate key pair...'
@@ -33,6 +37,9 @@ export async function start() {
         label: 'SSH: view public key...'
     };
 
+    theia.commands.registerCommand(GENERATE_FOR_HOST, () => {
+        generateKeyPairForHost(sshKeyManager);
+    });
     theia.commands.registerCommand(GENERATE, () => {
         generateKeyPair(sshKeyManager);
     });
@@ -46,45 +53,10 @@ export async function start() {
         viewPublicKey(sshKeyManager);
     });
 }
+
 const getHostName = async function (): Promise<string> {
     const hostName = await theia.window.showInputBox({ placeHolder: 'Please provide a Host name e.g. github.com' });
     return hostName ? hostName : '';
-};
-
-const getSshServiceName = async function (): Promise<string> {
-    /**
-     * Known Che services which can use the SSH key pairs.
-     */
-    const services: CheService[] = [
-        {
-            name: 'vcs',
-            displayName: 'VCS',
-            description: 'SSH keys used by Che VCS plugins'
-        },
-        {
-            name: 'machine',
-            displayName: 'Workspace Containers',
-            description: 'SSH keys injected into all Workspace Containers'
-        }
-    ];
-
-    const option: theia.QuickPickOptions = {
-        matchOnDescription: true,
-        matchOnDetail: true,
-        canPickMany: false,
-        placeHolder: 'Select object:'
-    };
-    const sshServiceValue = await theia.window.showQuickPick<theia.QuickPickItem>(services.map(service =>
-        ({
-            label: service.name,
-            description: service.description,
-            detail: service.displayName
-        })), option);
-    if (sshServiceValue) {
-        return Promise.resolve(sshServiceValue.label);
-    } else {
-        return Promise.resolve('');
-    }
 };
 
 const updateConfig = function (hostName: string): void {
@@ -97,7 +69,7 @@ const updateConfig = function (hostName: string): void {
         fs.appendFileSync(configFile, '');
         fs.chmodSync(configFile, '644');
     }
-    const keyConfig = `\nHost ${hostName}\nHostName ${hostName}\nIdentityFile ${sshDir}/${hostName.replace('.', '_')}\n`;
+    const keyConfig = `\nHost ${hostName.startsWith('default-') ? '*' : hostName}\nIdentityFile ${sshDir}/${hostName.replace('.', '_')}\n`;
     const configContent = fs.readFileSync(configFile).toString();
     if (configContent.indexOf(keyConfig) >= 0) {
         const newConfigContent = configContent.replace(keyConfig, '');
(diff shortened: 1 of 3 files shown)


sunix:
@vinokurig Could you describe the end user flow that would use that ?
How can I test it ? how can I set ssh keys ?

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -82,24 +83,57 @@ const getSshServiceName = async function (): Promise<string> {
...
+
+const writeKey = function (name: string, key: string): void {
+    const keyFile = os.homedir() + '/.ssh/' + name.replace('.', '_');
+    fs.appendFileSync(keyFile, key);
+    fs.chmodSync(keyFile, '600');
+};
+
 const generateKeyPair = async function (sshkeyManager: SshKeyManager): Promise<void> {
-    const keyName = await theia.window.showInputBox({ placeHolder: 'Please provide a key pair name' });
+    const keyName = await theia.window.showInputBox({ placeHolder: 'Please provide a key pair name (Host name e.g. github.com)' });
```
l0rd: Are we prompting the user for a key pair name or for an host name?

l0rd: And do we really need to prompt the user for this? Can't we use a default name?

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -82,24 +83,57 @@ const getSshServiceName = async function (): Promise<string> {
...
     const key = await sshkeyManager.generate(await getSshServiceName(), keyName ? keyName : '');
+    updateConfig(keyName);
+    writeKey(keyName, key.privateKey);
     const viewAction = 'View';
-    const action = await theia.window.showInformationMessage('Do you want to view the generated private key?', viewAction);
+    const action = await theia.window.showInformationMessage('Do you want to view the generated public key?', viewAction);
```
l0rd: We should inform the user that the generation of the key pair was successful.

vparfonov:
> I understand that a user will generate/create a key pair and update the ssh config in theia container. This is an operation that the user will do only once.
> 
> Then, when he delete this workspace and creates a new one, the SSH key pair will be mounted as a Kubernetes secret. But what about the SSH config? Will it be re-generated and mounted as well?

yes, it's done here https://github.com/eclipse/che/pull/13809/files#diff-5e2165ab176b2a1fb9829aa579fa546fR139

vinokurig:
@sunix 
> @vinokurig Could you describe the end user flow that would use that ?
How can I test it ? how can I set ssh keys ?

When user executes command: `SSH: generate key pair...` the plugin generates an SSH key pair and stores it in the che database. The improvement generates a file with a private ssh key and a config file that refers the private key to host. It allows to execute git operation via SSH. 
To test it you need to generate a new SSH key, store the public key in the git provider (github.com), then you can clone a repository by ssh url and push to remote.

plugins/ssh-plugin/src/ssh-plugin-backend.ts
vinokurig: > Are we prompting the user for a key pair name or for an host name?

Changed to `Host name`
> And do we really need to prompt the user for this? Can't we use a default name?

We need to distinguish keys by different git providers. We may have a key for github and bitbucket as well.

plugins/ssh-plugin/src/ssh-plugin-backend.ts
vinokurig: done

plugins/ssh-plugin/src/ssh-plugin-backend.ts
l0rd: I don't understand. The same key can be used with multiple git providers. 

When you generate a key pair, the public key is for sharing with everyone. There is no issue with sharing the same public key with multiple organisations.

vinokurig: Agree, but the SSH plugin is initially designed to use multiple keys. basically the model is taken from che 6 https://www.eclipse.org/che/docs/che-6/version-control.html#using-existing-ssh-keys. To use single ssh key we need to completely rework the plugin and the way of restoring the ssh key on workspace start

sunix:
ok so the ssh generation work is done on wsmaster ? and we have a dedicated service ou wsmaster API. I guess this is comming from che6 but it sounds weird in a Che7 architecture.

plugins/ssh-plugin/src/ssh-plugin-backend.ts
l0rd: I don't want to block this PR but I still can't understand why, as a user, I should provide an hostname (i.e. github.com). I don't do that on my host when I setup my SSH key. Why I need that on a Che workspace?

vinokurig:
@sunix 
> ok so the ssh key generation work is done on wsmaster ? and we have a dedicated service on wsmaster API. I guess this is comming from che6 but it sounds weird in a Che7 architecture.
is it just for generating keys ? if yes I am not sure it has to stay on wsmaster ... it could just be a microservice that runs on a container or invoking the right command locally... just thinking out loud ...

Agree, but that's a separate issue.

sunix:
> Agree, but that's a separate issue.

Is it an existing GH issue ? could you create one if not ?

vinokurig:
> Is it an existing GH issue ? could you create one if not ?

https://github.com/eclipse/che/issues/13522

vinokurig:
@l0rd Reworked the plugin according to https://github.com/eclipse/che/issues/13494#issuecomment-512761661, could you please review

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -45,76 +54,93 @@ export async function start() {
...
-        ({
-            label: service.displayName,
-            description: service.description,
-            detail: service.name,
-            name: service.name
-        })), option);
-    if (sshServiceValue) {
-        return Promise.resolve(sshServiceValue.label);
+const updateConfig = function (hostName: string): void {
+    const sshDir = os.homedir() + '/.ssh';
```
benoitf: why not use path.resolve https://nodejs.org/docs/latest/api/path.html#path_path_resolve_paths ? instead of these /

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -45,76 +54,93 @@ export async function start() {
...
+    const configFile = sshDir + '/config';
+    if (!fs.existsSync(configFile)) {
```
benoitf: as we're inside an async function we can use fs-extra and async functions instead of existsSync, mkdirSync, etc

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -45,76 +54,93 @@ export async function start() {
...
+        const newConfigContent = configContent.replace(keyConfig, '');
+        fs.writeFileSync(configFile, newConfigContent);
     } else {
-        return Promise.resolve('');
+        fs.appendFileSync(configFile, keyConfig);
     }
 };
 
+const writeKey = function (name: string, key: string): void {
+    const keyFile = os.homedir() + '/.ssh/' + name.replace('.', '_');
```
benoitf: path.resolve

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -45,76 +54,93 @@ export async function start() {
...
+            fs.mkdirSync(sshDir);
+        }
+        fs.appendFileSync(configFile, '');
+        fs.chmodSync(configFile, '644');
+    }
+    const keyConfig = `\nHost ${hostName.startsWith('default-') ? '*' : hostName}\nIdentityFile ${sshDir}/${hostName.replace('.', '_')}\n`;
+    const configContent = fs.readFileSync(configFile).toString();
+    if (configContent.indexOf(keyConfig) >= 0) {
+        const newConfigContent = configContent.replace(keyConfig, '');
+        fs.writeFileSync(configFile, newConfigContent);
```
benoitf: https://www.npmjs.com/package/fs-extra

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -45,76 +54,93 @@ export async function start() {
...
     const publicKey = await theia.window.showInputBox({ placeHolder: 'Enter public key' });
+    const privateKey = await theia.window.showInputBox({ placeHolder: 'Enter private key' });
 
     await sshkeyManager
-        .create({ name: keyName ? keyName : '', service: await getSshServiceName(), publicKey: publicKey })
-        .then(() => {
-            theia.window.showInformationMessage('Key "' + `${keyName}` + '" successfully created');
+        .create({ name: hostName, service: 'vcs', publicKey: publicKey, privateKey })
+        .then(async () => {
+            theia.window.showInformationMessage('Key "' + `${hostName}` + '" successfully created');
```
benoitf: why do we use string literals and concatenation there ?

-->
`Key "${hostName}" successfully created`

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -45,76 +54,93 @@ export async function start() {
...
-    const keys: cheApi.ssh.SshPair[] = await sshkeyManager.getAll(sshServiceName);
+    const keys: cheApi.ssh.SshPair[] = await sshkeyManager.getAll('vcs');
     const keyResp = await theia.window.showQuickPick<theia.QuickPickItem>(keys.map(key =>
         ({ label: key.name ? key.name : '' })), {});
     const keyName = keyResp ? keyResp.label : '';
     await sshkeyManager
-        .delete(sshServiceName, keyName)
+        .delete('vcs', keyName)
         .then(() => {
+            const keyFile = os.homedir() + '/.ssh/' + keyName.replace('.', '_');
```
benoitf: this ssh folder should be acimputed once, it's built all over this class many times

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -45,76 +54,93 @@ export async function start() {
...
+            if (fs.existsSync(keyFile)) {
```
benoitf: avoid any sync, we're inside async functions

plugins/ssh-plugin/src/ssh-plugin-backend.ts
vinokurig: fixed

plugins/ssh-plugin/src/ssh-plugin-backend.ts
vinokurig: done

plugins/ssh-plugin/src/ssh-plugin-backend.ts
vinokurig: done

plugins/ssh-plugin/src/ssh-plugin-backend.ts
vinokurig: done

plugins/ssh-plugin/src/ssh-plugin-backend.ts
vinokurig: fixed

plugins/ssh-plugin/src/ssh-plugin-backend.ts
vinokurig: done

plugins/ssh-plugin/src/ssh-plugin-backend.ts
vinokurig: done

vinokurig:
@benoitf I've fixed all your comments could you please review

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -45,93 +55,109 @@ export async function start() {
...
-            name: service.name
-        })), option);
-    if (sshServiceValue) {
-        return Promise.resolve(sshServiceValue.label);
+const getKeyFilePath = function (name: string): string {
+    return resolve(os.homedir(), '.ssh', name.replace('.', '_'));
+};
+
+const updateConfig = async function (hostName: string): Promise<void> {
+    const sshDir = resolve(os.homedir(), '.ssh');
```
benoitf: when I was referencing a function for ssh folder, it was for this part
`resolve(os.homedir(), '.ssh')`

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -45,93 +55,109 @@ export async function start() {
...
-        ({
-            label: service.displayName,
-            description: service.description,
-            detail: service.name,
-            name: service.name
-        })), option);
-    if (sshServiceValue) {
-        return Promise.resolve(sshServiceValue.label);
+const getKeyFilePath = function (name: string): string {
+    return resolve(os.homedir(), '.ssh', name.replace('.', '_'));
```
benoitf: probably function here should be for resolve(os.homedir(), '.ssh') 

also function() can be replaced by `getKeyFilePath = (name:string): string => {`

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -45,93 +55,109 @@ export async function start() {
...
-            displayName: 'VCS',
-            description: 'SSH keys used by Che VCS plugins'
-        },
-        {
-            name: 'machine',
-            displayName: 'Workspace Containers',
-            description: 'SSH keys injected into all Workspace Containers'
-        }
-    ];
+const getHostName = async function (): Promise<string> {
```
benoitf: `(): Promise<string> => {` ?

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -45,93 +55,109 @@ export async function start() {
...
+
+const updateConfig = async function (hostName: string): Promise<void> {
+    const sshDir = resolve(os.homedir(), '.ssh');
+    const configFile = resolve(sshDir, 'config');
+    await ensureFile(configFile);
+    await chmod(configFile, '644');
+    const keyConfig = `\nHost ${hostName.startsWith('default-') ? '*' : hostName}\nIdentityFile ${getKeyFilePath(hostName)}\n`;
+    const configContentBuffer = await readFile(configFile);
+    if (configContentBuffer.indexOf(keyConfig) >= 0) {
+        const newConfigContent = configContentBuffer.toString().replace(keyConfig, '');
```
benoitf: replace function is only replacing the first occurence, it's ok ?

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -45,93 +55,109 @@ export async function start() {
...
+    const keys: cheApi.ssh.SshPair[] = await sshkeyManager.getAll('vcs');
     const keyResp = await theia.window.showQuickPick<theia.QuickPickItem>(keys.map(key =>
         ({ label: key.name ? key.name : '' })), {});
     const keyName = keyResp ? keyResp.label : '';
     await sshkeyManager
-        .get(sshServiceName, keyName)
-        .then(key => {
-            theia.workspace.openTextDocument({ content: key.publicKey });
+        .get('vcs', keyName)
+        .then(async key => {
```
benoitf: why do we use a promise there with then and not with await ?
```typescript
const key = await sshkeyManager.get(sshServiceName, keyName)
const document = await theia.workspace.openTextDocument({ content: key.publicKey });
theia.window.showTextDocument(document);
```

plugins/ssh-plugin/src/ssh-plugin-backend.ts
vinokurig: > probably function here should be for resolve(os.homedir(), '.ssh')

We need to resolve key file name as well

plugins/ssh-plugin/src/ssh-plugin-backend.ts
vinokurig: Yes, it is not expected that the config file has 2 identical key configs

PR #360, opened 2019-07-19
Store SSH private key on the file system from SSH plugin

vinokurig:

### What does this PR do?
Store SSH private key file on the file system when key is generated from the SSH Plugin

PR to master: https://github.com/eclipse/che-theia/pull/352
fixes eclipse/che#13866

### DIFF
--- a/plugins/ssh-plugin/package.json
+++ b/plugins/ssh-plugin/package.json
@@ -22,6 +22,9 @@
       "files": [
         "src"
       ],
+      "dependencies": {
+        "fs-extra": "7.0.1"
+      },
       "devDependencies": {
         "@eclipse-che/api": "latest",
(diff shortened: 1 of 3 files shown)


plugins/ssh-plugin/package.json
```diff
@@ -22,6 +22,9 @@
       "files": [
         "src"
       ],
+      "dependencies": {
+        "fs-extra": "7.0.1"
```
evidolob: Do we have CQ for `fs-extra`?

vinokurig: There is a CQ: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20011, BTW it is already used in che-theia https://github.com/eclipse/che-theia/search?q=fs-extra&unscoped_q=fs-extra

PR #361, opened 2019-07-22
Add missing publisher field and set it to 'Eclipse Che'

evidolob:
### What does this PR do?
Add missing publisher field and set it to 'Eclipse Che'
According to this commit
https://github.com/theia-ide/theia/commit/6021d795d61eafb214b234288f1568cae0cfa9f5
`publisher` field is mandatory.
should fix https://github.com/eclipse/che/issues/13944


RomanNikitenko:
I defined theia and che-theia projects in my devfile and tested the changes - it works well - the both projects was cloned.

PR #369, opened 2019-07-25
move che-theia generator to che-theia repository

evidolob:
### What does this PR do?
move che-theia generator to che-theia repository, it also include https://github.com/ws-skeleton/che-theia-generator/pull/26 PR

eclipse/che#13738


sunix:
Don't know if that should be done in this PR but I am not a big fan of the name 'generator' and the README.md doesn't tell what it is for.

PR #378, opened 2019-07-26
Add a note about building Che Theia into generator readme

mmorhun:
We should have a note in Che Theia generator readme about building particularity for Che Theia: if a user builds something from non root directory it will mess up dependencies and end up in non operatable Che Theia or compiling errors.

### DIFF
--- a/generator/README.md
+++ b/generator/README.md
@@ -64,9 +64,17 @@ Dev mode is the way to use all new extensions from `master` branch:
 And `che:theia` will use `master` branch for all extensions and plugins, regardless of provided configuration
 
 ### Development life-cycle
-it's easy to check changes. Running yarn watch in a module and then running yarn watch in `examples/assembly` folder and using `yarn run start` in `examples/assembly`
+Che Theia should be built from root directory only (Root directory of Che Theia is the directory into which Theia was clonned and `che:theia init` was executed there). In case of building from subdirectories it will mess up dependencies, don't do it.
 
-### compiling the plugins
+To build whole Che Theia just execute `yarn` command in the root directory.
+If only one module should be built, use `npx run build <module-name>`. For example `npx run build @theia/plugin-ext`.
+
+Also one may set compilation on changes for some modules. To do so, run `npx run watch <module-name>` from the root directory. Then execute `yarn watch` from `examples/assembly` folder and run Che Theia with `yarn run start` command from the same directory. Make sure, you start watcher for all modules under development.
+
+Note, this is not the case for embedded plugins.
+To develop them one should place copy of their sources somewhere else (outside of the Che Theia folder) and then include new binaries into the assembly.
+
+### Compiling the plugins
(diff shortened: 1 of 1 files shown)


generator/README.md
```diff
@@ -64,9 +64,17 @@ Dev mode is the way to use all new extensions from `master` branch:
...
 
 ### Development life-cycle
-it's easy to check changes. Running yarn watch in a module and then running yarn watch in `examples/assembly` folder and using `yarn run start` in `examples/assembly`
+Che Theia should be built from root directory only. In case of building from subdirectories it will mess up dependencies, don't do it.
 
-### compiling the plugins
+To build whole Che Theia just execute `yarn` command in the root directory.
+If only one module should be built, use `npx run build <module-name>`. For example `npx run build @theia/plugin-ext`.
+
+Also it's easy to check changes. Run `yarn watch` in a module and then execute `yarn watch` from `examples/assembly` folder and run `yarn run start` from the same directory. Make sure, you start watcher in all modules under development.
```
akurinnoy: I suppose there should be "Run `npx run watch <module-name>` in the root directory" instead of "Run `yarn watch` in a module".

mmorhun: Yes, good note, thank you. I just left this one from previous doc writer...

generator/README.md
```diff
@@ -64,9 +64,17 @@ Dev mode is the way to use all new extensions from `master` branch:
 And `che:theia` will use `master` branch for all extensions and plugins, regardless of provided configuration
 
 ### Development life-cycle
-it's easy to check changes. Running yarn watch in a module and then running yarn watch in `examples/assembly` folder and using `yarn run start` in `examples/assembly`
+Che Theia should be built from root directory only. In case of building from subdirectories it will mess up dependencies, don't do it.
 
-### compiling the plugins
+To build whole Che Theia just execute `yarn` command in the root directory.
```
sunix: Could you replace `root directory` with `theia clone directory` ?

mmorhun: Got the point, thanks. Fixed.

generator/README.md
```diff
@@ -64,9 +64,17 @@ Dev mode is the way to use all new extensions from `master` branch:
 And `che:theia` will use `master` branch for all extensions and plugins, regardless of provided configuration
 
 ### Development life-cycle
-it's easy to check changes. Running yarn watch in a module and then running yarn watch in `examples/assembly` folder and using `yarn run start` in `examples/assembly`
+Che Theia should be built from root directory only (Root directory of Che Theia is the directory into which Theia was clonned and `che:theia init` was executed there). In case of building from subdirectories it will mess up dependencies, don't do it.
```
benoitf:
```suggestion
Che Theia should be built from root directory only (Root directory of Che Theia is the directory into which Theia was cloned and `che:theia init` was executed there). In case of building from subdirectories it will mess up dependencies, don't do it.
```

benoitf: if yarn workspaces are used as discussed at F2F and symlinks are gone, it shouldn't mess it up, correct ?

mmorhun: Yes, correct. But for now it is unfortunately happening...

PR #382, opened 2019-07-29
Fix che-theia/generator tests inside container without git user.name and user.email

AndrienkoAleksandr:
### What does this PR do?
Fix che-theia/generator tests inside container without git user.name and user.email

https://github.com/eclipse/che/issues/14056

### DIFF
--- a/generator/tests/init-sources/init-sources.spec.ts
+++ b/generator/tests/init-sources/init-sources.spec.ts
@@ -65,8 +65,8 @@ describe("Test Extensions", () => {
 
     function initGit(path: string) {
         cp.execSync('git init', { cwd: path });
-        cp.execSync('git config --local user.name "test test"', { cwd: path });
-        cp.execSync('git config --local user.email example.com', { cwd: path });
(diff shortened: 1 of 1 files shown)


musienko-maxim:

generator/tests/init-sources/init-sources.spec.ts
```diff
@@ -65,6 +65,8 @@ describe("Test Extensions", () => {
 
     function initGit(path: string) {
         cp.execSync('git init', { cwd: path });
+        cp.execSync('git config --local user.name "test test"', { cwd: path });
+        cp.execSync('git config --local user.email test@gmail.com', { cwd: path });
```
benoitf: maybe better to use `example.com` or `eclipse.org` than `gmail` there

AndrienkoAleksandr: Done

benoitf: @AndrienkoAleksandr I think email should be still with pattern something@something (like `user@example.com`)

PR #385, opened 2019-07-30
push to CDN on each build

evidolob:
### What does this PR do?
Remove check on unused variable, to enable pushing to CDN

eclipse/che#14039


dockerfiles/theia/push-cdn-files-to-akamai.sh
```diff
@@ -10,6 +10,7 @@
 base_dir=$(cd "$(dirname "$0")"; pwd)
 current_dir=$(pwd)
 if [ -z "${AKAMAI_CHE_AUTH:-}" ]; then
+  echo "CDN files will not be pushed to the Akamai directory since the 'AKAMAI_CHE_AUTH' environent variable is not set"
```
benoitf:
```suggestion
  echo "CDN files will not be pushed to the Akamai directory since the 'AKAMAI_CHE_AUTH' environment variable is not set"
```

PR #389, opened 2019-08-01
Show warning message after generating an SSH key

vinokurig:

### What does this PR do?
Show's a warning message that SSH key will be available in all containers (not only in `theia-ide` container) only after restarting the workspace. Workspace restart will mount the new SSH keys to all containers.

https://github.com/eclipse/che/issues/14073

### DIFF
--- a/plugins/ssh-plugin/src/ssh-plugin-backend.ts
+++ b/plugins/ssh-plugin/src/ssh-plugin-backend.ts
@@ -82,6 +82,10 @@ const writeKey = async (name: string, key: string) => {
     await chmod(keyFile, '600');
 };
 
+const showWarning = async () => {
+    theia.window.showWarningMessage('Che Git plugin can leverage the generated keys now. To make them available in every workspace containers please restart your workspace.');
+};
+
 const generateKeyPair = async (sshkeyManager: SshKeyManager) => {
     const keyName = `default-${Date.now()}`;
     const key = await sshkeyManager.generate('vcs', keyName);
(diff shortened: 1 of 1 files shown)


sunix:
Hello @vinokurig could you describe the flow how user would use this ? thanks

vinokurig:
@sunix 
> Hello @vinokurig could you describe the flow how user would use this ? thanks

The problem is that when SSH plugin generates a new SSH key it is written to the `/home/theia/.ssh`  directory of the `theia-ide` container, so the key is invisible from other containers. The aim of the message is to tell the user that after restarting the workspace the key will be mounted to all containers of the workspace.

sunix:
ok got it

vinokurig:
@sunix 
> Could we provide a button for the user to restart the workspace ?

The che workspace API commands that are needed to restart the workspace are not implemented yet.
https://github.com/eclipse/che-theia/blob/e331e7ff8f9e44a630ed3a7441721b6704d05df9/extensions/eclipse-che-theia-plugin-ext/src/plugin/che-workspace.ts#L28-L41

> To me it sounds not obvious for a normal user how to restart a workspace. What if the user restarts the browser ?

As for me the warning message should be as simpler and shorter as possible. I think it is not the place for explanations how to restart the workspace.

ibuziuk:
have added `7.0.0` milestone it is utterly important for Hosted Che to get this fix merged by the GA time

sunix:
@vinokurig if you can:
Propose the user to restart, restart would call stop and refresh the current page.
If you cannot. No worries. This is nice to have.

PR #398, opened 2019-08-05
Show warning message after generating an SSH key (#389)

vinokurig:
SSH plugin generates a new SSH key then it is written to the `/home/theia/.ssh` directory of the theia-ide container, so the key is invisible from other containers. The aim of the message is to tell the user that after restarting the workspace the key will be mounted to all containers of the workspace.

### What does this PR do?
Show's a warning message that SSH key will be available in all containers (not only in `theia-ide` container) only after restarting the workspace. Workspace restart will mount the new SSH keys to all containers.

https://github.com/eclipse/che/issues/14073

The PR to the master is merged: https://github.com/eclipse/che-theia/pull/389

### DIFF
--- a/plugins/ssh-plugin/src/ssh-plugin-backend.ts
+++ b/plugins/ssh-plugin/src/ssh-plugin-backend.ts
@@ -82,6 +82,10 @@ const writeKey = async (name: string, key: string) => {
     await chmod(keyFile, '600');
 };
 
+const showWarning = async () => {
+    theia.window.showWarningMessage('Che Git plugin can leverage the generated keys now. To make them available in every workspace containers please restart your workspace.');
+};
+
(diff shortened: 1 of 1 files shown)


maxura:

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -82,6 +82,10 @@ const writeKey = async (name: string, key: string) => {
     await chmod(keyFile, '600');
 };
 
+const showWarning = async () => {
+    theia.window.showWarningMessage('Che Git plugin can leverage the generated keys now. To make them available in every workspace containers please restart your workspace.');
```
amisevsk:
```suggestion
    theia.window.showWarningMessage('The Che Git plugin can now use the generated keys. To make them available in every workspace container please restart your workspace.');
```

PR #388, opened 2019-08-01
Cherry-pick generator and docker image commits

evidolob:
### What does this PR do?
It cherry-pick next commits:
1. https://github.com/eclipse/che-theia/commit/50767320ae233bad5109d308e9b87d52b97e8ea1 - move generator to che-theia repo
2. https://github.com/eclipse/che-theia/commit/e2d1f85c7f08b00fca06773ba26add056b6e1bb2 Use current che-theia sources for docker image build
3. https://github.com/eclipse/che-theia/commit/af1ec68caee8c59e269613e17ec11fbf01e836f4 Fix che-theia/generator tests inside container

### What issues does this PR fix or reference?
Part of eclipse/che#13738

PR #405, opened 2019-08-07
Support multi root workspace in generator

JPinkney:
### What does this PR do?
This PR sets the multi-root workspace setting to be true by default.

This is being merged into master rather than 7.0.0 branch because when it is merged and CI releases the newest version of the generator we can update the version of the generator [here](https://github.com/eclipse/che-theia/blob/7.0.0/dockerfiles/theia-dev/Dockerfile#L42). (7.0.0 branch doesn't have the generator. See discussion: https://github.com/eclipse/che-theia/pull/388).

Fixes part of https://github.com/eclipse/che/issues/13427.

#### Release Notes

Set workspace.supportMultiRootWorkspace to be true by default

### DIFF
--- a/generator/src/templates/assembly-package.mst
+++ b/generator/src/templates/assembly-package.mst
@@ -6,7 +6,8 @@
     "frontend": {
       "config": {
         "preferences": {
-          "editor.autoSave": "on"
+          "editor.autoSave": "on",
+          "workspace.supportMultiRootWorkspace": true
         }
       }
     }


benoitf:
Hello, I may have missed something but if you need to use in 7.0.0 then you need to use the previous generator there https://github.com/ws-skeleton/che-theia-generator (which can publish to npmjs and then use that in 7.0.0)

JPinkney:
@benoitf Must have been a misunderstanding on my side, I've opened a PR on the other repo instead: https://github.com/ws-skeleton/che-theia-generator/pull/32

PR #407, opened 2019-08-15
[SSH] Add no key pair warnings

gorkem:
Adds warnings to view and delete keys when no key pairs exist.
fixes https://github.com/eclipse/che/issues/14215

### DIFF
--- a/plugins/ssh-plugin/src/ssh-plugin-backend.ts
+++ b/plugins/ssh-plugin/src/ssh-plugin-backend.ts
@@ -82,8 +82,12 @@ const writeKey = async (name: string, key: string) => {
     await chmod(keyFile, '600');
 };
 
-const showWarning = async () => {
-    theia.window.showWarningMessage('Che Git plugin can leverage the generated keys now. To make them available in every workspace containers please restart your workspace.');
+const showReadyToUse = async () => {
+    theia.window.showWarningMessage('Git plugin can leverage the generated keys now. To make them available in every workspace containers please restart your workspace.');
+};
+
+const showNoKeys = async () => {
+    theia.window.showWarningMessage('No SSH key pairs has been defined.');
 };
 
 const generateKeyPair = async (sshkeyManager: SshKeyManager) => {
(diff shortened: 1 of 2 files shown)


plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -82,8 +82,16 @@ const writeKey = async (name: string, key: string) => {
     await chmod(keyFile, '600');
 };
 
-const showWarning = async () => {
-    theia.window.showWarningMessage('Che Git plugin can leverage the generated keys now. To make them available in every workspace containers please restart your workspace.');
+const warning = async (message: string) => {
```
vinokurig: It is a function I think `showWarning` suites more

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -82,8 +82,16 @@ const writeKey = async (name: string, key: string) => {
...
+    theia.window.showWarningMessage(message);
+};
+
+const showReadyToUse = async () => {
+    warning('Che Git plugin can leverage the generated keys now. To make them available in every workspace containers please restart your workspace.');
```
vinokurig: Why do we need to wrap one function to another? I think it would be simpler to define message constants and use only one function.

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -124,14 +132,18 @@ const createKeyPair = async (sshkeyManager: SshKeyManager) => {
...
-        showWarning();
+        showReadyToUse();
     } catch (error) {
         theia.window.showErrorMessage(error);
     }
 };
 
 const deleteKeyPair = async (sshkeyManager: SshKeyManager) => {
     const keys: cheApi.ssh.SshPair[] = await sshkeyManager.getAll('vcs');
+    if (keys.length < 1) {
```
vinokurig: The array check can be moved inside a function, it is used more than in one place

plugins/ssh-plugin/src/ssh-plugin-backend.ts
gorkem: yeah removed the middle man

dmytro-ndp:
crw-ci-test

benoitf:
closing as outdated

PR #408, opened 2019-08-19
WIP Make Che use multi root workspaces by default

JPinkney:
### What does this PR do?
This PR makes Che use multi root workspaces by default.

It is still WIP because there are underlying issues with the vscode-languageclient-node library that causes a race condition between when a project is cloned and when the didUpdateWorkspaceFolders is enabled. See: https://github.com/microsoft/vscode-languageserver-node/issues/451

Related issues for C# and PHP support:
https://github.com/eclipse/che/issues/14223
https://github.com/eclipse/che/issues/14225

This PR will be considered mergable when the vscode-languageserver-node fix is in and all the stacks are behaving as expected.

Currently, this is status of the stacks:

| Stack | First project works | Second project works | Issues/Comments |
|---|---|---|---|
Java-maven | 🚫 | 🚫 | still same classpath issue
Golang | 👍 | 👍 | everything is working as expected
PHP | 👍 | 👍 | everything is working as expected
PHP Symphony | 👍 | 👍 |  everything is working as expected
Python (without clicking "Try it now" in the popup) | 👍 | 👍 | everything is working as expected
Python (after clicking "Try it now" in the popup) | ⚠️ | 🚫 | Document symbols, hover, format, autocompletion worked on the first project. Validation did not. On second project import I have to re-click "Try it now" and then it says a process is now listening on port 2503
NodeJS React | 👍 | 👍  | everything is working as expected
.NET Core | 🚫 | 🚫 | No features are working with first project or second project.

Testing involved using the sample and adding in:
```
  - type: cheEditor
    reference: >-
      https://gist.githubusercontent.com/JPinkney/fd653195c71604e3b3002e46393b0f18/raw/1240df21746c56ac4ff66ecd8ee3a9767b074e68/reference.yaml
```
as the cheEditor in the devfile. That referenced yaml is the result of building a che-theia image with these changes.

Then I waited for clone to finish, tested out features. Then added in a second project (either by cloning or by creating a new folder in /projects) and tested the features on that. For Golang, Python and NodeJS React I just made new sample project myself inside of the /projects folder and added them to the workspace. For PHP I cloned https://github.com/banago/simple-php-website.

Part of the fixes for https://github.com/eclipse/che/issues/13427


dmytro-ndp:
crw-ci-test

plugins/factory-plugin/src/theia-commands.ts
```diff
@@ -68,13 +69,13 @@ export class TheiaCloneCommand {
         this.projectsRoot = projectsRoot;
     }
 
-    execute(): PromiseLike<void> {
```
sunix: i don't see any good reasons to rename `execute` to `clone`.
This is a command. At the moment we tell Che-theia to clone (running clone commands) at start up but we may chain other commands that are not about cloning.

plugins/factory-plugin/src/factory-initializer.ts
```diff
@@ -92,9 +92,16 @@ export class FactoryInitializer {
...
         }
 
+        const workspaceFolders: theia.Uri[] = [];
         await Promise.all(
-            cloneCommands.map(command => command.execute())
+            cloneCommands.map(cloneCommand => {
+                if (cloneCommand.isInTheiaWorkspace()) {
+                    workspaceFolders.push(theia.Uri.file(cloneCommand.folder));
+                }
+                cloneCommand.clone();
```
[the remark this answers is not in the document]
sunix: I am not in favour of having logic implementation here. It should go to the command implementation.

plugins/factory-plugin/src/workspace-projects-manager.ts
```diff
@@ -61,9 +54,16 @@ abstract class WorkspaceProjectsManager {
...
         await Promise.all(
-            cloneCommandList.map(cloneCommand => cloneCommand.execute())
+            cloneCommandList.map(cloneCommand => {
+                if (!cloneCommand.isInTheiaWorkspace()) {
+                    workspaceFolders.push(theia.Uri.file(cloneCommand.folder));
+                }
+                return cloneCommand.clone();
+            })
         );
+        await theia.commands.executeCommand('che.workspace.addFolder', workspaceFolders);
```
sunix: same comments i had for factory-initializer

> I am not in favour of having logic implementation here. It should go to the command implementation.

You see here you are duplicating code.
Also, workspace root folders may be handle differently in the future see: https://github.com/eclipse/che/issues/15347

benoitf:
currently superseed by https://github.com/eclipse/che-theia/pull/778
archiving

PR #367, opened 2019-07-23
Use 'containerName' instead of 'machineName' for che tasks

RomanNikitenko:
### What does this PR do?
Use 'containerName' instead of 'machineName' for che tasks

### What issues does this PR fix or reference?
https://github.com/eclipse/che/issues/13667

### How to test
Use the component in your devfile

```
- 
    alias: theia-editor
    reference: >-
      https://raw.githubusercontent.com/RomanNikitenko/che-plugin-registry/master/v3/plugins/eclipse/che-theia/custom/meta.yaml
    type: cheEditor

```
The image contains:
- che-7.0.0 theia branch
- current 7.0.0 branch of che-theia + changes from the PR
 
Use some command in your devfile, for example:
```
- name: che-theia:build
  actions:
  - type: exec
    component: che-dev
    command: >
              yarn
    workdir: /projects/che-theia
```
The `component` field defined for a command in a devfile should be converted into `containerName` (not `machineName`) field for a task configuration in a `tasks.json` file when a workspace has started.

containerName

Signed-off-by: Roman Nikitenko <rnikiten@redhat.com>

PR #414, opened 2019-08-22
Use 'containerName' instead of 'machineName' for che tasks

RomanNikitenko:
The changes are the same as in the PR https://github.com/eclipse/che-theia/pull/367

### What does this PR do?
Use 'containerName' instead of 'machineName' for che tasks

https://github.com/eclipse/che/issues/13667

### How to test
Use the component in your devfile

```
- 
    alias: theia-editor
    reference: >-
      https://raw.githubusercontent.com/RomanNikitenko/che-plugin-registry/master/v3/plugins/eclipse/che-theia/next/meta.yaml
    type: cheEditor

```
 
Use some command in your devfile, for example:
```
- name: che-theia:build
  actions:
  - type: exec
    component: che-dev
    command: >
              yarn
    workdir: /projects/che-theia
```
The `component` field defined for a command in a devfile should be converted into `containerName` (not `machineName`) field for a task configuration in a `tasks.json` file when a workspace has started.

containerName

### DIFF
--- a/plugins/task-plugin/src/task/che-task-provider.ts
+++ b/plugins/task-plugin/src/task/che-task-provider.ts
@@ -45,10 +45,10 @@ export class CheTaskProvider {
             resultTarget.workspaceId = await this.cheWorkspaceClient.getWorkspaceId();
         }
 
-        if (target && target.machineName) {
-            resultTarget.machineName = target.machineName;
+        if (target && target.containerName) {
+            resultTarget.containerName = target.containerName;
         } else {
-            resultTarget.machineName = await this.machinePicker.pick();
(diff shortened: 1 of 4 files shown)


vparfonov:
> Is it still relevant? I mean that there is another issue to use component instead. Or it will be the next step?

yes, it will be next step. With this PR we align with 7.0 version

RomanNikitenko:
@sleshchenko 
You are right - we are going to use `component` instead of 'machineName', but we are not ready for it and we have the corresponding issue.
So I would like to align the behavior for master branch with 7.0.0 and we can improve it when the corresponding issue is resolved.

RomanNikitenko:
I had to force-push the same changes to restart ci-build.

PR #427, opened 2019-09-04
Use python:3.7.4-slim as a base image for python

tolusha:

### What does this PR do?
It uses Debian based image because Microsoft Python Language Server doesn't start at Alpine based.

https://github.com/eclipse/che/issues/14074

### DIFF
--- a/dockerfiles/remote-plugin-python-3.7.3/Dockerfile
+++ b/dockerfiles/remote-plugin-python-3.7.3/Dockerfile
@@ -7,133 +7,26 @@
 #
 # Contributors:
 #   Red Hat, Inc. - initial API and implementation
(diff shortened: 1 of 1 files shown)


l0rd:
@tolusha in the context of this PR it would be nice if you could update the python stack [tooling image](https://github.com/eclipse/che-devfile-registry/blob/master/arbitrary-users-patch/base_images#L1) and it's [description](https://github.com/eclipse/che-devfile-registry/blob/master/devfiles/python/meta.yaml#L3).

tolusha:
@l0rd 
I've updated the python tooling image
https://github.com/eclipse/che-devfile-registry/pull/93
but it isn't clear regarding its description. It is still python 3.6 stack.

tolusha:
@benoitf @l0rd @amisevsk 
Could some one validate the PR ?

PR #429, opened 2019-09-06
Generate settings.xml if MAVEN_MIRROR_URL is set

tolusha:

### What does this PR do?
Generates  `/home/theia/.m2/settings.xml` if `MAVEN_MIRROR_URL` is set.

```xml
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0
                      https://maven.apache.org/xsd/settings-1.0.0.xsd">
  <mirrors>
    <mirror>
      <url>${env.MAVEN_MIRROR_URL}</url>
      <mirrorOf>external:*</mirrorOf>
    </mirror>
  </mirrors>
</settings>
```

https://github.com/eclipse/che/issues/14391

### DIFF
--- a/dockerfiles/remote-plugin-java11/etc/before-start.sh
+++ b/dockerfiles/remote-plugin-java11/etc/before-start.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 #
-# Copyright (c) 2018-2019 Red Hat, Inc.
+# Copyright (c) 2019 Red Hat, Inc.
 # This program and the accompanying materials are made
 # available under the terms of the Eclipse Public License 2.0
 # which is available at https://www.eclipse.org/legal/epl-2.0/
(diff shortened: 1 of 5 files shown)


tolusha:
@benoitf 
Could you validate the PR ?

benoitf:
@tolusha why do we have M2 settings in theia endpoint ? it's not related to any programming language

benoitf:
Probably you're requiring like `/entrypoint.sh` including optional like `/before-start.sh` and include that file in the java endpoint containers ?

tolusha:
@benoitf 
I've updated the PR

dockerfiles/remote-plugin-java11/etc/before-start.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/sh
+#
+# Copyright (c) 2018-2019 Red Hat, Inc.
```
benoitf: minor: just 2019

PR #430, opened 2019-09-06
Implement remote plugin resources retrieving

mmorhun:

### What does this PR do?
Adds PluginReaderExtension class which handles case when requested resource is not found locally and in such cases sends request to corresponding sidecar where plugin resources are actually stored.

This PR does not handle resources for remote plugin webview because they are handles in a different way than usual plugin resources. Created issue for that problem: https://github.com/eclipse/che/issues/14444

#### Changelog
Added support for remote plugins resources on Che Theia frontend side.

#### Release Notes
Added support for remote plugins resources on Che Theia frontend side.

#### Screenshots
Two remote plugins are loaded: Kubernetes plugins and Git UI Tools plugin (last two items in left sidebar)

remote-plugins
gut-ui-tools-remote-resources

Fixes https://github.com/eclipse/che/issues/13750
Depends on: https://github.com/theia-ide/theia/pull/6126

### DIFF
--- a/extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts
+++ b/extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts
@@ -16,6 +16,8 @@ import { Websocket } from './websocket';
 import { getPluginId } from '@theia/plugin-ext/lib/common';
 import { PluginDiscovery } from './plugin-discovery';
 
+type PromiseResolver = (value?: Buffer) => void;
+
 /**
  * Class handling remote connection for executing plug-ins.
  * @author Florent Benoit
@@ -32,14 +34,19 @@ export class HostedPluginRemote {
     protected hostedPluginMapping: HostedPluginMapping;
 
     /**
-     * mapping between endpoint name and the websockets
+     * Mapping between endpoint name and the websockets
      */
     private endpointsSockets = new Map<string, Websocket>();
 
     /**
-     * mapping between endpoint's name and the websocket endpoint
+     * Mapping between endpoint's name and the websocket endpoint
      */
     private pluginsMetadata: Map<string, PluginMetadata[]> = new Map<string, PluginMetadata[]>();
+
+    /**
+     * Mapping between resource request id (pluginId_resourcePath) and resource query callback.
+     */
+    private resourceRequests: Map<string, PromiseResolver> = new Map<string, PromiseResolver>();
 
     @postConstruct()
     protected postConstruct(): void {
@@ -155,6 +162,14 @@ export class HostedPluginRemote {
                     this.hostedPluginMapping.getPluginsEndPoints().set(entryName, jsonMessage.endpointName);
                 }
             });
+            return;
+        }
+
+        if (jsonMessage.method === 'getResource') {
+            const resourceBase64 = jsonMessage.data;
+            const resource = resourceBase64 ? Buffer.from(resourceBase64, 'base64') : undefined;
+            this.onGetResourceResponse(jsonMessage['pluginId'], jsonMessage['path'], resource);
+            return;
         }
     }
 
@@ -175,6 +190,52 @@ export class HostedPluginRemote {
      */
     async getExtraPluginMetadata(): Promise<PluginMetadata[]> {
         return [].concat.apply([], [...this.pluginsMetadata.values()]);
+    }
+
+    /**
+     * Sends request to retreive plugin resource from its sidecar.
+     * Returns undefined if plugin doesn't run in sidecar or doesn't exist.
+     * @param pluginId id of the plugin for which resource should be retreived
+     * @param resourcePath relative path of the requested resource based on plugin root directory
+     */
+    public requestPluginResource(pluginId: string, resourcePath: string): Promise<Buffer | undefined> {
+        if (this.hasEndpoint(pluginId) && resourcePath) {
+            return new Promise<Buffer | undefined>((resolve, reject) => {
+                const endpoint = this.hostedPluginMapping.getPluginsEndPoints().get(pluginId);
+                const targetWebsocket = this.endpointsSockets.get(endpoint);
+                if (!targetWebsocket) {
+                    reject(new Error(`No websocket connection for plugin: ${pluginId}`));
+                }
+
+                this.resourceRequests.set(this.getResourceRequestId(pluginId, resourcePath), resolve);
+                targetWebsocket.send(JSON.stringify({
+                    'internal': {
+                        'method': 'getResource',
+                        'pluginId': pluginId,
+                        'path': resourcePath
+                    }
+                }));
+            });
+        }
+        return undefined;
+    }
+
+    /**
+     * Handles all responses from all remote plugins.
+     * Resolves promise from getResource method with requested data.
+     */
+    onGetResourceResponse(pluginId: string, resourcePath: string, resource: Buffer | undefined): void {
+        const key = this.getResourceRequestId(pluginId, resourcePath);
+        const resourceResponsePromiseResolver = this.resourceRequests.get(key);
+        if (resourceResponsePromiseResolver) {
+            // This response is being waited for
+            this.resourceRequests.delete(key);
+            resourceResponsePromiseResolver(resource);
+        }
+    }
+
+    private getResourceRequestId(pluginId: string, resourcePath: string): string {
+        return pluginId + '_' + resourcePath;
     }
 
 }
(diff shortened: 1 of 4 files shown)


extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts
```diff
@@ -41,6 +43,11 @@ export class HostedPluginRemote {
      */
     private pluginsMetadata: Map<string, PluginMetadata[]> = new Map<string, PluginMetadata[]>();
 
+    /**
+     * mapping between resource request id (pluginId_resourcePath) and resource query callback.
```
AndrienkoAleksandr: Maybe `mapping` should starts from capital letter. But up to you.

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-reader-extension.ts
```diff
@@ -0,0 +1,69 @@
...
+
+import * as path from 'path';
+import * as express from 'express';
+import * as escape_html from 'escape-html';
+import { injectable, inject } from 'inversify';
+import { HostedPluginReader } from '@theia/plugin-ext/lib/hosted/node/plugin-reader';
+import { HostedPluginRemote } from './hosted-plugin-remote';
+
+/**
+ * Pathces original plugin reader to be able to retrieve remote plugin resources.
```
AndrienkoAleksandr: Little type `Pathces` -> `Patches`

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
```diff
@@ -214,6 +221,32 @@ to pick-up automatically a free port`));
                     return;
                 }
 
+                // asked to send plugin resource
+                if (jsonParsed.internal.method === 'getResource') {
```
AndrienkoAleksandr: Sorry, for stop method we have one more check:
`jsonParsed.internal.method && jsonParsed.internal.method === 'stop'`
Maybe you need to apply `jsonParsed.internal.method` to your code too? 
 => if (`jsonParsed.internal.methodjsonParsed.internal.method === 'getResource'`) {

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-reader-extension.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+@injectable()
+export class PluginReaderExtension {
+
+    // To be set on connection creation
+    // If there are more than one cnnection, the last one will be used.
```
AndrienkoAleksandr: cnnection => connection

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-reader-extension.ts
mmorhun: thanks

extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts
mmorhun: Other similar comments are written that way, so I decided not to break the style, but will change all of them to capital letter.

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
mmorhun: I don't think so. We are sure that `jsonParsed.internal` is present. Then we need to check `jsonParsed.internal.method` field. Of course, we can firstly check for undefined and then compare to method name, but I see no point because we may just check equality. If `method` field is undefined it is definitely not equal to `getResource`.

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-reader-extension.ts
mmorhun: thanks

mmorhun:
crw-ci-test

extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts
```diff
@@ -177,4 +192,50 @@ export class HostedPluginRemote {
...
     }
 
+    /**
+     * Sends request to retreive plugin resource from its sidecar.
+     * Returns undefined if plugin doesn't run in sidecar or doesn't exist.
+     * @param pluginId id of the plugin for which resource should be retreived
+     * @param resourcePath relative path of the requested resource based on plugin root directory
+     */
+    public requestPluginResource(pluginId: string, resourcePath: string): Promise<Buffer | undefined> {
+        if (this.hasEndpoint(pluginId) && resourcePath) {
```
olexii4: I think it will more readable
```
if (!this.hasEndpoint(pluginId) || !resourcePath) {
    return undefined;
}

return new Promise((resolve, reject) => {
 ...
});

```

olexii4: WDYT?

extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts
```diff
@@ -177,4 +192,50 @@ export class HostedPluginRemote {
...
+                }
+
+                this.resourceRequests.set(this.getResourceRequestId(pluginId, resourcePath), resolve);
+                targetWebsocket.send(JSON.stringify({
+                    'internal': {
+                        'method': 'getResource',
+                        'pluginId': pluginId,
+                        'path': resourcePath
+                    }
+                }));
```
vitaliy-guliy: Do we need to call `resolve()` somewhere in the promise?

extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts
```diff
@@ -177,4 +192,50 @@ export class HostedPluginRemote {
...
+    public requestPluginResource(pluginId: string, resourcePath: string): Promise<Buffer | undefined> {
+        if (this.hasEndpoint(pluginId) && resourcePath) {
+            return new Promise<Buffer | undefined>((resolve, reject) => {
+                const endpoint = this.hostedPluginMapping.getPluginsEndPoints().get(pluginId);
+                const targetWebsocket = this.endpointsSockets.get(endpoint);
+                if (!targetWebsocket) {
+                    reject(new Error(`No websocket connection for plugin: ${pluginId}`));
+                }
+
+                this.resourceRequests.set(this.getResourceRequestId(pluginId, resourcePath), resolve);
```
vitaliy-guliy: I would use contatenation or something like
`${pluginId}_${resourcePath}`

extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts
```diff
@@ -177,4 +192,50 @@ export class HostedPluginRemote {
...
+        }
+        return undefined;
+    }
+
+    /**
+     * Handles all responses from all remote plugins.
+     * Resolves promise from getResource method with requested data.
+     */
+    onGetResourceResponse(pluginId: string, resourcePath: string, resource: Buffer | undefined): void {
+        const key = this.getResourceRequestId(pluginId, resourcePath);
```
vitaliy-guliy: I would use concatenation or something like
`${pluginId}_${resourcePath}`
But it's more question of habits.

extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts
mmorhun: We have to call resolve, but not here. It is called [here](https://github.com/eclipse/che-theia/blob/75990857f320bebf9e511b2f13a72b64dbb75f04/extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts#L233).

extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts
mmorhun: I would like to keep the method. It's encapsulation of how we create that id.

extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts
mmorhun: Both variants have their own pros and cons. Your case is clearer in term of return value but has two negation in the expression. To me both are ok.

mmorhun:
crw-ci-test

PR #434, opened 2019-09-11
Make GitHub token not mandatory if rate limits are ok

mmorhun:

### What does this PR do?
Makes GITHUB_TOKEN not mandatory for building of Che-Theia images.

If host has enough quota left to GitHub API we do not require GitHub token any more.
If quota isn't enough we'll fail the build with message to provide GitHub token.

Also this logic is moved from dockerfiles into build bash scripts.

Fixes: https://github.com/eclipse/che/issues/14194


dockerfiles/theia-endpoint-runtime/Dockerfile
```diff
@@ -10,34 +10,21 @@
...
-        else \
-            echo "GITHUB_TOKEN variable is not set but https://api.github.com rate limit has enough slots"; \
-        fi \
-    fi
-
-#invalidate cashe
-ADD https://${GITHUB_TOKEN}:x-oauth-basic@api.github.com/repos/theia-ide/theia/git/refs/head /tmp/branch_info.json
-ADD https://${GITHUB_TOKEN}:x-oauth-basic@api.github.com/repos/eclipse/che-theia/git/refs/head /tmp/branch_info.json
+# Invalidate cashe
+ADD https://api.github.com/rate_limit /tmp/github_rate_limit.json
```
benoitf: we should invalidate cache only if remote code has changed, not if rate limit is changing

mmorhun: Yes, you are right, will revert cache invalidation lines

mmorhun: @benoitf done. Thanks for the remark.

PR #439, opened 2019-09-16
Replace github.com/theia-ide/theia by github.com/eclipse-theia/theia.

rhopp:
Fixes https://github.com/eclipse/che/issues/14529

### What does this PR do?
As theia upstream moved from github.com/theia-ide/theia to github.com/eclipse-theia/theia, I'm opening this PR to target that.

Also this caused failures like https://github.com/eclipse/che/issues/14529, because the second call in the redirect was not using the github token, thus we were hitting API rate limits. This PR should fix that.

Only required places to fix #14529 are the dockerfiles, but I thought it would be nice to change it everywhere.

https://github.com/eclipse/che/issues/14529


dmytro-ndp:
crw-ci-test

PR #440, opened 2019-09-16
Run Che command in selected container

mmorhun:

### What does this PR do?
In some cases it is impossible to figure out container for a command defined in Che workspace devfile. Then such a command is displayed in each container of the workspace in Workspace panel. But when a user clicks on it, Che asks the user to select container in which this command should be run. But in case of clicking on specific node of specific container it is obvious in which container this command should be run this time.
This PR addresses such case and runs the clicked command in the right container.

Example of workspace which has ambiguous commands:
<details>
<summary>devfile</summary>

```yaml
metadata:
  name: test-tasks
projects:
  - name: NodeJS-Sample-App.git
    source:
      location: 'https://github.com/sleshchenko/NodeJS-Sample-App.git'
      type: git
      branch: master
components:
  - id: eclipse/che-theia/next
    type: cheEditor
    alias: theia-editor
  - referenceContent: |
      ---
      apiVersion: v1
      kind: List
      items:
      -
        apiVersion: apps/v1
        kind: Deployment
        metadata:
          name: mongo
          labels:
            app.kubernetes.io/name: employee-manager
            app.kubernetes.io/component: database
        spec:
          selector:
            matchLabels:
              app.kubernetes.io/name: employee-manager
              app.kubernetes.io/component: database
          template:
            metadata:
              labels:
                app.kubernetes.io/name: employee-manager
                app.kubernetes.io/component: database
              name: mongo
            spec:
              containers:
              - name: mongo
                image: mongo
                ports:
                - name: mongo
                  containerPort: 27017
                volumeMounts:
                    - name: mongo-persistent-storage
                      mountPath: /data/db
              volumes:
                - name: mongo-persistent-storage
                  persistentVolumeClaim:
                    claimName: mongo-persistent-storage
      -
        apiVersion: v1
        kind: Service
        metadata:
          name: mongo
          labels:
            app.kubernetes.io/name: employee-manager
            app.kubernetes.io/component: database
        spec:
          ports:
            - port: 27017
              targetPort: 27017
          selector:
            app.kubernetes.io/name: employee-manager
            app.kubernetes.io/component: database
      -
        apiVersion: v1
        kind: PersistentVolumeClaim
        metadata:
          name: mongo-persistent-storage
          labels:
            app.kubernetes.io/name: employee-manager
            app.kubernetes.io/component: database
        spec:
          accessModes:
          - ReadWriteOnce
          resources:
            requests:
              storage: 1Gi
      -
        apiVersion: apps/v1
        kind: Deployment
        metadata:
          name: nodejs-app
          labels:
            app.kubernetes.io/name: employee-manager
            app.kubernetes.io/component: nodejs-app
        spec:
          selector:
            matchLabels:
              app.kubernetes.io/name: employee-manager
              app.kubernetes.io/component: nodejs-app
          template:
            metadata:
              labels:
                app.kubernetes.io/name: employee-manager
                app.kubernetes.io/component: nodejs-app
              name: nodejs-app
            spec:
              containers:
              - name: web-app
                command: ["tail"]
                args: ["-f", "/dev/null"]
                image: "sleshchenko/nodejs-mongo-sample:latest"
                ports:
                - containerPort: 3000
                  name: http-server
                env:
                  - name: MONGO_HOST
                    value: "mongo"
                  - name: MONGO_PORT
                    value: "27017"
                volumeMounts:
                  - name: projects
                    mountPath: /projects
              volumes:
                - name: projects
                  persistentVolumeClaim:
                    claimName: projects
      -
        apiVersion: v1
        kind: Service
        metadata:
          name: web
          labels:
            app.kubernetes.io/name: employee-manager
            app.kubernetes.io/component: nodejs-app
        spec:
          type: LoadBalancer
          ports:
            - name: web
              port: 80
              targetPort: 3000
              protocol: TCP
          selector:
            app.kubernetes.io/name: employee-manager
            app.kubernetes.io/component: nodejs-app
      -
        apiVersion: extensions/v1beta1
        kind: Ingress
        metadata:
          name: nodejs-ingress
          labels:
            app.kubernetes.io/name: employee-manager
            app.kubernetes.io/component: nodejs-app
          annotations:
            kubernetes.io/ingress.class: "nginx"
        spec:
          rules:
          - host: nodejs.192.168.99.100.nip.io
            http:
              paths:
              - path: /
                backend:
                  serviceName: web
                  servicePort: 80
      -
        apiVersion: v1
        kind: PersistentVolumeClaim
        metadata:
          name: projects
          labels:
            app.kubernetes.io/name: employee-manager
            app.kubernetes.io/component: nodejs-app
        spec:
          accessModes:
          - ReadWriteOnce
          resources:
            requests:
              storage: 1Gi
    type: kubernetes
    alias: employee-manager
    reference: deploy_k8s.yaml
apiVersion: 1.0.0
commands:
  - name: Test ls and echo
    actions:
      - workdir: /home
        type: exec
        command: ls -l && echo "hello world"
        component: employee-manager
  - name: Echo test works
    actions:
      - workdir: /projects
        type: exec
        command: ls -la && echo "hello world"
        component: theia-dev
  - name: Test echo
    actions:
      - type: exec
        command: echo "hello world"
        component: employee-manager
  - name: Test echo
    actions:
      - type: exec
        command: echo "hello world"
        component: employee-manager
  - name: build (test)
    actions:
      - workdir: /projects/theia
        type: exec
        command: yarn
        component: employee-manager

```

</details>

Provides fix for: https://github.com/eclipse/che/issues/13764

### DIFF
--- a/plugins/containers-plugin/src/containers-tree-data-provider.ts
+++ b/plugins/containers-plugin/src/containers-tree-data-provider.ts
@@ -106,7 +106,10 @@ export class ContainersTreeDataProvider implements theia.TreeDataProvider<ITreeN
                         name: command.commandName,
                         tooltip: command.commandLine,
                         iconPath: 'fa-cogs medium-yellow',
-                        command: { id: 'task:run', arguments: [this.getRootPath(), command.commandName] }
+                        command: {
+                            id: CONTAINERS_PLUGIN_RUN_TASK_COMMAND_ID,
+                            arguments: [this.getRootPath(), command.commandName, container.name]
+                        }
                     });
                 });
             }
@@ -266,4 +269,28 @@ export class ContainersTreeDataProvider implements theia.TreeDataProvider<ITreeN
             return this.treeNodeItems.filter(item => item.parentId === undefined);
         }
     }
+}
+
+export const CONTAINERS_PLUGIN_RUN_TASK_COMMAND_ID = 'containers-plugin-run-task';
+/**
+ * Command handler which is invoked when a user clicks on a task in the Workspace panel.
+ * This is needed if a command doesn't have container to be run in specified.
+ * In such case we run the command in the container under which this command was clicked.
+ */
+export async function containersTreeTaskLauncherCommandHandler(source: string, label: string, containerName: string): Promise<void> {
+    const tasks: theia.Task[] = await theia.tasks.fetchTasks({ type: 'che' });
+    for (const task of tasks) {
+        if (task.name === label && task.source === source) {
+            if (!task.definition.target) {
+                task.definition.target = {};
+            }
+            task.definition.target.containerName = containerName;
+
+            theia.tasks.executeTask(task);
+            return;
+        }
+    }
+
+    // Shouldn't happen. Fallback to default behaviour.
+    theia.commands.executeCommand('task:run', source, label);
 }
--- a/plugins/containers-plugin/src/containers-plugin.ts
+++ b/plugins/containers-plugin/src/containers-plugin.ts
@@ -8,9 +8,13 @@
  * SPDX-License-Identifier: EPL-2.0
  **********************************************************************/
 
-import { ContainersTreeDataProvider } from './containers-tree-data-provider';
-import { ContainersService } from './containers-service';
 import * as theia from '@theia/plugin';
+import { ContainersService } from './containers-service';
+import {
+    ContainersTreeDataProvider,
+    CONTAINERS_PLUGIN_RUN_TASK_COMMAND_ID,
+    containersTreeTaskLauncherCommandHandler
+} from './containers-tree-data-provider';
 
 export function start(context: theia.PluginContext) {
     const treeDataProvider = new ContainersTreeDataProvider();
@@ -28,6 +32,11 @@ export function start(context: theia.PluginContext) {
         console.error(error);
         theia.window.showErrorMessage(error);
     });
+
+    const containersTreeTaskLauncherCommand = { id: CONTAINERS_PLUGIN_RUN_TASK_COMMAND_ID };
+    context.subscriptions.push(
+        theia.commands.registerCommand(containersTreeTaskLauncherCommand, containersTreeTaskLauncherCommandHandler)
+    );
 }
 
 export function stop() {
--- a/yarn.lock
+++ b/yarn.lock
@@ -8270,9 +8270,9 @@ move-concurrently@^1.0.1:
     rimraf "^2.5.4"
     run-queue "^1.0.3"
 
-"moxios@git://github.com/stoplightio/moxios.git#v1.3.0":
+"moxios@git://github.com/stoplightio/moxios#v1.3.0":
   version "1.3.0"
-  resolved "git://github.com/stoplightio/moxios.git#9d702c8eafee4b02917d6bc400ae15f1e835cf51"
+  resolved "git://github.com/stoplightio/moxios#9d702c8eafee4b02917d6bc400ae15f1e835cf51"
   dependencies:
     class-autobind "^0.1.4"
 


plugins/containers-plugin/src/containers-tree-data-provider.ts
```diff
@@ -267,3 +270,27 @@ export class ContainersTreeDataProvider implements theia.TreeDataProvider<ITreeN
...
+ */
+export async function containersTreeTaskLauncherCommandHandler(source: string, label: string, containerName: string): Promise<void> {
+    const tasks: theia.Task[] = await theia.tasks.fetchTasks({ type: 'che' });
+    for (const task of tasks) {
+        if (task.name === label && task.source === source) {
+            if (!task.definition.target) {
+                task.definition.target = {};
+            }
+            task.definition.target.containerName = containerName;
+
```
olexii4:
```
task.definition.target = { containerName };
```
Is better instead of 
```
           if (!task.definition.target) {
                task.definition.target = {};
            }
            task.definition.target.containerName = containerName;
```
WDYT?

mmorhun: We cannot do that because we'll lose other properties which are stored under `target` key if any.

mmorhun:
@benoitf because I need newer version of some Theia dependencies and I just regenerated yarn.lock file. Anyway we need to update it soon or late. WDYT?

plugins/containers-plugin/src/containers-tree-data-provider.ts
olexii4: ok agree

olexii4:
@mmorhun In the case when we regenerate yarn.lock file just because we need to update it soon or later - add these changes in a separate commit.

dmytro-ndp:
crw-ci-test

mmorhun:
@RomanNikitenko thanks for your review.
I've investigated the issue which you have and I can say that it doesn't change the current behaviour, just has slightly different one.
Let me explain. I use [executeTask](https://github.com/eclipse-theia/theia/blob/df1e55a882e3dff7b2d0a9984fd6e8ae890f85d5/packages/plugin/src/theia.d.ts#L8433) API method in my implementation, which consumes task configuration. Under the hood, this method just [invokes](https://github.com/eclipse-theia/theia/blob/df1e55a882e3dff7b2d0a9984fd6e8ae890f85d5/packages/plugin-ext/src/main/browser/tasks-main.ts#L148) [runTask](https://github.com/eclipse-theia/theia/blob/df1e55a882e3dff7b2d0a9984fd6e8ae890f85d5/packages/task/src/browser/task-service.ts#L400) method of Theia task service, which process given configuration and executes the task. So, if a task has a problem matcher in its configuration - the matcher will be applied. If it doesn't - it doesn't. I retrieve task configuration using [fetchTasks](https://github.com/eclipse-theia/theia/blob/df1e55a882e3dff7b2d0a9984fd6e8ae890f85d5/packages/plugin/src/theia.d.ts#L8425) API method which returns all available tasks of Che type. Everything what is changed is the target container name, and it is done to avoid select container popup. The only difference between invocations from `Workspace` panel and `Terminal` -> `Run Task` is that if use Task menu and a task doesn't have a problem matcher then a user will be asked to select one. In case of workspace panel user won't be asked to select a problem matcher if task doesn't have one in its configuration.
Now let's look at which commands we have in the `Workspace` panel. All of them are coming from workspace devfile, so they are designed to be complete and do not bother user with questions about problem matcher or anything else. They have to do their job. And if a task from the `Workspace` panel should have a problem matcher, I believe, that the right place for defining them is the workspace devfile.

@l0rd @slemeur @benoitf WDYT?

RomanNikitenko:
@mmorhun
thank you for clarification! 

Problem matchers for tasks was added to theia project recently (about 2 weeks) and at the moment user doesn't have ability to define problem matchers for actions with type `exec` in devfile (but, I guess, user can add it for action with type `vscode-task`), so list of problem matchers is displayed before running a task first time, then user's choice is saved to configuration and list of problem matchers is not displayed for the task.

So, to be exact - yes, the PR changes the current behavior for first time running task without problem matchers from `Workspace` panel.

mmorhun:
@slemeur please take a look

mmorhun:
crw-ci-test

PR #442, opened 2019-09-19
Add ability to expand zipped projects defined in a devfile

akurinnoy:
### What does this PR do?

This PR improves the `factory-plugin` in order to correctly handle zipped projects defined in a devfile.

fix https://github.com/eclipse/che/issues/14206


plugins/factory-plugin/src/theia-commands.ts
```diff
@@ -28,7 +29,33 @@ function isDevfileProjectConfig(project: cheApi.workspace.ProjectConfig | cheApi
...
+        return;
+    }
+
+    switch (project.source.type) {
+        case 'git':
+            return new TheiaGitCloneCommand(project, projectsRoot);
+        case 'zip':
+            return new TheiaZipCloneCommand(project, projectsRoot);
+        default:
+            const message = `Project type "${project.source.type}" is not supported yet.`;
```
AndrienkoAleksandr: Maybe without `yet`? With `yet` it sounds like we are promising to support any source type.

akurinnoy: Agree, will fix. Thanks

plugins/factory-plugin/src/theia-commands.ts
```diff
@@ -117,6 +144,50 @@ export class TheiaCloneCommand {
...
+        if (!this.locationURI) {
+            return new Promise(() => { });
+        }
+
+        const clone = async (progress: theia.Progress<{ message?: string; increment?: number }>, token: theia.CancellationToken): Promise<void> => {
+            try {
+                const zipfilePath = await zip.execWget(this.locationURI, this.folder, this.zipfile);
+                await zip.execZip(this.folder, zipfilePath);
+            } catch (e) {
+                theia.window.showErrorMessage(`Couldn't download ${this.locationURI}: ${e.message}`);
```
AndrienkoAleksandr: Maybe for error we should use token: theia.CancellationToken to stop progressor?

plugins/factory-plugin/src/theia-commands.ts
```diff
@@ -117,6 +144,50 @@ export class TheiaCloneCommand {
...
+                console.error(`Couldn't download ${this.locationURI}`, e);
+            }
+        };
+
+        return theia.window.withProgress({
+            location: theia.ProgressLocation.Notification,
+            title: `Cloning ${this.locationURI} ...`
```
AndrienkoAleksandr: Maybe unzipping or extraction?

plugins/factory-plugin/src/theia-commands.ts
```diff
@@ -28,7 +29,33 @@ function isDevfileProjectConfig(project: cheApi.workspace.ProjectConfig | cheApi
     return !!project.name && !!project.source && !!project.source.type && !!project.source.location && !project.source['parameters'];
 }
 
-export class TheiaCloneCommand {
+export interface TheiaCloneCommand {
+    execute(): PromiseLike<void>;
+}
+
+export function getTheiaCloneCommand(
```
sunix: maybe rename it `getProjectImportCommand` or `createProjectImportCommand` or `buildProjectImportCommand` or `generateProjectImportCommand`
but zip is not about cloning so we should avoid using the word clone imho

plugins/factory-plugin/src/zip.ts
```diff
@@ -0,0 +1,70 @@
...
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ **********************************************************************/
+
+const { spawn } = require('child_process');
+const path = require('path');
+const fs = require('fs');
+
+export async function execZip(folder: string, zipfile: string): Promise<void> {
```
sunix: As far as I remember, we already have existing generic code to invoke spawn (we use it for git) so maybe you could reuse it here.

plugins/factory-plugin/src/theia-commands.ts
akurinnoy: AFAIK, cancellation token is used to check or get notified when long-running command is cancelled by user.

plugins/factory-plugin/src/theia-commands.ts
akurinnoy: import?

plugins/factory-plugin/src/zip.ts
```diff
@@ -0,0 +1,70 @@
...
+                console.error(`Child process 'rm' stderr: ${data}`);
+            });
+        });
+    });
+}
+
+export async function execWget(locationURI: string, folder: string, zipfile: string): Promise<string> {
+    return new Promise<string>((resolve, reject) => {
+        fs.mkdirSync(folder, { recursive: true });
+        const zipfilePath = path.join(folder, zipfile);
```
AndrienkoAleksandr: Maybe we should download zip to the /tmp folder and use something like https://nodejs.org/api/fs.html#fs_fs_mkdtemp_prefix_options_callback

plugins/factory-plugin/src/theia-commands.ts
akurinnoy: thanks, Sun
`buildProjectImportCommand` sounds good to me

plugins/factory-plugin/src/theia-commands.ts
AndrienkoAleksandr: > importing
Sounds good.

plugins/factory-plugin/src/zip.ts
akurinnoy: I guess you're talking about generator: 
[generator/src/command.ts](https://github.com/eclipse/che-theia/blob/50767320ae233bad5109d308e9b87d52b97e8ea1/generator/src/command.ts)

I don't see such generic code in `factory-plugin`.

plugins/factory-plugin/src/zip.ts
akurinnoy: good point! I will update the PR

plugins/factory-plugin/src/zip.ts
sunix: I was talking about 
https://github.com/eclipse/che-theia/blob/master/plugins/factory-plugin/src/git.ts#L57
but looks like it is not as generic as it could be ...

sunix: but you could try to make it a bit more tiny, you have 2 commands that return the quite the same error message and do quite the same thing

akurinnoy: OK, I will refactor this part, 
thanks for review!

plugins/factory-plugin/src/theia-commands.ts
```diff
@@ -117,6 +144,50 @@ export class TheiaCloneCommand {
...
+
+    constructor(project: cheApi.workspace.ProjectConfig | cheApi.workspace.devfile.Project, projectsRoot: string) {
+        if (isDevfileProjectConfig(project)) {
+            const source = project.source;
+
+            this.locationURI = source.location;
+            this.folder = path.join(projectsRoot, project.name);
+            const zipfileRE = /[^\/]+(?:\.zip)$/i;
+            const match = this.locationURI.match(zipfileRE);
+            this.zipfile = match ? match[0] : `${project.name}.zip`;
```
olexii4: We can simplify this code 
```
            const zipfileRE = /[^\/]+(?:\.zip)$/i;
            const match = this.locationURI.match(zipfileRE);
            this.zipfile = match ? match[0] : `${project.name}.zip`;
```
because it is just a temporary file, which is going to delete. Just use ```this.zipfile = `${project.name}.zip`;```

akurinnoy: you're right, will update the PR

PR #443, opened 2019-09-19
Update go version to 1.12.9

tolusha:
### What does this PR do?
It updates go version to 1.12.9.

https://github.com/eclipse/che/issues/13803

### Related PRs
https://github.com/eclipse/che-devfile-registry/pull/96
https://github.com/eclipse/che-plugin-registry/pull/224


tolusha:
@benoitf 
Could you validate the PR pls?

@amisevsk 
I am wondering if we have to keep original `dockerfiles/remote-plugin-go-1.10.7` ?

amisevsk:
> I am wondering if we have to keep original `dockerfiles/remote-plugin-go-1.10.7` ?

I'm not sure the policy on the che-theia side, but it's worth noting that we should go through a deprecation procedure for images in general. We can stop building the image but it should remain available for workspaces that were created using it.

tolusha:
@amisevsk 
Just added dockerfile for go 1.12.9 and excluded from the build dockerfile with go 1.10.7

PR #449, opened 2019-09-25
Fix undefined error

evidolob:
### What does this PR do?
Fix remote plugin/workspace start

eclipse/che#14645


evidolob:
crw-ci-test

benoitf:
@dmytro-ndp there is not even che-server logs, is that happy path is even starting ?
if che-theia is broken at least we should have che server logs

dmytro-ndp:
@benoitf: there is an error in job config, which I am fixing to take into involve plugins update.

PR #450, opened 2019-09-26
Add 'github' project type to factory-plugin

akurinnoy:

### What does this PR do?

Adds missed project type to factory plugin.

fix https://github.com/eclipse/che/issues/14668

### DIFF
--- a/plugins/factory-plugin/src/theia-commands.ts
+++ b/plugins/factory-plugin/src/theia-commands.ts
@@ -46,6 +46,7 @@ export function buildProjectImportCommand(
 
     switch (project.source.type) {
         case 'git':
+        case 'github':
             return new TheiaGitCloneCommand(project, projectsRoot);
         case 'zip':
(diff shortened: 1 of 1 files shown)


sleshchenko:
@l0rd seems for WsAgent and GWT IDE github project type meant that OAuth authentication should be done if there is 403 issue during try to clone a project.
Do you know if we plan to implement the same for Theia? Maybe we already have it but I did not catch.

skabashnyuk:
Can someone explain the logic of this change? Is this a real bug https://github.com/eclipse/che/issues/14668, or we just forget some unknow functionality?

skabashnyuk:
ok. @sleshchenko an @akurinnoy explained to me. It comes from https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-factory-github/src/main/java/org/eclipse/che/api/factory/server/github/GithubSourceStorageBuilder.java#L62

PR #451, opened 2019-09-27
Use 0.11.0 upstream theia before fixing breaking changes in upstream master

benoitf:
### What does this PR do?
Use theia 0.11.0 before fixing code due to upstream changes

https://github.com/eclipse-theia/theia/pull/6252

### DIFF
--- a/extensions/eclipse-che-theia-user-preferences/package.json
+++ b/extensions/eclipse-che-theia-user-preferences/package.json
@@ -12,14 +12,14 @@
     "src"
   ],
   "dependencies": {
-    "@theia/core": "next",
+    "@theia/core": "0.11.0",
     "nsfw": "^1.2.2",
-    "@theia/preferences": "next",
+    "@theia/preferences": "0.11.0",
     "@eclipse-che/theia-plugin-ext": "^0.0.1"
   },
   "devDependencies": {
     "rimraf": "latest",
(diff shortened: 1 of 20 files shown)


extensions/eclipse-che-theia-user-preferences/package.json
```diff
@@ -12,14 +12,14 @@
...
+    "@theia/core": "0.11.0",
     "nsfw": "^1.2.2",
-    "@theia/preferences": "next",
+    "@theia/preferences": "0.11.0",
     "@eclipse-che/theia-plugin-ext": "^0.0.1"
   },
   "devDependencies": {
     "rimraf": "latest",
-    "typescript": "latest"
+    "typescript": "3.1.3"
```
mmorhun: @benoitf could you please explain why do we need to fix typescript version here?

benoitf: @mmorhun because it's tagged everywhere else. I think it's a mistake to have latest there

benoitf: https://github.com/eclipse/che-theia/blob/master/package.json#L13

yes, it was done on purpose. I had invalid typescript dependencies.

we could cleanup later dependencies as it seems a lot of dependencies are in main package.json

PR #456, opened 2019-09-27
make the --skip-tests flag actually skip...

nickboldt:
make the --skip-tests flag actually skip tests

Change-Id: I2137d9c47d7b28b9a8ae8db8f8adfd10281c45df

### DIFF
--- a/dockerfiles/build.include
+++ b/dockerfiles/build.include
@@ -15,7 +15,7 @@ ERROR=${ERROR:-}
 DIR=${DIR:-}
 
 skip_tests() {
-  if [ $SKIP_TESTS = "true" ]; then
+  if [[ $SKIP_TESTS == "true" ]]; then
     return 0
   else
     return 1
--- a/dockerfiles/theia-dev/build.sh
+++ b/dockerfiles/theia-dev/build.sh
@@ -42,6 +42,8 @@ cp "${CHE_THEIA_GENERATOR_PACKAGE}" "${LOCAL_ASSEMBLY_DIR}"
 
 init --name:theia-dev "$@"
 build
-if ! skip_tests; then
+if [[ $SKIP_TESTS == "false" ]]; then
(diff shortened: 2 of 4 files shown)


benoitf:
it is actually skipping tests on my side (also it's the same script in che dockerfiles and it's skipping tests on my side as well)

benoitf:
```bash
$ SKIP_TESTS="false" && if [ $SKIP_TESTS = "true" ]; then echo "skip tests is true"; else  echo "skip tests is false"; fi
skip tests is false
```
```bash
$ SKIP_TESTS="true" && if [ $SKIP_TESTS = "true" ]; then echo "skip tests is true"; else  echo "skip tests is false"; fi
skip tests is true
```

nickboldt:
OK, but for me the `if ! skip_tests; then` was not working, and since I'm a stickler for the `[[ ]]` for instead of the `[ ]` form (for reasons I can't recall but involve long hours cursing at bash scripts), I figured while I'm fixing the `if ! skip_tests; then` stuff that wasn't working, I'd also update from `[ ]` to `[[ ]]` in the other if-checks.

Does the `if ! skip_tests; then` test work for you? It was never skipping tests for me, especially if I was invoking it from a subfolder like `theia/build.sh` or `theia-dev/build.sh` rather than the root `build.sh`.

benoitf:
yes it's skipping test when I run the script `build.sh` in a given dockerfiles folder
I'm not using top level script as it's running yarn, it's invalidating all layers for nothing.

and AFAIK root.sh is calling build.sh without provided args so calling root build.sh like
```
./build.sh
```

or
```
./build.sh compiling with Nick
```
will produce the same result as top level build.sh is ignoring all arguments

benoitf:
closing as now root build.sh can accept and propagate args

dmytro-ndp:
crw-ci-tests

PR #457, opened 2019-09-30
Use of theia upstream master branch

benoitf:
### What does this PR do?
Make remote extension compliant with upstream theia

Some changes were made in Eclipse Theia upstream that broke totally the remote extensions

Fix in upstream performed there: https://github.com/eclipse-theia/theia/pull/6301

https://github.com/eclipse/che/issues/14685

Change-Id: Id5cb613ca44e3e86ed22000ab2dd674f8dce4f91

### DIFF
--- a/extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
+++ b/extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
@@ -16,7 +16,7 @@ import * as path from 'path';
 import { logger } from '@theia/core';
 import { ILogger } from '@theia/core/lib/common';
 import { Emitter } from '@theia/core/lib/common/event';
-import { MAIN_RPC_CONTEXT, PluginDeployer, PluginDeployerEntry, PluginMetadata } from '@theia/plugin-ext';
+import { MAIN_RPC_CONTEXT, PluginDeployer, PluginDeployerEntry, PluginDependencies, DeployedPlugin, PluginEntryPoint } from '@theia/plugin-ext';
 import pluginVscodeBackendModule from '@theia/plugin-ext-vscode/lib/node/plugin-vscode-backend-module';
 import { RPCProtocolImpl } from '@theia/plugin-ext/lib/common/rpc-protocol';
 import { PluginDeployerHandler } from '@theia/plugin-ext/lib/common';
@@ -29,6 +29,7 @@ import pluginRemoteBackendModule from './plugin-remote-backend-module';
 import { TerminalContainerAware } from './terminal-container-aware';
 import { PluginDiscovery } from './plugin-discovery';
 import { PluginReaderExtension } from './plugin-reader-extension';
+import { Deferred } from '@theia/core/lib/common/promise-util';
 
 interface CheckAliveWS extends ws {
     alive: boolean;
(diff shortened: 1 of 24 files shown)


benoitf:
crw-ci-test

benoitf:
@dmytro-ndp it seems happy path check is using the custom image for che-theia but for plug-ins it's using published image on quay.io and not changes from this PR.
As the API has been changed in theia upstream, both che-theia image and plugin-runner-java8, etc images need to be checked at the same time.

How do you think we can proceed to check ?

benoitf:
@dmytro-ndp because this PR will never get green state

benoitf:
here is a link to test with images I built because happy path is not checking correctly with all images:

image

And test on che.openshift.io : http://che.openshift.io/f/?url=https://gist.githubusercontent.com/benoitf/77de8234c275e7f3125d8e6dbc8df23b/raw/cc3cc4af6b62b20d79d9bc0a4014abc257a34aa6/devfile.yaml

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
```diff
@@ -333,9 +334,29 @@ class PluginDeployerHandlerImpl implements PluginDeployerHandler {
...
+
+    async getDeployedBackendPluginIds(): Promise<string[]> {
+        // await first deploy
+        await this.backendPluginsMetadataDeferred.promise;
+        // fetch the last deployed state
+        return [...this.deployedBackendPlugins.keys()];
+    }
+
+    getDeployedPlugin(pluginId: string): DeployedPlugin | undefined {
+        return this.deployedBackendPlugins.get(pluginId);
```
AndrienkoAleksandr: Maybe for this method we need `await this.backendPluginsMetadataDeferred.promise;`  too and then return?

benoitf: I don't see why we would need await ?

it's a copy of 
https://github.com/eclipse-theia/theia/blob/6347c5ccbf016b7cfa70976f69616eb1bb49c5b0/packages/plugin-ext/src/hosted/node/hosted-plugin-deployer-handler.ts#L60-L62
and there is not

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
```diff
@@ -362,9 +380,47 @@ class PluginDeployerHandlerImpl implements PluginDeployerHandler {
...
+    protected async deployPlugin(entry: PluginDeployerEntry, entryPoint: keyof PluginEntryPoint): Promise<void> {
+        const pluginPath = entry.path();
+        try {
+            const manifest = await this.reader.readPackage(pluginPath);
+            if (!manifest) {
+                return;
+            }
+
+            const metadata = this.reader.readMetadata(manifest);
+            if (this.deployedBackendPlugins.has(metadata.model.id)) {
```
AndrienkoAleksandr: Maybe log warning

benoitf: it's also from upstream: 
https://github.com/eclipse-theia/theia/blob/6347c5ccbf016b7cfa70976f69616eb1bb49c5b0/packages/plugin-ext/src/hosted/node/hosted-plugin-deployer-handler.ts#L116

dmytro-ndp:
@benoitf: 
> @dmytro-ndp it seems happy path check is using the custom image for che-theia but for plug-ins it's using published image on quay.io and not changes from this PR.
As the API has been changed in theia upstream, both che-theia image and plugin-runner-java8, etc images need to be checked at the same time.
> How do you think we can proceed to check ? 

I can tested on PR check job with config patched to build custom plugin and to apply it to test workspace.

benoitf:
@dmytro-ndp  In the meantime I've provided a workspace ready to use on che.openshift.io so I think it's ok.

PR #464, opened 2019-10-02
Use theia-plugin-ext from node_modules.

AndrienkoAleksandr:
### What does this PR do?
Use theia-plugin-ext from node_modules. We need this because nexe tool doesn't resolve path with `__dirname`.

Needed for https://github.com/eclipse/che/issues/13387

### DIFF
--- a/extensions/eclipse-che-theia-plugin-ext/src/node/che-plugin-api-provider.ts
+++ b/extensions/eclipse-che-theia-plugin-ext/src/node/che-plugin-api-provider.ts
@@ -22,7 +22,7 @@ export class ChePluginApiProvider implements ExtPluginApiProvider {
                 initFunction: 'initializeApi',
                 initVariable: 'che_api_provider'
             },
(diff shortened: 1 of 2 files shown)


AndrienkoAleksandr:
ci-build-check

benoitf:
I know why you're doing that but you could say in description that it's required as relative paths are not working when using nexe or pkg when trying to build single executable file, etc.

artaleks9:
ci-build-check

artaleks9:
ci-build

AndrienkoAleksandr:
crw-ci-test

AndrienkoAleksandr:
crw-ci-test

AndrienkoAleksandr:
@benoitf happy path passed, I think you waited on the result.

benoitf:
@AndrienkoAleksandr the issue is that it doesn't test extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-backend-module.ts in the current happy path as it's using a predefined image for sidecar (and not the one that is built)

PR #470, opened 2019-10-06
add --image-name: flag so that full image...

nickboldt:
add --image-name: flag so that full image name can be overridden, not just fragments (eg., when we don't want PREFIX-NAME as the container name, but just NAME)

Related to https://github.com/eclipse/che/issues/13765

Change-Id: I91333d8404c7c36fec1d58d4bfe488da7eb3cde9

### DIFF
--- a/dockerfiles/build.include
+++ b/dockerfiles/build.include
@@ -92,6 +92,9 @@ init() {
       --git-ref:*)
         GIT_REF="${1#*:}"
         shift ;;
+      --image-name:*)
+        IMAGE_NAME="${1#*:}"
+        shift ;;
       --*)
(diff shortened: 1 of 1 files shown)


nickboldt:
@benoitf before you reject this, the reason we need this is so that I can build a container called `quay.io/crw/theia-rhel8:2.0` without needing four flags that kind of hack around the image name:

`--organization:quay.io/crw --prefix:theia --name:rhel8 --tag:2.0`

(since the `-rhel8` suffix isn't really the container NAME, it feels like a hack to do the above.)

`--image-name:quay.io/crw/theia-rhel8:2.0` is simpler and more accurate, but should still fall back to the current default upstream approach

PR #471, opened 2019-10-06
allow slashes in ORGANIZATION field by using...

nickboldt:
allow slashes in ORGANIZATION field by using @ for sed separators instead of /

Related to https://github.com/eclipse/che/issues/13765

Change-Id: I69d8d6c6c3c21531e29deaf68e95a8bd7dad14b7


nickboldt:
Related to #470 -- need this change to allow the ORGANIZATION field to include both the registry and the organization. Eventually Theia will be publishing builds to quay.io, not docker.io, right? So we'll need this. :)

benoitf:
ci-build

PR #482, opened 2019-10-12
Update: Display product logo as image instead displaying in text

vitaliy-guliy:
### What does this PR do?
Allows to display product logo as image instead displaying it in text.
To customize Che-Theia product JSON file should have two logo images for dark and light theme. If so, logo images will be shown depending on the current theme.
If product JSON will have only one image, it will be not theme dependent.

Screenshot from 2019-10-11 16-30-25

Screenshot from 2019-10-11 16-30-35

https://github.com/eclipse/che/issues/14845
https://github.com/eclipse/che/issues/14110


vitaliy-guliy:
@benoitf in what case?
In welcome plugin it works fine using CSS. We don't have theme plugin API to handle switching the theme.
There is one subscription on changing the theme in About dialog. It works fine.

benoitf:
@vitaliy-guliy also before merging, a PR to documentation needs to be done for updated changes

vitaliy-guliy:
@benoitf let's merge a document after mergin this PR, as it will use an example for the Code Ready Workspaces, which will use the feature provided by this PR

vitaliy-guliy:
I need `@eclipse/che-theia:next` container image with this feature

vitaliy-guliy:
Draft PR to che-docs https://github.com/eclipse/che-docs/pull/843

PR #488, opened 2019-10-15
Modify 'configStorage' path for remote plugin

evidolob:
### What does this PR do?
Modify 'configStorage' path for remote plugin, as `/home/theia/` may not present in remote plugin image

eclipse/che#14797

### DIFF
--- a/extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
+++ b/extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
@@ -13,10 +13,11 @@ import * as http from 'http';
 import * as ws from 'ws';
 import * as fs from 'fs';
 import * as path from 'path';
+import * as os from 'os';
 import { logger } from '@theia/core';
 import { ILogger } from '@theia/core/lib/common';
 import { Emitter } from '@theia/core/lib/common/event';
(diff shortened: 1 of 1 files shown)


extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
```diff
@@ -99,6 +105,15 @@ export class PluginRemoteInit {
 
         this.pluginReaderExtension = inversifyContainer.get(PluginReaderExtension);
 
+        // Modify 'configStorage' objects path, to use current user home directory
+        // in remote plugin image '/home/theia' doesn't exist
+        const originalStart = PluginManagerExtImpl.prototype.$start;
+        PluginManagerExtImpl.prototype.$start = async function (params: PluginManagerStartParams): Promise<void> {
```
tsmaeder: This looks like a hack: if the storage path is not in fact determined from the parameters, we should refactor the plugin API to make the  storage path the responsiblity of the plugin runner. 
I understand that we're trying to get this in urgently, so we do a solution in che-theia, but could we open an issue on the theia repo to do the refactoring?

evidolob:
Yes, it required for 7.3.0, I'm going to create issue in Theia repository

PR #494, opened 2019-10-17
Fix publish images function

evidolob:
### What does this PR do?
Fix publish images function


benoitf:
I think shellcheck added commas around the full line

evidolob:
ci-build

PR #499, opened 2019-10-18
Start remote endpoint binary using entrypint.sh script with UID fix.

AndrienkoAleksandr:
### What does this PR do?
https://github.com/eclipse/che/issues/13387

Start remote endpoint binary using entrypint.sh script with UID fix.

### DIFF
--- a/dockerfiles/theia-endpoint-runtime-binary/src/entrypoint.sh
+++ b/dockerfiles/theia-endpoint-runtime-binary/src/entrypoint.sh
@@ -1,0 +1,27 @@
+#!/bin/sh
+# Copyright (c) 2019 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+set -e
+set -x
+
+# Ensure $HOME exists when starting
+if [ ! -d "${HOME}" ]; then
+    mkdir -p "${HOME}"
+fi
+
+# Add current (arbitrary) user to /etc/passwd
+if ! whoami >/dev/null 2>&1; then
+    if [ -w /etc/passwd ]; then
+        echo "${USER_NAME:-user}:x:$(id -u):0:${USER_NAME:-user} user:${HOME}:/bin/sh" >> /etc/passwd
+    fi
+fi
+
+base_dir=$(cd "$(dirname "$0")"; pwd)
+
+exec "${base_dir}"/plugin-remote-endpoint
--- a/dockerfiles/theia-endpoint-runtime-binary/Dockerfile
+++ b/dockerfiles/theia-endpoint-runtime-binary/Dockerfile
@@ -30,5 +30,6 @@ RUN nexe -v && \
 #{INCLUDE:docker/${BUILD_IMAGE_TARGET}/runtime-post-env.dockerfile}
 
 COPY --from=builder /home/theia/plugin-remote-endpoint /plugin-remote-endpoint
+COPY src/entrypoint.sh /
 
-ENTRYPOINT cp -rf /plugin-remote-endpoint /remote-endpoint/plugin-remote-endpoint
+ENTRYPOINT cp -rf /plugin-remote-endpoint /entrypoint.sh /remote-endpoint/


dockerfiles/theia-endpoint-runtime-binary/src/entrypoint.sh
```diff
@@ -0,0 +1,27 @@
...
+# Add current (arbitrary) user to /etc/passwd
+if ! whoami >/dev/null 2>&1; then
+    if [ -w /etc/passwd ]; then
+        echo "${USER_NAME:-user}:x:$(id -u):0:${USER_NAME:-user} user:${HOME}:/bin/sh" >> /etc/passwd
+    fi
+fi
+
+base_dir=$(cd "$(dirname "$0")"; pwd)
+
+exec "${base_dir}"/plugin-remote-endpoint
```
benoitf: hello, how can you be sure that /etc/passwd is writeable ?

I mean it's too late BTW if it's not the case.

I would say that this endpoint is part of the devfile 'uid images fix' but it's not part of che-theia endpoint

dockerfiles/theia-endpoint-runtime-binary/src/entrypoint.sh
```diff
@@ -0,0 +1,27 @@
...
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+set -e
+set -x
+
+# Ensure $HOME exists when starting
+if [ ! -d "${HOME}" ]; then
+    mkdir -p "${HOME}"
```
mmorhun: Do we always have permissions to create home directory?

dockerfiles/theia-endpoint-runtime-binary/src/entrypoint.sh
AndrienkoAleksandr: > But how does it work when image entrypoint is already writing data to passwd file.

 I think whoami returns than user name and we don't touch /etc/passwd.

dockerfiles/theia-endpoint-runtime-binary/src/entrypoint.sh
AndrienkoAleksandr: Yes, but we already use this approach in the devfile registry images: https://github.com/eclipse/che-devfile-registry/blob/master/arbitrary-users-patch/entrypoint.sh#L4

dockerfiles/theia-endpoint-runtime-binary/src/entrypoint.sh
benoitf: in which case do we need it ? If images are patched this is not used and if images are not patched it will do nothing as passwd is not writable

AndrienkoAleksandr: Image could contains own entrypoint.sh. We override entrypoint to start remote binary and this script doesn't start.  Plugin writer can fix it in the meta.yaml definition: simply changing command to start entrypoint.sh and than start remote binary using env. For me it's looks good and pretty configurable, but not all developers happy of that, see discussion in the pr: https://github.com/eclipse/che-plugin-registry/pull/233

benoitf: @AndrienkoAleksandr the question is why we need to override entrypoint and not executing command after container is available ?
I mean entrypoint could do a lot of stuff and responsibility is to image creator.

AndrienkoAleksandr: > why we need to override entrypoint and not executing command after container is available 

There are some images which doesn't has entrypoint at all, but we want to use them like sidecar image with remote plugin. So override command is pretty good for this case. And we don't know what is CMD and Entrypoint content in the Dockerfile, so we can execute it and automate this process. Image creator knows or should know about Entrypoint and CMD and if he really need to start some extra stuff, he can do it the plugin meta.yaml: meta.yaml#Containers#Container[i]#(Command or Args).

**But actually there is one way when we can start entrypoint.sh and then start remote binary, but for me it's looks to much tricky**:

So, let's skip guessing where is located entrypoint.sh. Some our images uses wrapper script pattern(some code-ready images, and images for dev containers https://github.com/eclipse/che-devfile-registry/blob/master/arbitrary-users-patch/entrypoint.sh). Such scripts ussualy has such content:

/some-path/entrypont.sh
```shell
#!/bin/sh

# fix some permission

exec "$@" # to start command defined in the Dockerfile CMD
``

Dockerfile
```
### some usefull stuff 

ENTRYPOINT [ "/entrypoint.sh" ]
CMD ["some", "command"]
```

 Currently remote-runtime injection override command(ENTRYPOINT in the Dockerfile), so this entrypoint.sh script never happen, only if sidecar writer defines in the meta.yaml: start entrypoint.sh and than remote binary: `/entrypoint.sh; {PLUGIN_REMOTE_ENDPOINT_EXECUTABLE}`. But we can change this behavior and override args(CMD in the Dockerfile), than if image has entrypoint.sh wrapper script, this script will be launched and than script starts remote binary by `exec "$@"`. Because our CMD will be overrided to start remote runtime {PLUGIN_REMOTE_ENDPOINT_EXECUTABLE}. 

 For all another cases plugin writer could take a look image and consider how to override command(ENTRYPOINT) with args(CMD) in the plugin meta.yaml or write can hardcode in the image start remote binary using env varibles. Maybe plugin writer wants to start some endless process with service for VSCode extension and detach it and then start {PLUGIN_REMOTE_ENDPOINT_EXECUTABLE}.

benoitf: I think the plugin writer should only add an "infinite command" like `tail -f /dev/null` if the container image is not living forever. This is what users do on their devfile when picking up external images.

But IMHO plugin's writer should not care about how injection of Che-Theia remote endpoint binary is managed.
And in another hand, endpoint should respect what has been defined by image creator.

It means also that che can change the logic without impacting at all all plug-ins.

AndrienkoAleksandr: So, what do you propose?

benoitf: If creating home directory is mandatory for the endpoint, it should be done by plugin-remote code
Uid fix should be done elsewhere ( if images need to be patched)
So no custom entrypoint there, enhance plugin-remote.ts checks that will be included in the all-in-one binary. And if some of requirements are not there, binary should exit with exit code and error message

dockerfiles/theia-endpoint-runtime-binary/src/entrypoint.sh
amisevsk: We do directly ensure group write permissions to the `/home` directory in the arbitrary patch [dockerfile](https://github.com/eclipse/che-devfile-registry/blob/d38c074eb895d20987c8e5820ea14b951bda0af3/arbitrary-users-patch/Dockerfile#L6)

On OpenShift, the requirement is that any user-writable files are owned by the root group and have group read/write/execute permissions.

PR #508, opened 2019-10-23
[Remote plugins] Call OnDidDeploy only if client is set

mmorhun:
### What does this PR do?
In remote plugins mechanism we send some meta information between main Theia and remote plugin host. This PR provides the fix for the case when client has not connected yet, but we've got a message to send. It is safe to not to send the info, because when client is ready it queries the info itself.

https://github.com/eclipse/che/issues/14942

### DIFF
--- a/extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts
+++ b/extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts
@@ -25,7 +25,7 @@ type PromiseResolver = (value?: Buffer) => void;
 @injectable()
 export class HostedPluginRemote {
 
-    private client: HostedPluginClient;
+    private client: HostedPluginClient | undefined;
 
     @inject(ILogger)
     protected readonly logger: ILogger;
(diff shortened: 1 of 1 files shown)


mmorhun:
crw-ci-test

extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts
```diff
@@ -160,7 +160,9 @@ export class HostedPluginRemote {
                 const entryName = getPluginId(deployedPlugin.metadata.model);
                 if (!this.hostedPluginMapping.getPluginsEndPoints().has(entryName)) {
                     this.hostedPluginMapping.getPluginsEndPoints().set(entryName, jsonMessage.endpointName);
-                    this.client.onDidDeploy();
+                    if (this.client) {
```
olexii4: I think it is better:
```
if (this.endpointsSockets.has(endpointAdress)) {
  return;
}
this.hostedPluginMapping.getPluginsEndPoints().set(entryName, jsonMessage.endpointName);
if (this.client === undefined) {
    return;
}
this.client.onDidDeploy();
```

mmorhun: @olexii4 could you please elaborate on your suggested code?

olexii4: I just prefer ```if (this.client !== undefined) {``` instead of ```if(this.client) {``` as an alternative(more obvious as for me). 
Let it be as is.

mmorhun: I am not sure that it is clearer...

benoitf:
FYI ci-build check is failing

PR #514, opened 2019-10-25
Prevent remote plugin nodejs instance crash in case of unhandled exception inside a plugin

mmorhun:

### What does this PR do?
This PR adds handling of unhandled promises rejection (like it is already done on Theia side) in Che Theia remote plugin nodejs instance. The fix prevents crash of the remote plugin runtime in case if a plugin throws an exception to plugin API code.

The problem was found while working on https://github.com/eclipse/che/issues/14942

### DIFF
--- a/extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote.ts
+++ b/extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote.ts
@@ -20,6 +20,39 @@ process.on('SIGINT', () => {
     process.exit();
 });
 
+process.on('uncaughtException', (err: Error) => {
+    console.error('Remote plugin node: got an uncaught exception', err);
+});
+
+// tslint:disable-next-line: no-any
+const unhandledPromises: Promise<any>[] = [];
+
+// tslint:disable-next-line: no-any
+process.on('unhandledRejection', (reason: any, promise: Promise<any>) => {
+    unhandledPromises.push(promise);
+    setTimeout(() => {
+        const index = unhandledPromises.indexOf(promise);
+        if (index >= 0) {
+            promise.catch(err => {
+                unhandledPromises.splice(index, 1);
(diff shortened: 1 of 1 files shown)


extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote.ts
```diff
@@ -20,6 +20,38 @@ process.on('SIGINT', () => {
...
+
+// tslint:disable-next-line: no-any
+process.on('unhandledRejection', (reason: any, promise: Promise<any>) => {
+    unhandledPromises.push(promise);
+    setTimeout(() => {
+        const index = unhandledPromises.indexOf(promise);
+        if (index >= 0) {
+            promise.catch(err => {
+                unhandledPromises.splice(index, 1);
+                console.error(`Remote plugin node: promise rejection not handled in one second: ${err}`);
```
benoitf: I think we can remove the 'node' word there as it can be confusing
We could add container name

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote.ts
```diff
@@ -20,6 +20,38 @@ process.on('SIGINT', () => {
...
+                if (err.stack) {
+                    console.error(`Remote plugin node: promise rejectionh stack trace: ${err.stack}`);
+                }
+            });
+        }
+    }, 2000);
```
akurinnoy: > rejection not handled in ***one*** second

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote.ts
mmorhun: ok

mmorhun:
crw-ci-test

mmorhun:
Merging as happy path is broken at the moment and this is an urgent fix.

PR #516, opened 2019-10-25
to move additional step to crw-theia, need a...

nickboldt:
to move additional step to crw-theia, need a new include step

Change-Id: I40518fd6c85781cf04a33a597142ff9a8204e95e


nickboldt:
Good call, didn't think to add a second zero-length file. Maybe the generator could be more forgiving if the file doesn't exist so we don't need empty files? Anyway PR updated. Thx for catching that.

nickboldt:
crw-ci-test

benoitf:
ci-build

PR #518, opened 2019-10-25
in order to be able to use...

nickboldt:
in order to be able to use updateBaseImages.sh integration in Jenkins jobs to automatically update to newer image tags, we need the URL from which to get the image tag info; also set a tag for ubi-minimal -- can't just use latest or else builds are not reproduceable

Change-Id: I6680871b856c756a04d60f66c83f2ae445315b67


nickboldt:
crw-ci-test

nickboldt:
gotta love that a no-op change like this is blocked by failing e2e test. :(

benoitf:
It's blocked as another committer need to approve

benoitf:
ci-build

PR #485, opened 2019-10-15
Rework images for remote plugins

evidolob:
### What does this PR do?
Rewrites remote plugins images, to avoid extending  `theia-endpoint-runtime` image and using new remote plugin runtime binary injection.

### What issues does this PR fix or reference?
 eclipse/che#14797

PR #521, opened 2019-10-29
Disable building and publishing of 'eclipse/che-theia-endpoint-runtime' image

evidolob:
### What does this PR do?
Disable building and publishing of `eclipse/che-theia-endpoint-runtime` image
After https://github.com/eclipse/che-theia/pull/485 PR we use `eclipse/che-theia-endpoint-runtime-binary` image to provide remote plugin runtime.

### DIFF
--- a/build.include
+++ b/build.include
@@ -56,7 +56,6 @@ is_publish_images() {
 PR_IMAGES=(
   dockerfiles/theia-dev
   dockerfiles/theia
-  dockerfiles/theia-endpoint-runtime
   dockerfiles/theia-endpoint-runtime-binary
 )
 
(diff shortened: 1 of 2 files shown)


dmytro-ndp:
Sorry for silly question, but what about https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/redhat/vscode-yaml/ plugin which is using `eclipse/che-theia-endpoint-runtime`?

benoitf:
it could be switched to https://github.com/eclipse/che-theia/blob/master/dockerfiles/remote-plugin-node/Dockerfile

evidolob:
Yes, I miss that, I'm going to create PR for plugin-registry

dmytro-ndp:
crw-ci-test

PR #527, opened 2019-10-31
Provide ability to edit URL in the preview panel

RomanNikitenko:
### What does this PR do?
Override `toolbar` property for mini browser to provide ability to edit URL in the preview panel.

**Note:** the PR does preview URL editable, but does not resolve security related problems, so user can run into a security exception due to CORS.

https://github.com/eclipse/che/issues/14829

### How to test
1. Use the following devfile, which contains the reference for `docker image: maxura/che-theia:527` related to the current PR.

<details>
<summary>Devfile</summary>

```
apiVersion: 1.0.1-beta
metadata:
 name: test-editable-preview-url
projects:
  - name: dummy-http-server
    source:
      type: git
      location: "https://github.com/RomanNikitenko/dummy-http-server.git"
components:
  - 
    alias: theia-editor
    reference: >-
      https://raw.githubusercontent.com/RomanNikitenko/che-plugin-registry/master/v3/plugins/eclipse/che-theia/custom/meta.yaml
    type: cheEditor
  
  -
    type: dockerimage
    image: quay.io/eclipse/che-golang-1.10:nightly
    alias: go-cli
    mountSources: true
    memoryLimit: 256Mi

commands:

- name: test editable preview url
  previewUrl:
    port: 8081
    path: "/hello"
  actions:
      -
        type: exec
        component: go-cli
        command: "go run main.go 8081 /hello"
        workdir: ${CHE_PROJECTS_ROOT}/dummy-http-server

```
</details>

2. Go to `Terminal => Run Task` and run `test editable preview url` task
3. Use `Preview` button to open the URL in `Preview` panel
4. You can see `Hello World!` message by default
5. Replace `hello` by `test1` in the URL and press `Enter`
6. Replace `test1` by `test2` in the URL and press `Enter` 
7. You can test `F1 => Preview: Open URL` command as well.

### DIFF
--- a/extensions/eclipse-che-theia-plugin-ext/package.json
+++ b/extensions/eclipse-che-theia-plugin-ext/package.json
@@ -16,6 +16,7 @@
     "@eclipse-che/plugin": "0.0.1",
     "@eclipse-che/workspace-client": "latest",
     "@theia/core": "next",
+    "@theia/mini-browser": "next",
     "@theia/plugin-ext": "next",
     "axios": "0.18.1",
     "js-yaml": "3.13.1"
--- a/extensions/eclipse-che-theia-plugin-ext/src/browser/che-frontend-module.ts
+++ b/extensions/eclipse-che-theia-plugin-ext/src/browser/che-frontend-module.ts
@@ -36,8 +36,10 @@ import { ChePluginMenu } from './plugin/che-plugin-menu';
 import { ChePluginCommandContribution } from './plugin/che-plugin-command-contribution';
 import { bindChePluginPreferences } from './plugin/che-plugin-preferences';
 import { CheSideCarContentReaderRegistryImpl, CheSideCarResourceResolver } from './che-sidecar-resource';
+import { CheMiniBrowserOpenHandler } from './che-mini-browser-open-handler';
+import { MiniBrowserOpenHandler } from '@theia/mini-browser/lib/browser/mini-browser-open-handler';
 
-export default new ContainerModule(bind => {
+export default new ContainerModule((bind, unbind, isBound, rebind) => {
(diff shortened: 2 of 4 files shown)


RomanNikitenko:
crw-ci-test

RomanNikitenko:
@mmorhun 
thank you very much for testing!

To summarize you comment - the problem is: user can not select **dedicated segment** of URL using mouse (**whole** URL is selected), but can do it using keyboard .
Is it correct?

If so, I think we can merge the PR and consider fixing the described case in another issue.

@slemeur @mmorhun 
please let me know if you prefer fixing the case within the current PR.

mmorhun:
I do not mind fixing it in another PR.

l0rd:
@mmorhun @RomanNikitenko the URL selection on mouse click works as on my chrome URL bar so we should not change it.

mmorhun:
@l0rd in case of single click yes, but I meant selection some part of url (a path segment or a query parameter) with mouse. And it is not possible in Theia preview browser now (it will still select whole url unlike Chrome which works correctly). But this is minor issue I would say.

l0rd:
>
>
> @l0rd <https://github.com/l0rd> in case of single click yes, but I meant
> selection some part of url (a path segment or a query parameter) with
> mouse. And it is not possible in Theia preview browser now (it will still
> select whole url unlike Chrome which works correctly). But this is minor
> issue I would say.
>

Strange. I was able to select just part of the URL on second click.

RomanNikitenko:
I understand what @mmorhun means:
- first click: whole URL is selected
- second click: user can edit the URL, can select a dedicated segment of URL using keyboard 
- whole URL is selected when user selects a dedicated segment of URL using mouse

please see https://youtu.be/8Z2dlflL9OY

PR #528, opened 2019-11-01
Use Theia logging support and make log listeners recursion-safe

tsmaeder:
### What does this PR do?
* Uses regular Theia mechanisms for logging (configurable via cli contribution)
* Should be safe from stack overflow when logging from a log listener

https://github.com/eclipse/che/issues/14978

#### Docs PR
I am reworking Contributing.md and would include the ability to configure logging as part of that work.

### DIFF
--- a/extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
+++ b/extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
@@ -15,8 +15,8 @@ import * as ws from 'ws';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as os from 'os';
+import * as util from 'util';
 import { logger } from '@theia/core';
-import { ILogger } from '@theia/core/lib/common';
 import { Emitter } from '@theia/core/lib/common/event';
 import { MAIN_RPC_CONTEXT, PluginDeployer, PluginDeployerEntry, PluginDependencies } from '@theia/plugin-ext';
 import { DeployedPlugin, PluginEntryPoint, PluginManagerStartParams, PluginInfo } from '@theia/plugin-ext';
@@ -26,7 +26,7 @@ import { PluginDeployerHandler, OutputChannelRegistryExt } from '@theia/plugin-e
 import { PluginHostRPC } from '@theia/plugin-ext/lib/hosted/node/plugin-host-rpc';
 import { HostedPluginReader } from '@theia/plugin-ext/lib/hosted/node/plugin-reader';
 import pluginExtBackendModule from '@theia/plugin-ext/lib/plugin-ext-backend-module';
-import { Container, inject, injectable } from 'inversify';
+import { Container, inject, injectable, ContainerModule, interfaces } from 'inversify';
 import { RemoteHostTraceLogger, LogCallback } from './remote-trace-logger';
 import pluginRemoteBackendModule from './plugin-remote-backend-module';
 import { TerminalContainerAware } from './terminal-container-aware';
@@ -34,6 +34,13 @@ import { PluginDiscovery } from './plugin-discovery';
 import { PluginReaderExtension } from './plugin-reader-extension';
 import { Deferred } from '@theia/core/lib/common/promise-util';
 import { PluginManagerExtImpl } from '@theia/plugin-ext/lib/plugin/plugin-manager';
+import { CliManager, CliContribution } from '@theia/core/lib/node';
+import { bindContributionProvider } from '@theia/core/lib/common/contribution-provider';
+import { ILogger, LoggerFactory, Logger, setRootLogger, LoggerName, rootLoggerName } from '@theia/core/lib/common/logger';
+// import { ConsoleLogger } from '@theia/core/lib/common/logger-protocol';
+import { LoggerWatcher } from '@theia/core/lib/common/logger-watcher';
+import { LogLevelCliContribution } from '@theia/core/lib/node/logger-cli-contribution';
+import { ILoggerServer, LogLevel } from '@theia/core/lib/common/logger-protocol';
 
 interface CheckAliveWS extends ws {
     alive: boolean;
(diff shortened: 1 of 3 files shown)


benoitf:
crw-ci-test

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
```diff
@@ -104,9 +118,22 @@ export class PluginRemoteInit {
 
         // bind local stuff
         inversifyContainer.load(pluginRemoteBackendModule);
-
+        inversifyContainer.load(new ContainerModule(bind => {
+            bind(CliManager).toSelf().inSingletonScope();
```
benoitf: AFAIK we don't need cli in remote

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
```diff
@@ -104,9 +118,22 @@ export class PluginRemoteInit {
...
+            this.bindLogger(bind);
+        }));
         inversifyContainer.bind<number>('plugin.port').toConstantValue(this.pluginPort);
 
+        if (argv === undefined) {
+            argv = process.argv;
+        }
+
+        const cliManager = inversifyContainer.get(CliManager);
+        await cliManager.initializeCli(argv!);
```
benoitf: looks like a hack ?

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
```diff
@@ -26,14 +26,21 @@ import { PluginDeployerHandler, OutputChannelRegistryExt } from '@theia/plugin-e
...
 import pluginRemoteBackendModule from './plugin-remote-backend-module';
 import { TerminalContainerAware } from './terminal-container-aware';
 import { PluginDiscovery } from './plugin-discovery';
 import { PluginReaderExtension } from './plugin-reader-extension';
 import { Deferred } from '@theia/core/lib/common/promise-util';
 import { PluginManagerExtImpl } from '@theia/plugin-ext/lib/plugin/plugin-manager';
+import { CliManager, CliContribution } from '@theia/core/lib/node';
+import { bindContributionProvider } from '@theia/core/lib/common/contribution-provider';
+import { ILogger, LoggerFactory, Logger, setRootLogger, LoggerName, rootLoggerName } from '@theia/core/lib/common/logger';
+// import { ConsoleLogger } from '@theia/core/lib/common/logger-protocol';
```
benoitf: if not used, remove

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
```diff
@@ -122,15 +149,16 @@ export class PluginRemoteInit {
             return originalStart.call(this, params);
         };
 
+        this.initWebsocketServer(inversifyContainer.get(ILoggerServer));
```
benoitf: if you need stuff from inversifyContainer let's make it private variable of the class instead of giving to arguments to local methods

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
tsmaeder: Yes we do to be able to configure log levels at startup.

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
tsmaeder: Why?

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
tsmaeder: What is the advantage?

benoitf: you're passing in tons of method a logger while you're always in the same class and the attribute is local, so no need to add extra parameters.

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
benoitf: well, first when we try to use container.get() there is something wrong with dependency injection logic

Then we don't need yargs or cli dependency there.

logic is brought forward by the main theia instance, remote is a delegator.

so if log configuration is defined, it's not through cli.

benoitf:
Hi,

from what I see, plugin-remote process is parsing arguments of the launch parameter while it shouldn't.

it' kind of "zero-conf", configuration is defined by the main theia instance.

So basically if you want to enable some debug level, it should be brought by main theia instance to all plugin-remote nodes.

tsmaeder:
@benoitf I disagree with that "zero-conf" idea. There is some configuration that is determined by the main theia instance, but some is local (for example, the plugin state storage directory can't be in /home/theia). If we don't want to have something configurable in the back-end, we should not configure the cli contribution (in particular, where the set of plugin is involved). 
I don't see how some stuff being determined by the main container implies a ban on command-line configuration. In particular logging needs to happen (and be configured) before we even connect to the main instance.

> So basically if you want to enable some debug level, it should be brought by main theia instance to all plugin-remote nodes.

If I want to debug running a particular plugin, I don't want debug logging on _all_ plugin runners, only one in particular.

benoitf:
while you 'disagree' with the 'zero-conf' instance, it's how it's now implemented. you don't need to provide arguments, node starts, it informs main theia instance about the plug-ins that are provided etc.

So basically if you want to enable some debug level, it should be brought by main theia instance to all plugin-remote nodes.

> If I want to debug running a particular plugin, I don't want debug logging on all plugin runners, only one in particular.

yes but I don't see why it should be applied to ALL runners. I never said that. Just that it's main theia instance that notify nodes their logging level.

tsmaeder:
> while you 'disagree' with the 'zero-conf' instance, it's how it's now implemented. you don't need to provide arguments, node starts, it informs main theia instance about the plug-ins that are provided etc.

Yes, and this PR changes how it's implemented. I still don't see why we should not change this. Can you explain?

benoitf:
> Yes, and this PR changes how it's implemented. I still don't see why we should not change this. Can you explain?

because:
- UX is clearly bad: 
      - I don't want to restart a workspace to change logging for example on a running workspace.
      - I don't need to know how enable logging or have to edit devfile.
 - as node is ran by CMD we may not provide args easily as well.
- so I want something that really work in current design architecture. And this PR is not changing design architecture at all, just making it worst.

evidolob:
I don't like the idea that to change log level I need to rebuild remote plugin docker image or modify plugins meta.yml.

I think we can provide something like remote plugin dev view, on which developer/user can view status of remote plugin runtimes, theirs status(CPU/RAM)  and change log level for specific runtime.

It also can improve troubleshooting if remote plugin runtime is not working for some reason and Theia do not show that.

tsmaeder:
@evidolob nobody says you have to modify the meta.yaml or plugin image. What I'm talking about is the ability to configure logging in development when following our process described in Contributing.md.
We can always add more configuration options through some plugin API: nothing in this PR makes that impossible. But right now, we have no ability at all to configure logging, and even if we did have that API, it would not allow us to configure logging during startup, for example to debug problems in plugin host discovery.

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
tsmaeder: No, container.get is necessary because plugin-remote-init.ts itself is not created by the inversion container.

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
tsmaeder: sure

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
tsmaeder: Passing it as an argument constrains the visiblity of the logger server. I don't think passing parameters is a bad thing in itself, as long as we don't have too many.

benoitf: BTW the whole stuff should be reworked while here it was just a tiny remark.

nickboldt:
Changes to this PR were requested months ago and never implemented. 

Will this PR be cleaned up and merged into Che 7.9? Or abandoned and closed? 

@tsmaeder @benoitf @azatsarynnyy 

Relevant to https://issues.redhat.com/browse/CRW-614

tsmaeder:
Since @benoitf objects on principle to command line configurability of logging in the plugin host, (and that was one of the points of the PR), I don't think we'll merge this one. Parts could be salvaged, probably, that's why I linked it.

benoitf:
https://issues.redhat.com/browse/CRW-614 has been fixed and linked issue has been closed as staled issue, closing this PR
feel free to reopen when working on it

PR #529, opened 2019-11-01
Add Dockerfile for OpenShift Connector plugin v0.1.1

azatsarynnyy:

### What does this PR do?
Adds Dockerfile to build an image for running OpenShift Connector Che Plugin based on the VS Code extension v0.1.1. The latest version contains an important [fix](https://github.com/redhat-developer/vscode-openshift-tools/pull/1254) for `Open URL` command.
Also, the Dockerfile updates odo from v1.0.0-beta5 to v1.0.0 as it's required by the VS Code extension.

https://github.com/eclipse/che/issues/13406
https://github.com/eclipse/che/issues/13410

### DIFF
--- a/build.include
+++ b/build.include
@@ -79,6 +79,7 @@ DOCKER_FILES_LOCATIONS=(
   dockerfiles/remote-plugin-openshift-connector-0.0.17
   dockerfiles/remote-plugin-openshift-connector-0.0.21
   dockerfiles/remote-plugin-openshift-connector-0.1.0
+  dockerfiles/remote-plugin-openshift-connector-0.1.1
   dockerfiles/remote-plugin-camelk-0.0.9
   dockerfiles/remote-plugin-dependency-analytics-0.0.12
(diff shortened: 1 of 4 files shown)


azatsarynnyy:
@benoitf if you have no objections, could you approve it, please?

benoitf:
hello, do we need a new docker image if it's the same content than for the previous one ?

benoitf:
we can't upgrade odo for all extensions ?

azatsarynnyy:
> do we need a new docker image if it's the same content than for the previous one ?

@benoitf it's not the same

> we can't upgrade odo for all extensions ?

we can't because of each version of extension requires a different version of the odo tool

> it seems there is a ci failure

do you happen to know can I rerun ci-build-check job through a GH comment?

azatsarynnyy:
BTW I think we need to deprecate all the  previous versions of the plugin and remove the corresponding images

azatsarynnyy:
crw-ci-test

PR #534, opened 2019-11-05
Fixes failing unstable tests

vitaliy-guliy:
Signed-off-by: Vitaliy Guliy <vgulyy@redhat.com>

### What does this PR do?
The main reason while tests are failing is timeout when cloning che-theia sources.

To make tests stable, it was decided to:
- increase timeout to 30 seconds for failing tests
- add an option to omit git history when cloning che-theia sources

We can add a command line option for the generator for omitting git history. This improvement we could use when building docker images.

https://github.com/eclipse/che/issues/14276

### DIFF
--- a/generator/tests/init-sources/init-sources.spec.ts
+++ b/generator/tests/init-sources/init-sources.spec.ts
@@ -21,6 +21,9 @@ import { YargsMockup } from "../cdn.spec";
 
 describe("Test Extensions", () => {
 
+    // Increase timeout to 30 seconds
+    jest.setTimeout(30000);
+
     const THEIA_DUMMY_VERSION = '1.2.3';
     const rootFolder = process.cwd();
     const assemblyExamplePath = path.resolve(rootFolder, "tests/init-sources/assembly-example");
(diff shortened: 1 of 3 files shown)


benoitf:
would be good to have offline mode (not using remote che-theia)

generator/tests/init-sources/init-sources.spec.ts
```diff
@@ -21,6 +21,9 @@ import { YargsMockup } from "../cdn.spec";
 
 describe("Test Extensions", () => {
 
+    // Increase timeout to 30 seconds
+    jest.setTimeout(30000);
```
AndrienkoAleksandr: Mabe it's mine own opinion, but  I think you should not Increase timeout for all tests. You can set up longer timeout for one of them:

```
test("test init sources generator", async () => {
// blah-blah some test.
}, 30000);
```

vitaliy-guliy: There are many tests which can failed with timeout

vitaliy-guliy:
@benoitf let it be as improvement. It's another issue.

PR #539, opened 2019-11-13
Use self signed certificate when do request to plugin registry

evidolob:
### What does this PR do?
Fix get plugins from theia-ide does not work if che deployed with switched on TLS

CRW-466


evidolob:
Thx @sleshchenko for investigation 🤝

evidolob:
This fix depends on https://github.com/eclipse/che/issues/15189

PR #548, opened 2019-11-15
[Port plugin] Do not suggest to open link of exposed by workspace config port

mmorhun:

### What does this PR do?
Removes functionality from port plugin which suggest to open link of already opened in workspace configuration port.
This is mainly done because of preview url suggestion when a task is started. In such case two popups are raised.

https://github.com/eclipse/che/issues/14802

### DIFF
--- a/plugins/ports-plugin/src/ports-plugin.ts
+++ b/plugins/ports-plugin/src/ports-plugin.ts
@@ -122,6 +122,12 @@ async function onOpenPort(port: Port) {
         if (matchingWorkspacePort.serverName.startsWith(SERVER_REDIRECT_PATTERN)) {
             return;
         }
+
+        // check if endpoint has preview url, and if so do not show dialog to avoid duplication with task plugin
+        if (matchingWorkspacePort.previewUrl) {
+            return;
+        }
+
         const interactions: MessageItem[] = [{ title: 'Open Link' }];
         const msg = `A process is now listening on port ${matchingWorkspacePort.portNumber}. External URL is ${matchingWorkspacePort.url}`;
         const result = await theia.window.showInformationMessage(msg, { modal: true }, ...interactions);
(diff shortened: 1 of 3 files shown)


benoitf:
What happens if I run with command line the process ?
I guess I'm no longer notified --> not good UX to me.

mmorhun:
Not exactly. I removed notification for open link only (in other words for ports which is exposed by workspace configuration).
I agree that suggesting to open link could be useful, but it also could be a little bit annoying (if I start DB server for example, especially modal window which I cannot ignore). And I am not sure what is better solution here.

Actually, the goal is to avoid duplicate notification. So, if a user starts a task, which has preview url, another notification is shown. I've thought about ability to retrieve tasks configuration and look at preview url (which seems is not trivial at the moment, if not impossible), but it will be guessing anyway...

@benoitf maybe you have some thought on this topic, I would be interested to hear because I do not see 100% clear solution.

benoitf:
clear solution would be to rework port plugin to be a widget (or integrate into a widget) but it's clearly out of range of this PR
My issue is still for ppl importing their project. They don't have preview URL and they'll miss the open link and they have no idea where to get this external route. (I also agree that it can be annoying when you start java in debug mode, etc this is why I'm saying it should have been a widget instead of popups)

So for now the temp solution would be to check if there is in workspace config a preview URL for this port and do not raise notification for this port.

mmorhun:
Unfortunately it is not possible, because `previewUrl` in runtime workspace config doesn't contain actual port number:
```json
"previewUrl": "${previewurl/Testpreviewurl_1433077079}/",
```

benoitf:
@mmorhun then server should also provide the previewURL port
it's not even available in devfile ?

mmorhun:
@benoitf you are right, that's a way to go.
Updated this PR.

mmorhun:
crw-ci-test

mmorhun:
@benoitf could you take a look please?

benoitf:
yes today I'm at an eclipse day event
I looked at thecode yesterday and I'm fine just wanted to check it live

PR #553, opened 2019-11-19
Implement 'StorageService' to store Theia state in workspace

evidolob:
### What does this PR do?
Implements 'StorageService' to store Theia state in workspace. Before Theia uses browser `LocalStorage` to store it state, but in Che after each workspace start html page with Theia has different host, so we loose that state. With this PR we store state inside workspace.

### DIFF
--- a/extensions/eclipse-che-theia-user-preferences/src/node/che-storage-server.ts
+++ b/extensions/eclipse-che-theia-user-preferences/src/node/che-storage-server.ts
@@ -14,8 +14,9 @@ import { readFile, writeFile, ensureDir, pathExists } from 'fs-extra';
 import { injectable, postConstruct } from 'inversify';
 
 const THEIA_STORAGE_PATH = resolve(homedir(), '.theia', 'storage');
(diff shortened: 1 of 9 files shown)


extensions/eclipse-che-theia-user-preferences/src/node/che-storage-server.ts
```diff
@@ -0,0 +1,39 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ **********************************************************************/
+import { StorageServer } from '../common/storage-server';
+import { resolve } from 'path';
+import { homedir } from 'os';
+import { readFile, writeFile, ensureDir, pathExists } from 'fs-extra';
+import { injectable, postConstruct } from 'inversify';
+
+const THEIA_STORAGE_PATH = resolve(homedir(), '.theia', 'storage');
+const ILLEGAL_CHARACTER = /[\/\?<>\\:\*\|"]/g;
```
mmorhun: Maybe `ILLEGAL_CHARACTERS` is a bit better? WDYT?

PR #562, opened 2019-11-26
Intercept uri arguments and set container name into file scheme

vzhukovskii:
### What does this PR do?
Register ArgumentProcessor that is responsible to process only URI argument types. Argument processor checks for the path and if path is not starts with project root, then file scheme modifies with the container name to be looks like this: `file-sidecar-${machineName}`. This will allow che-theia to pass into executed commands correct file resources, taken from different containers.

https://github.com/eclipse/che/issues/13044

#### Release Notes

Intercept uri arguments and set container name into file scheme

#### Docs PR

N/A - internal modification.


dmytro-ndp:
crw-ci-test

musienko-maxim:
Was checked locally. There was not any regression.

PR #563, opened 2019-11-26
Fix webview after changes in upstream.

mmorhun:
Co-authored-by: Oleksandr Andriienko <oandriie@redhat.com>

### What does this PR do?
Fixes webview after changes in upstream Eclipse Theia.

According to changes in upstream Eclipse Theia, each webview is run on its own domain by default. In this PR we reconfigure Theia to run all webviews on the same as Che Theia host (actually as it was before).
Also after upstream changes it is possible to run webviews only with secure connection (requires https protocol), the only exception is localhost for testing purposes.

https://github.com/eclipse/che/issues/15283

### DIFF
--- a/dockerfiles/theia/Dockerfile
+++ b/dockerfiles/theia/Dockerfile
@@ -163,5 +163,6 @@ COPY --from=builder /home/theia-dev/theia-source-code/production/plugins /defaul
 COPY --chown=theia:root --from=builder /home/theia-dev/theia-source-code/production /home/theia
 USER theia
 WORKDIR /projects
+COPY src/get-webview-route.js ${HOME}/get-webview-route.js
 COPY src/entrypoint.sh /entrypoint.sh
 ENTRYPOINT ["/entrypoint.sh"]
--- a/dockerfiles/theia/src/entrypoint.sh
+++ b/dockerfiles/theia/src/entrypoint.sh
@@ -76,7 +76,8 @@ if [ "${NOCDN}" == "true" ]; then
 fi
 shopt -u nocasematch
 
-# run che
+# run Che Theia
+export THEIA_WEBVIEW_EXTERNAL_ENDPOINT=$(node get-webview-route.js)
 node src-gen/backend/main.js /projects --hostname=0.0.0.0 --port=${THEIA_PORT} &
 
 PID=$!
--- a/dockerfiles/theia/src/get-webview-route.js
+++ b/dockerfiles/theia/src/get-webview-route.js
@@ -1,0 +1,73 @@
+/*********************************************************************
+ * Copyright (c) 2019 Red Hat, Inc.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ **********************************************************************/
+
+const fs = require('fs');
+const workspaceClient = require('@eclipse-che/workspace-client');
+
+// Assume that self-signed certificate is located by the following path
+const SS_CRT_PATH = '/tmp/che/secret/ca.crt';
+
+// Configure workspace API client
+const restAPIConfig = {
+    baseUrl: process.env.CHE_API
+};
+
+const token = process.env.CHE_MACHINE_TOKEN;
+if (token) {
+    restAPIConfig.headers = {};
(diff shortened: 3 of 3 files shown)


dockerfiles/theia/src/entrypoint.sh
```diff
@@ -76,7 +76,8 @@ if [ "${NOCDN}" == "true" ]; then
 fi
 shopt -u nocasematch
 
-# run che
+# run Che Theia
+export THEIA_WEBVIEW_EXTERNAL_ENDPOINT=$(node get-webview-route.js)
```
benoitf: question: we don't have the endpoint as environment variable ?

benoitf: I mean, che-server is not providing that information ?

dockerfiles/theia/src/get-webview-route.js
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ **********************************************************************/
+
+const fs = require('fs');
+const workspaceClient = require('@eclipse-che/workspace-client');
```
benoitf: where does it find this dependency ?

dockerfiles/theia/src/entrypoint.sh
mmorhun: No, we don't have external route in env vars.

dockerfiles/theia/src/get-webview-route.js
mmorhun: It gets the dependency from Che Theia node modules.

AndrienkoAleksandr: In the dockerfile we copy script to the /home/theia, where is located node_modules

dockerfiles/theia/Dockerfile
```diff
@@ -163,5 +163,6 @@ COPY --from=builder /home/theia-dev/theia-source-code/production/plugins /defaul
 COPY --chown=theia:root --from=builder /home/theia-dev/theia-source-code/production /home/theia
 USER theia
 WORKDIR /projects
+COPY src/get-webview-route.js ${HOME}/get-webview-route.js
```
benoitf: Should it be possible (maybe later) to use a runtime information instead of using environment variable.

Explanation: Here we're using a shell script before running the theia process in the container (by setting environment variable)

I think it would be cleaner if we bring a hook/extension point of theia at runtime by updating the setting on the fly. So no need to use separate script and fix env variable. It's when theia webview will resolve stuff, it will deal with che-theia code

benoitf: like to override this class: https://github.com/eclipse-theia/theia/blob/fc49b2c3b8a690ba5b94e1fd48cf17469b51057d/packages/plugin-ext/src/main/browser/webview/webview-environment.ts#L35

AndrienkoAleksandr: Yes, I think we can do it in runtime.

mmorhun: I think we should try, but as for now I would like to merge current PR as it is tested and fixes (for https case) blocker issue.

benoitf: yes this is why I said later

benoitf:
image

Here is the image of welcome plug-in (so there is not only icon issue on my side)

AndrienkoAleksandr:
> Here is the image of welcome plug-in (so there is not only icon issue on my side)

Known issue https://github.com/eclipse/che/issues/15324

benoitf:
@AndrienkoAleksandr I know about the icon as I said, my question was more on the font colors/style:

Before:
image

AndrienkoAleksandr:
> @benoitf my question was more on the font colors/style

Yes, Ok, I will take a look what was changed. But from first look welcome plugin used some styles from theia, and am not sure that they are accessible after web-view rework.

benoitf:
@AndrienkoAleksandr sure, we can't get exact same stuff than before.
Just need to check new rendering if @slemeur still like it

AndrienkoAleksandr:
> sure, we can't get exact same stuff than before.

Ok, If @slemeur is ok with current rendering I will close this one https://github.com/eclipse/che/issues/15334

PR #564, opened 2019-11-27
Rebind webview-environment to set up webview domain

AndrienkoAleksandr:
### What does this PR do?
Code improve: Rebind webview-environment to set up webview domain.

https://github.com/eclipse/che/issues/15283

Co-authored-by: Mykola Morhun <mmorhun@redhat.com>


extensions/eclipse-che-theia-plugin-ext/src/browser/che-webview-environment.ts
```diff
@@ -0,0 +1,75 @@
...
+    @inject(EnvVariablesServer)
+    protected readonly environments: EnvVariablesServer;
+
+    @inject(CheApiService)
+    private cheApi: CheApiService;
+
+    @postConstruct()
+    protected async init(): Promise<void> {
+        try {
+            const variable = await this.environments.getValue(WebviewExternalEndpoint.pattern);
```
benoitf: do we still need to allow env variable overrides ? as probably it will never work

extensions/eclipse-che-theia-plugin-ext/src/browser/che-webview-environment.ts
```diff
@@ -0,0 +1,75 @@
...
+            const constainers = await this.cheApi.getCurrentWorkspacesContainers();
```
benoitf:
```suggestion
            const containers = await this.cheApi.getCurrentWorkspacesContainers();
```

extensions/eclipse-che-theia-plugin-ext/src/browser/che-webview-environment.ts
mmorhun: I believe it would be useful for hosted plugin mode.

mmorhun: Maybe with some modifications, though

benoitf: ok

AndrienkoAleksandr:
ci-build

PR #567, opened 2019-11-29
Add patch to che-theia 7.3.2

evidolob:
Add patch to che-theia 7.3.2, with fix Clipboard API


dockerfiles/theia/Dockerfile
```diff
@@ -21,7 +21,7 @@ ENV GITHUB_TOKEN=$GITHUB_TOKEN
 ARG THEIA_GITHUB_REPO=eclipse-theia/theia
 
 # Define upstream version of theia to use
-ARG THEIA_VERSION=f255f5a8cb22010aacb8196c8a8f248e111e9945
+ARG THEIA_VERSION=0.12.0-next.f255f5a8
```
nickboldt: Does this version have to exist somewhere? I don't see it in https://github.com/eclipse-theia/theia/tags Or is it an arbitrary string simply used to define which patch(es) to apply when building che-theia?

benoitf: it's a sha1 commit:
https://github.com/eclipse-theia/theia/commit/f255f5a8

evidolob: Yes, basically it folder name where we looking on patch file, see https://github.com/eclipse/che-theia/blob/02fd13ed377c690292e1d83db5728ea687c9d5a8/dockerfiles/theia/Dockerfile#L41

PR #570, opened 2019-12-02
Recovery welcome plugin styles after changing webview implementation.

AndrienkoAleksandr:
### What does this PR do?
Recovery welcome plugin styles after changing webview implementation.

https://github.com/eclipse/che/issues/13442 
https://github.com/eclipse/che/issues/15334

### DIFF
--- a/plugins/welcome-plugin/resources/welcome-page.css
+++ b/plugins/welcome-plugin/resources/welcome-page.css
@@ -287,8 +287,8 @@
 }
 
 html, body {
-    font-family: var(--theia-ui-font-family);
-    background: var(--theia-layout-color0);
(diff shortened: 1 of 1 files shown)


benoitf:
FYI I've asked @beaumorley about layout for che-theia welcome page layout

AndrienkoAleksandr:
> FYI I've asked @beaumorley about layout for che-theia welcome page layout

After this pr welcome plugin should looks like it was before...

benoitf:
as it was broken, https://github.com/eclipse/che/issues/15334 still being on a backlog and not being 'in-progress' I was just assuming that we could wait the feedback from Beau before fixing the issue

PR #579, opened 2019-12-06
Remove che-theia e2e typescript test

evidolob:
### What does this PR do?
Removes `typescript.spec.ts` test from che-theia e2e tests as `typescript` theia extension is not included in che-theia image(in che we use typescript plugin, with side car image), so this test is not relevant and may cause false build fail.


benoitf:
Hi,
I'm fine to remove native typescript check but we should still keep somehow
- a basic editor test
- or test that VS Code typescript extension is working.

else soon, there is no more e2e at build time

benoitf:
replaced by https://github.com/eclipse/che-theia/pull/593

PR #585, opened 2019-12-12
Deprecate current.project.path variable

vzhukovskii:
### What does this PR do?
For the current state `${current.project.path}` is redundant and with this changes proposal marks as deprecated. User may use `${workspaceFolder}` instead.

https://github.com/eclipse/che/issues/13636

#### Release Notes
Deprecate `${current.project.path}` macro

### DIFF
--- a/plugins/task-plugin/src/variable/project-path-variable-resolver.ts
+++ b/plugins/task-plugin/src/variable/project-path-variable-resolver.ts
@@ -9,17 +9,17 @@
  **********************************************************************/
 
 import { injectable } from 'inversify';
-import Uri from 'vscode-uri';
 import * as che from '@eclipse-che/plugin';
 import * as theia from '@theia/plugin';
 import * as startPoint from '../task-plugin-backend';
 
 const VARIABLE_NAME = 'current.project.path';
-const SELECTED_CONTEXT_COMMAND = 'theia.plugin.workspace.selectedContext';
 const PROJECTS_ROOT_VARIABLE = 'CHE_PROJECTS_ROOT';
 const ERROR_MESSAGE_TEMPLATE = 'Can not resolve \'current.project.path\' variable.';
 /**
  * Contributes the variable for getting path for current project as a relative path to the first directory under the root workspace.
+ * 
+ * @deprecated will be removed soon. Use ${workspaceFolder} macro instead.
  */
 @injectable()
 export class ProjectPathVariableResolver {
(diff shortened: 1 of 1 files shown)


plugins/task-plugin/src/variable/project-path-variable-resolver.ts
```diff
@@ -38,33 +38,7 @@ export class ProjectPathVariableResolver {
             return this.onError('Projects root is not set');
         }
 
-        const selections = await theia.commands.executeCommand<Uri[]>(SELECTED_CONTEXT_COMMAND);
```
RomanNikitenko: Maybe we should not remove this logic if at the moment it's just deprecated? 
For example, user starts a workspace from devfile with two defined projects: `theia` and `che-theia`.
By default user has `/projects` as workspace folder.
According to these changes  `${current.project.path}` returns `/projects`, not  `/projects/theia` for example.
Isn't it?
Please correct me if I'm wrong.

vzhukovskii: This command relies on the current selection. When workspace opens at first time, the selection is not initialized and we're getting error message `... bad substitution ...`. It will return `/project` only when user opened explorer part or some file in editor.

vzhukovskii: `selections` in case of opening workspace at first time will be `undefined`. So it useless and calling `${workspaceFolder}` will return the same result if editor or explorer is opened.

RomanNikitenko: I understand that there are bugs related to using this variable if user has incorrect selection, that's why we decided to use `${workspaceFolder}` instead of `${current.project.path}`.
But if user or our samples rely on `${current.project.path}`, these places will be broken after your changes!
isn't is?

plugins/task-plugin/src/variable/project-path-variable-resolver.ts
```diff
@@ -9,17 +9,17 @@
...
 import * as startPoint from '../task-plugin-backend';
 
 const VARIABLE_NAME = 'current.project.path';
-const SELECTED_CONTEXT_COMMAND = 'theia.plugin.workspace.selectedContext';
 const PROJECTS_ROOT_VARIABLE = 'CHE_PROJECTS_ROOT';
 const ERROR_MESSAGE_TEMPLATE = 'Can not resolve \'current.project.path\' variable.';
 /**
  * Contributes the variable for getting path for current project as a relative path to the first directory under the root workspace.
+ * 
+ * @deprecated will be removed soon. Use ${workspaceFolder} macro instead.
```
RomanNikitenko: I think it's not enough just replace `${workspaceFolder}` by `${current.project.path}`. Instead user should use `${workspaceFolder}/theia` for example

plugins/task-plugin/src/variable/project-path-variable-resolver.ts
vzhukovskii: Our samples don't use this macro. If user uses this macro in own devfile, it will broken anyway and command won't execute. There will be created another PR into docs that describes, that fact, that this macro is deprecated and has to be replaced with another one. So, yes, this changes proposal changes behavior of the flow for this macro.

plugins/task-plugin/src/variable/project-path-variable-resolver.ts
vzhukovskii: There will be created PR into docs that describes that fact, that this macro is deprecated and has to be replaced

plugins/task-plugin/src/variable/project-path-variable-resolver.ts
RomanNikitenko: steps to reproduce:
- start a workspace from a devfile, get project structure like:
pr_structure
- run the task:
```
{
                "type": "che",
                "label": "test2",
                "command": "sleep 3 && echo ${workspaceFolder}",
                "target": {
                    "workingDir": "/projects"
                }
            },
```
output: `/projects`
- select `che-theia` project and run the task:
```
{
                "type": "che",
                "label": "test2",
                "command": "sleep 3 && echo ${current.project.path}",
                "target": {
                    "workingDir": "/projects"
                }
            }
```
output: `/projects/che-theia`

vzhukovskii:
Closing this PR due to rework the mechanism of returning the value for current project path. There will be another PR.

PR #586, opened 2019-12-12
Post message to Dashboard in order to refresh keycloak token.

akurinnoy:

### What does this PR do?

Activity tracker extension sends message to User Dashboard which makes UD to refresh keycloak token so as to avoid User Dashboard reloading. 

fixes https://github.com/eclipse/che/issues/14933


akurinnoy:
Depends on https://github.com/eclipse/che/pull/15465

PR #591, opened 2019-12-29
Use a dedicated server as webview domain

azatsarynnyy:

### What does this PR do?
Uses a dedicated endpoint to deploy all webviews separately from Che Theia editor endpoint.
It increases security since all webviews don't have access to the data of Che Theia window.

**Note**, this PR depends on a couple of other PRs (already merged):
- plugin unique endpoint feature introduced by the PR https://github.com/eclipse/che/pull/15432
- PR https://github.com/eclipse/che-plugin-registry/pull/335 that enables additional endpoint for the webviews

To test it:
- run Che with enabled TLS and using self-signed certificates. For the details, see the comments https://github.com/eclipse/che/issues/14035#issuecomment-519822676 and https://github.com/eclipse/che/issues/14035#issuecomment-562141679. Use Che Server from the image `quay.io/lkrejci/che-server:7.7.0-issue-15283` which includes unique endpoint feature
- run a workspace from the following Devfile:
```yaml
metadata:
  name: webviews
apiVersion: 1.0.0
components:
  - reference: https://raw.githubusercontent.com/azatsarynnyy/che-plugin-registry/master/v3/plugins/eclipse/che-theia/next/meta.yaml
    type: cheEditor

```
- see that theia sidecar has `webviews` endpoint which is used for loading webview of Welcome page

closes https://github.com/eclipse/che/issues/15616


extensions/eclipse-che-theia-plugin-ext/src/node/plugin-service.ts
```diff
@@ -46,22 +46,23 @@ export class PluginApiContributionIntercepted extends PluginApiContribution {
...
-        });
+        this.cheApi.findUniqueServerByAttribute(SERVER_TYPE_ATTR, SERVER_WEBVIEWS_ATTR_VALUE).then(server => {
+                let domain;
+                if (server.url) {
+                    domain = getUrlDomain(server.url);
+                }
+                const hostName = this.handleAliases(process.env[WebviewExternalEndpoint.pattern] || domain || WebviewExternalEndpoint.pattern);
+                webviewApp.use('/webview', serveStatic(webviewStaticResources));
+    
+                console.log(`Configuring to accept webviews on '${hostName}' hostname.`);
```
vzhukovskii: Maybe it would be better to use the logger? How it was done in [hosted-plugin-remote.ts#L30-L31](https://github.com/eclipse/che-theia/blob/master/extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts#L30-L31)
```
import { ILogger } from '@theia/core/lib/common';
...
@inject(ILogger)
protected readonly logger: ILogger;
...
this.logger.debug('...');
```

azatsarynnyy: Thanks for noticing this.
Updated

PR #597, opened 2020-01-08
Implementation of testing api for testing vscode extensions language features

JPinkney:
### What does this PR do?
This PR introduces testing api that allows you to talk to a specific vscode extensions language features. It gives the ability to create tests that call into a vscode language extension for two reasons:
1. Languages team can run tests against newer version of language plugins so we know if a language plugin can be updated or something needs to be fixed
2. Che tests can talk directly with the language server so we know that if the language server request fails then it's an actual failure, rather than a test failing for a different reason and making it look like it was the cause 

Examples of such tests can be found in: https://github.com/jpinkney/che-java-tests though they're WIP.

https://github.com/eclipse/che/issues/14409


extensions/eclipse-che-theia-testing-service-ext/src/common/test-protocol.ts
```diff
@@ -0,0 +1,78 @@
...
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+import * as vst from 'vscode-languageserver-types';
+
+// Expose additional API that allows you to know if a language server is connected and build a map of the language servers
+export interface LanguageMainTestInterface {
+    // tslint:disable-next-line:no-any
+    $provideCompletionItems(pluginID: string, model: monaco.editor.ITextModel, position: monaco.Position,
```
tsmaeder: I don't think we should use stuff from the monaco workspace here. This API should be pretty much the same as LanguagesExt, but with the "handle" replaced by "pluginId".

tsmaeder:
One thing that is not clear to me is how tests are loaded and run.

JPinkney:
I originally experimented with having the test loader as a che theia plugin but over time it felt like it made more sense to go with the strategy where each plugin will load the tests themselves. This allows the tests to have a certain level of configurability, such as how long the tests should try to run before timing out, and gives the additional benefit of not having to lock into using specific testing frameworks. 

I've done some documentation on how the tests will work in https://github.com/jpinkney/che-java-tests#running-these-tests-automatically-when-a-workspace-starts but the basic idea is that you first start a workspace using a devfile that references a theia testing plugin (like https://github.com/jpinkney/che-java-tests). When the workspace starts the testing plugin loads in all the tests and communicates with the testing-api that is available in this PR

extensions/eclipse-che-theia-testing-service-ext/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "theiaExtensions": [
+    {
+      "backend": "lib/node/testservice-ext-backend-module",
+      "frontend": "lib/browser/testservice-ext-frontend-module"
+    }
+  ],
+  "keywords": [
+    "theia-extension"
+  ],
+  "license": "EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0",
```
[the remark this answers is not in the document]
benoitf: I think it may be linked to copy/paste as upstream theia is EPL2+classpath exception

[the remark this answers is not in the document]
tsmaeder: Ok, but since this is 100% our code, let's make it simple and use EPL 2.0, like the rest of our code.

extensions/eclipse-che-theia-testing-service-ext/src/browser/languages-test-main.ts
```diff
@@ -0,0 +1,170 @@
...
+    'codeActions' |
+    'codeLenses' |
+    'references' |
+    'symbols' |
+    'documentColors' |
+    'foldingRange' |
+    'renameEdits';
+
+@injectable()
+export class LanguagesMainTestImpl extends LanguagesMainImpl {
```
tsmaeder: There really is no reason to extend LanguagesMainImpl. Implementation inheritance is that big stick of OO and should only be used if appropriate (https://en.wikipedia.org/wiki/Fragile_base_class). You can just implement LanguagesMain and delegate to a wrapped instance of LanguagesMainImpl, no?

extensions/eclipse-che-theia-testing-service-ext/src/browser/languages-test-main.ts
```diff
@@ -0,0 +1,170 @@
...
+        this.registerPluginWithFeatureHandle(handle, pluginInfo.id, 'documentColors');
+        super.$registerDocumentColorProvider(handle, pluginInfo, selector);
+    }
+
+    $registerRenameProvider(handle: number, pluginInfo: PluginInfo, selector: SerializedDocumentFilter[], supportsResolveLocation: boolean): void {
+        this.registerPluginWithFeatureHandle(handle, pluginInfo.id, 'renameEdits');
+        super.$registerRenameProvider(handle, pluginInfo, selector, supportsResolveLocation);
+    }
+
+    private registerPluginWithFeatureHandle(handle: number, extensionId: string, newlyRegisteredAction: string): void {
```
tsmaeder: I would still move this method to the pluginHandleRegistry: it uses implementation details of the registry unnecessarily.

extensions/eclipse-che-theia-testing-service-ext/src/browser/plugin-handle-registry.ts
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * This class keeps a registry of which plugins map to which handle
+ */
+
+export interface LanguageFeatureRegistration {
+    languagesExtImpl: LanguagesExt; // The languagesMain that registered this plugin in the registry
+    providerHandles: Map<LanguageServerAction, number>; // A map of language actions to their handle
+}
+
+export interface LanguagesMainHandle {
```
tsmaeder: Isnt' it more "LanguagesExtHandle"?

extensions/eclipse-che-theia-testing-service-ext/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@eclipse-che/testing-service-ext",
+  "version": "0.0.1",
+  "description": "Theia - TestService Extension",
+  "dependencies": {
+    "@theia/plugin": "^0.13.0",
```
benoitf:
```suggestion
    "@theia/plugin": "next",
```

extensions/eclipse-che-theia-testing-service-ext/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "@theia/plugin-ext": "^0.13.0",
```
benoitf:
```suggestion
    "@theia/plugin-ext": "next",
```

extensions/eclipse-che-theia-testing-service-ext/package.json
```diff
@@ -0,0 +1,47 @@
...
+  },
+  "devDependencies": {
+    "clean-webpack-plugin": "^0.1.19",
+    "ts-loader": "^4.1.0",
+    "webpack": "^4.20.2",
+    "webpack-cli": "^3.1.1",
+    "typescript-formatter": "7.2.2",
+    "@types/js-yaml": "3.11.2"
+  }
+}
```
benoitf: missing EOF

extensions/eclipse-che-theia-testing-service-ext/src/browser/languages-test-api.ts
```diff
@@ -0,0 +1,252 @@
...
+import { CancellationToken, FoldingContext } from '@theia/plugin';
+import { SymbolInformation } from 'vscode-languageserver-types';
+import {
+    Position,
+    Selection,
+    RawColorInfo,
+    WorkspaceEditDto
+} from '@theia/plugin-ext/lib/common/plugin-api-rpc';
+
+export class TestAPIImpl implements TestAPI {
```
benoitf: maybe some comments ?

extensions/eclipse-che-theia-testing-service-ext/src/browser/languages-test-api.ts
```diff
@@ -0,0 +1,252 @@
...
+    private readonly pluginHandleRegistry: PluginHandleRegistry;
+
+    constructor(container: interfaces.Container) {
+        this.pluginHandleRegistry = container.get(PluginHandleRegistry);
+    }
+
+    $provideCompletionItems(pluginID: string, resource: UriComponents, position: Position,
+        context: CompletionContext, token: CancellationToken): Promise<CompletionResultDto | undefined> {
+
+        return this.pluginHandleRegistry.lookupLanguagesExtForPluginAndAction(pluginID, 'completion').then(potentialLanguagesExt => {
```
benoitf: use of await form ?
```
const potentialLanguagesExt = await this.pluginHandleRegistry.lookupLanguagesExtForPluginAndAction(pluginID, 'completion')
if (potentialLanguagesExt) {
```

extensions/eclipse-che-theia-testing-service-ext/src/browser/languages-test-api.ts
```diff
@@ -0,0 +1,252 @@
...
+            if (potentialLanguagesExt) {
+                return potentialLanguagesExt.languagesExt.$provideCompletionItems(potentialLanguagesExt.handle, resource, position, context, token);
```
benoitf: I don't see why we give the parameter `potentialLanguagesExt.handle` as we're calling potentialLanguagesExt instance so it should be already contained in this object no ?

extensions/eclipse-che-theia-testing-service-ext/src/browser/languages-test-api.ts
```diff
@@ -0,0 +1,252 @@
...
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
```
benoitf: as Thomas has already said, SPDX should be EPL-2.0

```
SPDX-License-Identifier: EPL-2.0
```

extensions/eclipse-che-theia-testing-service-ext/src/browser/languages-test-main.ts
benoitf: yes agree, it looks quite hard to maintain.
Also it looks like a proxy object

extensions/eclipse-che-theia-testing-service-ext/src/browser/plugin-handle-registry.ts
```diff
@@ -0,0 +1,76 @@
...
+    handle: number;
+    languagesExt: LanguagesExt;
+}
+
+@injectable()
+export class PluginHandleRegistry {
+    pluginRegistrationMap: Map<string, LanguageFeatureRegistration> = new Map();
+
+    private sleep(ms: number) {
```
benoitf: Shouldn't we resolve a promise when language feature is there instead of having waiting loops ?

code is doing: 
wait for 60s with small loops of 100ms to get something and use that object

while maybe it should do:
call something after language is ready.(callback when language is ready) (no wait loop)

extensions/eclipse-che-theia-testing-service-ext/src/node/testservice-api-node-provider.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+};
+
+function overrideInternalLoad(): void {
+    const module = require('module');
+    // save original load method
+    const internalLoad = module._load;
+
+    // if we try to resolve theia module, return the filename entry to use cache.
```
benoitf: copy/paste of theia but here it's `@eclipse-che/testing-service`

extensions/eclipse-che-theia-testing-service/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "prepare": "yarn run clean && yarn run build",
+    "clean": "rimraf lib",
+    "format": "tsfmt -r --useTsfmt ../../configs/tsfmt.json",
+    "lint": "tslint -c ../../configs/tslint.json --project tsconfig.json",
+    "compile": "tsc",
+    "build": "concurrently -n \"format,lint,compile\" -c \"red,green,blue\" \"yarn format\" \"yarn lint\" \"yarn compile\"",
+    "watch": "tsc -w"
+  },
+  "dependencies": {
+    "@theia/plugin": "^0.13.0",
```
benoitf:
```suggestion
    "@theia/plugin": "next",
```

extensions/eclipse-che-theia-testing-service/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "@theia/plugin-ext": "^0.13.0"
```
benoitf:
```suggestion
    "@theia/plugin-ext": "next"
```

extensions/eclipse-che-theia-testing-service/src/testing-service.d.ts
```diff
@@ -0,0 +1,97 @@
+/********************************************************************************
+ * Copyright (C) 2018 Red Hat, Inc. and others.
```
benoitf:
```suggestion
 * Copyright (C) 2020 Red Hat, Inc.
```

extensions/eclipse-che-theia-testing-service-ext/src/browser/languages-test-api.ts
```diff
@@ -0,0 +1,252 @@
+/********************************************************************************
+ * Copyright (C) 2020 Red Hat, Inc. and others.
```
benoitf:
```suggestion
 * Copyright (C) 2020 Red Hat, Inc.
```

extensions/eclipse-che-theia-testing-service-ext/src/node/testservice-ext-backend-module.ts
```diff
@@ -0,0 +1,24 @@
+/********************************************************************************
+ * Copyright (C) 2020 Red Hat, Inc. and others.
```
benoitf:
```suggestion
 * Copyright (C) 2020 Red Hat, Inc.
```

extensions/eclipse-che-theia-testing-service-ext/src/plugin/testservice-api.ts
```diff
@@ -0,0 +1,154 @@
+/********************************************************************************
+ * Copyright (C) 2020 Red Hat, Inc. and others.
```
benoitf:
```suggestion
 * Copyright (C) 2020 Red Hat, Inc.
```

extensions/eclipse-che-theia-testing-service-ext/src/testservice-plugin-api-provider.ts
```diff
@@ -0,0 +1,34 @@
+/********************************************************************************
+ * Copyright (C) 2020 Red Hat, Inc. and others.
```
benoitf:
```suggestion
 * Copyright (C) 2020 Red Hat, Inc.
```

JPinkney:
@benoitf @tsmaeder Do you mind taking a look again ?

JPinkney:
@benoitf I can move it to under the `@eclipse-che/plugin` in a test namespace. Should I also move everything in testing-service-ext to under plugin-ext?

azatsarynnyy:
I'm +1 for keeping everything under a single `che` namespace.

tsmaeder:
> 
> I was wondering why we need another root namespace for the test-service ? `'@eclipse-che/testing-service` and not included it in `@eclipse-che/plugin` under a test namespace

I think it makes perfect sense to have a separate API namespace: the che namespace essentially deals with communicating with the WSMaster API: workspace, devfiles, etc. The testing namespace has no substantial connection with that topic.
Secondly, by keeping it separate, the testing API could moved upstream eventually. Marrying it to Che workspaces would make that impossbile.
Thirdly, it could make sense not to ship the testing API in a production version of Che, but only in a development version. Think of `build.sh --production`.
So I'm all for keeping it separate.

benoitf:
sorry Thomas, but for all the reasons you're referring I'm even more for included in `che` namespace

che namespace is what you have when using che instead of theia, its not for only using workspace API. (it's just hat it's evolving on the needs)
If people write plugin's for che, they don't want to know if a separate API will be provided or not.
And if they need to find yet another namespace when writing code, it's even more complex.

> Secondly, by keeping it separate, the testing API could moved upstream eventually. Marrying it to Che workspaces would make that impossbile.
I don't see where it would make it impossible, you can link namespace. And what if people already starts to use `@eclipse-che/testing-service` it would break for them as well.

Also I'm talking about `che namespace`. Che namespace is what you get when using Theia in Che, it's not only `che workspace`

> Thirdly, it could make sense not to ship the testing API in a production version of Che, but only in a development version. Think of build.sh --production.

Why you would cut that ? If I use VS Code I can run tests on the 'production binary'. As a user I would want to test my plug-in on che.openshift.io for example.
Also people will never know if they could deploy or not the plugin in that case.

tsmaeder:
@benoitf the testing API is _not_ che-specific, there is really no reason to include it in the Che namespace other than pure convenience. If Che is mentioned elsewhere in the PR, we should probably change those occurrences. We even thought about doing all the works upstream, but decided on doing it downstream because it's simpler to proceed with the task.

tsmaeder:
> you can link namespace

Could you explain what you mean by this? I don't understand.

extensions/eclipse-che-theia-testing-service-ext/src/browser/languages-test-api.ts
```diff
@@ -0,0 +1,186 @@
...
+} from '@theia/plugin-ext/lib/common/plugin-api-rpc';
+
+/**
+ * This class redirects language api requests to the correct sidecars and returns the results
+ */
+export class TestAPIImpl implements TestAPI {
+
+    private readonly pluginHandleRegistry: PluginHandleRegistry;
+
+    constructor(container: interfaces.Container) {
```
tsmaeder: Is there a reason we're not using injection here?

extensions/eclipse-che-theia-testing-service-ext/src/browser/languages-test-api.ts
```diff
@@ -0,0 +1,186 @@
+/********************************************************************************
+ * Copyright (C) 2020 Red Hat, Inc.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
```
tsmaeder: Since we own the code 100%, we can make it straight EPL 2.0, no? Please search through the code and fix all locations.

extensions/eclipse-che-theia-testing-service-ext/src/browser/languages-test-api.ts
```diff
@@ -0,0 +1,186 @@
...
+
+    private readonly pluginHandleRegistry: PluginHandleRegistry;
+
+    constructor(container: interfaces.Container) {
+        this.pluginHandleRegistry = container.get(PluginHandleRegistry);
+    }
+
+    async $provideCompletionItems(pluginID: string, resource: UriComponents, position: Position,
+        context: CompletionContext, token: CancellationToken): Promise<CompletionResultDto | undefined> {
+        const { languagesExt, handle } = await this.pluginHandleRegistry.lookupLanguagesExtForPluginAndAction(pluginID, 'completion');
```
tsmaeder: What happens when lookup fails?

extensions/eclipse-che-theia-testing-service-ext/src/browser/plugin-handle-registry.ts
```diff
@@ -0,0 +1,98 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ ********************************************************************************/
+
+import { injectable } from 'inversify';
+import { LanguageServerAction } from './languages-test-main';
+import { LanguagesExt } from '@theia/plugin-ext';
+import { Emitter } from '@theia/core/lib/common/event';
+
+/**
+ * This class keeps a registry of which plugins map to which handle
```
tsmaeder: If you mean "LanguageFeatureRegistration", it's an interface, not a class.

extensions/eclipse-che-theia-testing-service-ext/src/browser/plugin-handle-registry.ts
```diff
@@ -0,0 +1,98 @@
...
+    handle: number;
+    languagesExt: LanguagesExt;
+}
+
+export interface LanguageRegistrationEvent extends LanguagesExtHandle {
+    pluginID: string;
+    action: LanguageServerAction;
+}
+
+// tslint:disable
```
tsmaeder: Didn't we move to eslint recently? Also, how far down the file does that disablement reach?

extensions/eclipse-che-theia-testing-service-ext/src/browser/plugin-handle-registry.ts
```diff
@@ -0,0 +1,98 @@
...
+    registerPluginWithFeatureHandle(handle: number, extensionId: string, newlyRegisteredAction: string, languagesExtProxy: LanguagesExt): void {
+        let potentialRegistration = this.pluginRegistrationMap.get(extensionId);
+        if (!potentialRegistration) {
+            potentialRegistration = {
+                providerHandles: new Map(),
+                languagesExtImpl: languagesExtProxy
+            };
+        }
+        potentialRegistration.providerHandles.set(newlyRegisteredAction, handle);
+        this.pluginRegistrationMap.set(extensionId, potentialRegistration);
```
tsmaeder: I think you can move this up into the `if` block.

extensions/eclipse-che-theia-testing-service-ext/src/browser/languages-test-api.ts
JPinkney: The lookup technically can't fail anymore. The promise will only get resolved when it finds the registered feature. If the feature isn't registered in time then the test will just timeout

extensions/eclipse-che-theia-testing-service-ext/src/browser/plugin-handle-registry.ts
JPinkney: I think only theia uses eslint right now. I think che-theia still uses tslint if I'm not mistaken

extensions/eclipse-che-theia-testing-service-ext/src/browser/languages-test-api.ts
JPinkney: It's just because of how the class is created https://github.com/eclipse/che-theia/pull/597/files#diff-0d93537fbf286e960396f1ccef10a48aR21 and following how its been done in other instances of che-theia

tsmaeder:
lgtm!

extensions/eclipse-che-theia-testing-service/src/testing-service.d.ts
```diff
@@ -0,0 +1,97 @@
...
+import {
+    Position,
+    Selection,
+    RawColorInfo,
+    WorkspaceEditDto
+} from '@theia/plugin-ext/lib/common/plugin-api-rpc';
+
+declare module '@eclipse-che/testing-service' {
+
+    export namespace languageserver {
```
benoitf: All these methods should be well documented, this is what the user will see when using code completion on the namespace

extensions/eclipse-che-theia-testing-service/src/testing-service.d.ts
```diff
@@ -0,0 +1,97 @@
...
+            pluginID: string,
+            resource: UriComponents,
+            context: FoldingContext,
+            token: CancellationToken
+        ): PromiseLike<FoldingRange[] | undefined>;
+        export function documentColors(pluginID: string, resource: UriComponents, token: CancellationToken): PromiseLike<RawColorInfo[]>;
+        export function renameEdits(pluginID: string, resource: UriComponents, position: Position, newName: string, token: CancellationToken): PromiseLike<WorkspaceEditDto | undefined>;
+    }
+
+}
```
benoitf: Hi, are you following some rules to have sometimes `Promise` vs `PromiseLike` ?

tsmaeder: I this case, it should reflect what ever the "non-testing" version of the function uses.

benoitf: ok b/c I see in "non-testing"

```typescript
provideRenameEdits(document: TextDocument, position: Position, newName: string, token: CancellationToken): ProviderResult<WorkspaceEdit>

provideDocumentSymbols(document: TextDocument, token: CancellationToken): ProviderResult<SymbolInformation[] | DocumentSymbol[]>;
```

and `renameEdits` is using PromiseLike while `documentSymbols` is using Promise

extensions/eclipse-che-theia-testing-service/src/testing-service.d.ts
```diff
@@ -0,0 +1,97 @@
...
+        export function documentSymbols(pluginID: string, resource: UriComponents, token: CancellationToken): Promise<DocumentSymbol[] | undefined>;
+        export function workspaceSymbols(pluginID: string, query: string, token: CancellationToken): PromiseLike<SymbolInformation[]>;
+        export function foldingRange(
+            pluginID: string,
+            resource: UriComponents,
+            context: FoldingContext,
+            token: CancellationToken
+        ): PromiseLike<FoldingRange[] | undefined>;
+        export function documentColors(pluginID: string, resource: UriComponents, token: CancellationToken): PromiseLike<RawColorInfo[]>;
+        export function renameEdits(pluginID: string, resource: UriComponents, position: Position, newName: string, token: CancellationToken): PromiseLike<WorkspaceEditDto | undefined>;
```
benoitf: Why do we have `WorkspaceEditDto` and not `WorkspaceEdit` ?

Lots of types should come from `@theia/plugin` no ?

extensions/eclipse-che-theia-testing-service/src/testing-service.d.ts
```diff
@@ -0,0 +1,97 @@
...
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import * as vst from 'vscode-languageserver-types';
```
benoitf: where is used that import ?

tsmaeder:
> 
> 
> please merge test service into `@eclipse-che/plugin` into `test` namespace or `language.test`

@JPinkney and I discussed this in yesterdays daily and we both agree that the testing stuff should be a separate module and API namespace if that's what you object to (for the reasons outlined before).

extensions/eclipse-che-theia-testing-service/src/testing-service.d.ts
```diff
@@ -0,0 +1,97 @@
...
+    ReferenceContext,
+    Location,
+    SignatureHelpContext,
+    CodeActionContext,
+    CodeAction,
+    FoldingRange,
+} from '@theia/plugin-ext/lib/common/plugin-api-rpc-model';
+import { UriComponents } from '@theia/plugin-ext/lib/common/uri-components';
+import { CancellationToken, FoldingContext } from '@theia/plugin';
+import { SymbolInformation } from 'vscode-languageserver-types';
```
benoitf: why are we not using https://github.com/eclipse-theia/theia/blob/master/packages/plugin/src/theia.d.ts#L5448  (I would say all types should come from `@theia/plugin` no) ?

extensions/eclipse-che-theia-testing-service/src/testing-service.d.ts
JPinkney: https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/plugin/languages.ts#L543 it's just following what provideRenameEdits uses

benoitf: IMHO it should follow the declaration, not the implementation

extensions/eclipse-che-theia-testing-service/src/testing-service.d.ts
JPinkney: Types are different because when I was creating the d.ts file I used the types that came out of https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/common/plugin-api-rpc.ts#L1146 not the types that came out of the theia.d.ts

extensions/eclipse-che-theia-testing-service/src/testing-service.d.ts
JPinkney: It's following the LanguageExt interface: https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/common/plugin-api-rpc.ts#L1196

extensions/eclipse-che-theia-testing-service/src/testing-service.d.ts
benoitf: yes I see that but we should map to external/public stuff, not implementation, no ?

JPinkney:
@benoitf I've tried switching the types to be compatible with @theia/plugin but I'm not entirely sure if its possible

in che-languages-test-api when we call:
```typescript
(languagesExt as LanguagesExt).$provideCompletionItems(......)
```
we are recieving back
```typescript
Promise<CompletionResultDto | undefined>
```
from https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/plugin/languages/completion.ts#L39. We aren't actually calling the delegate that has the type theia.CompletionItemProvider https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/plugin/languages/completion.ts#L34.

Instead, we are getting back the converted result that happens after the delegate is called.

When I tried to modify che-languages-test-api to support @thiea/plugin types like
```typescript
$provideImplementation(pluginID: string, document: TextDocument, position: Position, token: CancellationToken): ProviderResult<Definition | DefinitionLink[]> {
        return this.pluginHandleRegistry.lookupLanguagesExtForPluginAndAction(pluginID, 'implementation').then(({ languagesExt, handle }) => {
            return (languagesExt as LanguagesExt).$provideImplementation(handle, document.uri, { column: position.character, lineNumber: position.line }, token);
        });
    }
```
it leads to a lot of incompatible types for all of the methods (provideCompletionItems, provideImplementation, etc etc) in that class.

I do have a working branch that has moved the testing api into @eclipse-che/api though.

tsmaeder:
> I do have a working branch that has moved the testing api into @eclipse-che/api though.

Let's please not do that. It's wrong! Before we do that, let's discuss it in a call.

ericwill:
Superseded by #802

PR #598, opened 2020-01-11
Use theia-plugin instead of theia:plugin.

monaka:
### What does this PR do?

Uses `theia-plugin` instead of `theia:plugin` in each `compile` phase.

Will fix eclipse/che#14145 eclipse/che#15552.


mmorhun:
I think all docs should be changed as well.

monaka:
@mmorhun I roughly grepped in `che-docs` and `che-theia`  with `theia:plugin`. But I couldn't find. Which docs should we fix?

PR #602, opened 2020-01-16
Track 'git commit' and 'git push' commands and push corresponded telemetry events

vparfonov:
Signed-off-by: Vitalii Parfonov <vparfono@redhat.com>

### What does this PR do?
Track 'git commit' and 'git push' commands and push corresponded telemetry events

eclipse/che#15622

### DIFF
--- a/extensions/eclipse-che-theia-plugin-remote/src/node/execute-command-container-aware.ts
+++ b/extensions/eclipse-che-theia-plugin-remote/src/node/execute-command-container-aware.ts
@@ -8,18 +8,18 @@
  * SPDX-License-Identifier: EPL-2.0
  **********************************************************************/
 import { CommandRegistryImpl } from '@theia/plugin-ext/lib/plugin/command-registry';
-// import { overrideUri } from './che-content-aware-utils';
+import { overrideUri } from './che-content-aware-utils';
 
 export class ExecuteCommandContainerAware {
     static makeExecuteCommandContainerAware(commandRegistryExt: CommandRegistryImpl) {
-        // const executeCommandContainerAware = new ExecuteCommandContainerAware();
-        // executeCommandContainerAware.overrideExecuteCommand(commandRegistryExt);
+        const executeCommandContainerAware = new ExecuteCommandContainerAware();
+        executeCommandContainerAware.overrideExecuteCommand(commandRegistryExt);
     }
 
     overrideExecuteCommand(commandRegistryExt: CommandRegistryImpl) {
-        // const originalExecuteCommand = commandRegistryExt.executeCommand.bind(commandRegistryExt);
+        const originalExecuteCommand = commandRegistryExt.executeCommand.bind(commandRegistryExt);
         // tslint:disable-next-line:no-any
-        // const executeCommand = (id: string, ...args: any[]) => originalExecuteCommand(id, ...args.map(arg => arg.scheme ? overrideUri(arg) : arg));
-        // commandRegistryExt.executeCommand = executeCommand;
+        const executeCommand = (id: string, ...args: any[]) => originalExecuteCommand(id, ...args.map(arg => arg.scheme ? overrideUri(arg) : arg));
+        commandRegistryExt.executeCommand = executeCommand;
     }
 }
--- a/extensions/eclipse-che-theia-plugin-remote/src/node/execute-command-container-aware.ts
+++ b/extensions/eclipse-che-theia-plugin-remote/src/node/execute-command-container-aware.ts
@@ -8,18 +8,18 @@
  * SPDX-License-Identifier: EPL-2.0
  **********************************************************************/
 import { CommandRegistryImpl } from '@theia/plugin-ext/lib/plugin/command-registry';
-import { overrideUri } from './che-content-aware-utils';
+// import { overrideUri } from './che-content-aware-utils';
 
 export class ExecuteCommandContainerAware {
     static makeExecuteCommandContainerAware(commandRegistryExt: CommandRegistryImpl) {
-        const executeCommandContainerAware = new ExecuteCommandContainerAware();
-        executeCommandContainerAware.overrideExecuteCommand(commandRegistryExt);
+        // const executeCommandContainerAware = new ExecuteCommandContainerAware();
+        // executeCommandContainerAware.overrideExecuteCommand(commandRegistryExt);
     }
 
     overrideExecuteCommand(commandRegistryExt: CommandRegistryImpl) {
-        const originalExecuteCommand = commandRegistryExt.executeCommand.bind(commandRegistryExt);
+        // const originalExecuteCommand = commandRegistryExt.executeCommand.bind(commandRegistryExt);
         // tslint:disable-next-line:no-any
-        const executeCommand = (id: string, ...args: any[]) => originalExecuteCommand(id, ...args.map(arg => arg.scheme ? overrideUri(arg) : arg));
-        commandRegistryExt.executeCommand = executeCommand;
+        // const executeCommand = (id: string, ...args: any[]) => originalExecuteCommand(id, ...args.map(arg => arg.scheme ? overrideUri(arg) : arg));
+        // commandRegistryExt.executeCommand = executeCommand;
     }
 }
--- a/plugins/telemetry-plugin/src/telemetry-plugin.ts
+++ b/plugins/telemetry-plugin/src/telemetry-plugin.ts
@@ -22,6 +22,10 @@ export function start(context: theia.PluginContext) {
             ]);
     });
 
+    che.telemetry.onWillExecuteCommand(async event => {
+        console.log(' ===========================> ' + event.commandId);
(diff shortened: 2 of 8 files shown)


plugins/telemetry-plugin/src/telemetry-plugin.ts
```diff
@@ -21,6 +21,51 @@ export function start(context: theia.PluginContext) {
                 ['programming language', e.document.languageId]
             ]);
     });
+
+    let gitLogHandlerInitialized: boolean;
```
benoitf: hello, question: can't we hook on global command invocation instead of intercept the plugin's code ?

so we could just add `git.commit` and `git.push` as commands to intercept/send telemetry events ?

plugins/telemetry-plugin/src/telemetry-plugin.ts
```diff
@@ -21,6 +21,51 @@ export function start(context: theia.PluginContext) {
...
+            let url: string;
+            const listener = async (out: string) => {
+                // Parse Git log events.
+                const split = out.split(' ');
+                if (out.startsWith('> git commit') || out.startsWith('> git push')) {
+                    command = split[2];
+                    url = split[3];
+                    switch (command) {
+                        case 'commit': {
+                            che.telemetry.event('GIT_USED', context.extensionPath,
```
davidfestal: According to [this code in the existing telemetry backend base](https://github.com/che-incubator/che-workspace-telemetry-client/blob/master/backend-base/src/main/java/org/eclipse/che/incubator/workspace/telemetry/base/AnalyticsEvent.java#L24), the name of the event would be: `COMMIT_LOCALLY`

Additionally, for now we don't have any parameter related to the commit event (no URL).

It could be added in the future (if PM agrees), but then it should be added in the [`AnalyticsEvents`](https://github.com/che-incubator/che-workspace-telemetry-client/blob/master/backend-base/src/main/java/org/eclipse/che/incubator/workspace/telemetry/base/AnalyticsEvent.java) and [`EventProperties`](https://github.com/che-incubator/che-workspace-telemetry-client/blob/master/backend-base/src/main/java/org/eclipse/che/incubator/workspace/telemetry/base/EventProperties.java) files

plugins/telemetry-plugin/src/telemetry-plugin.ts
```diff
@@ -21,6 +21,51 @@ export function start(context: theia.PluginContext) {
...
+                            [
+                                ['commit command', url]
+                            ]);
+                            console.log(`Sent telemetry event: git commit : project ${url}`);
+                            break;
+                        }
+                        case 'push': {
+                            che.telemetry.event('GIT_USED', context.extensionPath,
```
davidfestal: According to [this code in the existing telemetry backend base](https://github.com/che-incubator/che-workspace-telemetry-client/blob/master/backend-base/src/main/java/org/eclipse/che/incubator/workspace/telemetry/base/AnalyticsEvent.java#L23), the name of the event would be: `PUSH_TO_REMOTE`

Additionally, for now we don't have any parameter related to the commit event (no URL).

It could be added in the future (if PM agrees), but then it should be added in the [`AnalyticsEvents`](https://github.com/che-incubator/che-workspace-telemetry-client/blob/master/backend-base/src/main/java/org/eclipse/che/incubator/workspace/telemetry/base/AnalyticsEvent.java) and [`EventProperties`](https://github.com/che-incubator/che-workspace-telemetry-client/blob/master/backend-base/src/main/java/org/eclipse/che/incubator/workspace/telemetry/base/EventProperties.java) files

plugins/telemetry-plugin/src/telemetry-plugin.ts
vparfonov: I think it will be possible if we extend in some way Theia CommandRegistry, but not sure is it make sense.

benoitf: I think it would allow to plug more easily other new events (or to track for example all commands the user is invoking)
besides that, it'll be more robust than going inside internal structure/objects of some plug-ins.


It would be like just providing the name of the command to intercept and the value to insert in telemetry

```typescript
something.intercept('git.commit', 'COMMIT_LOCALLY');
something.intercept('git.push', 'PUSH_TO_REMOTE');
```

davidfestal: @benoitf Do you think it could be kept for a next PR ?

benoitf: to me, it's looking as too fragile right now.

plugins/telemetry-plugin/src/telemetry-plugin.ts
vparfonov: done

plugins/telemetry-plugin/src/telemetry-plugin.ts
vparfonov: in case 
`something.intercept('git.commit', 'COMMIT_LOCALLY');` how we will send parameters if needed?

benoitf: I noticed that there were no parameters right now, but you can opt for another way like

format is open :-)
```
something.intercep('git.commit', (callback, args, etc) => callback.send('COMMIT_LOCALLY', args[2].substring(123).toUpperCase().whatever()) )
```

vparfonov: @benoitf  Other possible solution 
https://github.com/eclipse-theia/theia/commit/b39a852192256e333ed6f4e0a13bd48d127123ae#diff-ab6e2d4c57658cf60e550cfe9ba87eacR295
WDYT?

benoitf: @vparfonov I'm not sure to land on the correct link ? it's about activation events ?

benoitf: but event contains only the commandID, correct, not the args you were asking for ?

vparfonov: Yep probably this link better https://github.com/eclipse-theia/theia/blob/master/packages/core/src/common/command.ts#L273 
We can add parameters to the event.
Because `intercept` in this case something new for me I don't know how it works

benoitf: ok, for me, intercept was a method you were implementing

basically, it waits that command is executed (for example tracking all events you mentioned) and then it send the telemetry event with associated name if there is a matching reference (git.commit --> COMMIT_LOCALLY)

davidfestal: @benoitf Do you have an estimate of how long it would be to do the change you're discussing with @vparfonov ? If it's not more than a few hours, let's do the change in this PR and avoid adding some more technical debt.

benoitf: I don't think it's long task as the event is there (Vitalii showed a link) so it's just handle the command event callback and if it's matching command, send the expected telemetry event

RomanNikitenko: If I understand correctly, the event is on `theia` side and there is no API to track the event on `plugin` side. So we need to provide this ability first.
please correct me if I'm wrong.

benoitf: on plugin side we have `che.telemetry` object and implementation of that is not on plug-in side but on `ext` so it can hook on events.

this is why I proposed
```
che.telemetry.addIntercept('git.commit', 'COMMIT_LOCALLY')
```
or
```
che.telemetry.addIntercept('git.commit', (command) => {che.telemetry.event('COMMIT_LOCALLY'....)
```

davidfestal: @vparfonov according to the last answers @benoitf provided, it seems to me that we should probably do that in the current PR.
cc @ibuziuk

extensions/eclipse-che-theia-plugin-remote/src/node/execute-command-container-aware.ts
```diff
@@ -8,18 +8,18 @@
  * SPDX-License-Identifier: EPL-2.0
  **********************************************************************/
 import { CommandRegistryImpl } from '@theia/plugin-ext/lib/plugin/command-registry';
-import { overrideUri } from './che-content-aware-utils';
+// import { overrideUri } from './che-content-aware-utils';
```
ibuziuk: @vparfonov a lot of commented lines in the file. Do we need them?

vparfonov: just testing, will remove

vparfonov:
crw-ci-test

dmytro-ndp:
crw-ci-test

vparfonov:
special thanks to @RomanNikitenko for help

PR #604, opened 2020-01-17
Apply GitHub authentication plugin

vinokurig:
### What does this PR do?
Add a plugin that authenticates [vscode GitHub PR plugin](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github) launched in Che-Theia
ezgif-2-e8952916c88e

The plugin calls Che oAuth API service to get the GitHub token. Then it injects it to the user preferences. When the browser page is refreshed the vscode GitHub PR plugin fetches the token and cleans the token from the preferences file.

This works only in single user Che. When https://github.com/eclipse/che/issues/15672 is applied the token could be fetched in the multi-user.

**What can be improved:**
The login actions that provided by the native vscode GitHub plugin are still redirect to the vscode authentication service. To override them we need:
1.  Apply `Urihandler` plugin API method: https://github.com/eclipse-theia/theia/issues/5119. It is used by the vscode GitHub PR plugin to handle the callback request from authentication server.
2. Authentication server url is hard-coded in the vscode GitHub PR plugin: https://github.com/microsoft/vscode-pull-request-github/blob/ddaa2bcd06ebc931e075cc9cab5a83f683e93e83/src/authentication/githubServer.ts#L16-L17
https://github.com/microsoft/vscode-pull-request-github/blob/ddaa2bcd06ebc931e075cc9cab5a83f683e93e83/src/authentication/githubServer.ts#L177-L179
so we need to propose a PR to the [vscode GitHub PR plugin](https://github.com/Microsoft/vscode-pull-request-github) repository that allows to override the authentication urls.

When the steps above are done we can setup the vscode GitHub PR plugin to use Che oAuth service for authentication and get rid of this authentication plugin.

https://github.com/eclipse/che/issues/14217

#### Docs PR

https://github.com/eclipse/che-docs/pull/1055

### DIFF
--- a/plugins/github-auth-plugin/README.md
+++ b/plugins/github-auth-plugin/README.md
@@ -1,0 +1,5 @@
+# Theia - Telemetry Plug-in
+
+## License
+
+[EPL-2.0](http://www.eclipse.org/legal/epl-2.0)
--- a/plugins/github-auth-plugin/README.md
+++ b/plugins/github-auth-plugin/README.md
@@ -1,4 +1,8 @@
-# Theia - Telemetry Plug-in
+# Theia - vscode Github pull-request plugin authenticator
+
+The plugin calls Che oAuth API service to get the GitHub token.
+Then it injects it to the user preferences. When the browser page is refreshed,
+the vscode GitHub PR plugin fetches the token and cleans the token from the preferences file.
 
 ## License
 
--- a/plugins/github-auth-plugin/package.json
+++ b/plugins/github-auth-plugin/package.json
@@ -5,7 +5,7 @@
       "keywords": [
         "theia-plugin"
       ],
-      "description": "Manage Eclipse Che Telemetry Plugin",
+      "description": "Authenticates the vscode Github pull-request plugin",
       "license": "EPL-2.0",
 
       "files": [
--- a/plugins/github-auth-plugin/package.json
+++ b/plugins/github-auth-plugin/package.json
@@ -1,0 +1,42 @@
+{
+      "name": "@eclipse-che/github-auth-plugin",
+      "version": "0.0.1",
+      "publisher": "Eclipse Che",
+      "keywords": [
+        "theia-plugin"
+      ],
+      "description": "Manage Eclipse Che Telemetry Plugin",
+      "license": "EPL-2.0",
+
+      "files": [
+        "src"
+      ],
+      "activationEvents": [
+        "*"
+      ],
+      "devDependencies": {
+        "@theia/plugin": "next",
+        "@theia/plugin-packager": "latest",
+        "@eclipse-che/plugin": "0.0.1",
+        "rimraf": "2.6.2",
+        "typescript-formatter": "7.2.2",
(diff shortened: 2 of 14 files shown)


plugins/github-auth-plugin/README.md
```diff
@@ -0,0 +1,5 @@
+# Theia - Telemetry Plug-in
```
RomanNikitenko: something wrong here :-)

vinokurig: Thanks, fixed the description :)

plugins/github-auth-plugin/src/github-auth-plugin.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ **********************************************************************/
+
+import * as theia from '@theia/plugin';
+import * as che from '@eclipse-che/plugin';
+
+export function start(context: theia.PluginContext) {
+    if (theia.plugins.getPlugin('github.vscode-pull-request-github')) {
+        const informationMessageTestCommand = {
```
benoitf: looks the variable name is not self explanatory

benoitf:
hello, why do we need to refresh the browser ? (on UX it may not be so nice to do that)

plugins/github-auth-plugin/src/github-auth-plugin.ts
vinokurig: fixed

vinokurig:
@benoitf I agree that it is not a good UX, but the vscode GitHub PR plugin loads the token from configuration on it's initialisation https://github.com/microsoft/vscode-pull-request-github/blob/6b019aa02b90cc5729f14ba8b98f653b34fa64a6/src/extension.ts#L37

plugins/github-auth-plugin/package.json
```diff
@@ -0,0 +1,42 @@
+{
+      "name": "@eclipse-che/github-auth-plugin",
+      "version": "0.0.1",
+      "publisher": "Eclipse Che",
+      "keywords": [
+        "theia-plugin"
+      ],
+      "description": "Manage Eclipse Che Telemetry Plugin",
```
RomanNikitenko: please fix the description

plugins/github-auth-plugin/src/github-auth-plugin.ts
```diff
@@ -0,0 +1,35 @@
+/*********************************************************************
+ * Copyright (c) 2019 Red Hat, Inc.
```
RomanNikitenko: 2020?

plugins/github-auth-plugin/package.json
vinokurig: done

plugins/github-auth-plugin/src/github-auth-plugin.ts
vinokurig: done

vinokurig:
crw-ci-test

PR #605, opened 2020-01-23
Serialize webview in welcome plugin to be able to restore it after page reload

vzhukovskii:
### What does this PR do?
This changes proposal add several fixes to the welcome page plugin. The main problem was related to the restoring opened editors after page reloading. Welcome page represents as a _WebView_ panel, that should be restored after plugin starting phase. But at workspace opening phase the plugin tried to open _WebView_ that is not consistent with the main bootstrap flow. Here is some fixes that performed in this PR:

- Add ability to serialize welcome page _WebView_ to be able to restore it.

- Reduced delay **from 1 to 0.1 second** when welcome page should be opened after workspace is started.

- Changed mechanism of restoring welcome page. **Now it opens in case when there isn't any opened other editors and there is enabled option in Theia Preferences.** 

**_Note that application state is storing after 10 second when any operation with any widget was performed._**

https://github.com/eclipse/che/issues/15317

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/extensions/eclipse-che-theia-user-preferences/src/browser/che-storage-service.ts
+++ b/extensions/eclipse-che-theia-user-preferences/src/browser/che-storage-service.ts
@@ -7,10 +7,11 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  **********************************************************************/
-import { injectable, inject } from 'inversify';
+import { injectable, inject, postConstruct } from 'inversify';
 import { StorageService } from '@theia/core/lib/browser';
 import { StorageServer } from '../common/storage-server';
 import { ILogger } from '@theia/core/lib/common/logger';
+import { WorkspaceService } from '@theia/workspace/lib/browser/workspace-service';
 
 @injectable()
 export class CheStorageService implements StorageService {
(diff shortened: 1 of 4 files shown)


extensions/eclipse-che-theia-user-preferences/src/browser/che-storage-service.ts
```diff
@@ -7,25 +7,38 @@
...
-    setData<T>(key: string, data?: T): Promise<void> {
+    private initialized: Promise<void>;
+
+    @postConstruct()
+    protected init(): void {
+        this.initialized = this.workspaceService.roots.then(() => {});
+    }
+
+    async setData<T>(key: string, data?: T): Promise<void> {
+        await this.initialized;
```
vitaliy-guliy: Is it posible to call `await this.workspaceService.roots()` directly?

extensions/eclipse-che-theia-user-preferences/src/browser/che-storage-service.ts
```diff
@@ -7,25 +7,38 @@
...
         return this.storageServer.setData(key, JSON.stringify(data));
     }
 
     async getData<T>(key: string, defaultValue?: T): Promise<T | undefined> {
+        await this.initialized;
```
vitaliy-guliy: The same as above.

extensions/eclipse-che-theia-user-preferences/src/browser/che-storage-service.ts
vzhukovskii: Yes, it is possible, but this variable was introduced in the same way how it done it https://github.com/eclipse-theia/theia/blob/master/packages/workspace/src/browser/workspace-storage-service.ts#L30-L52, which we rebind.

vitaliy-guliy: Just try, it should work.

PR #606, opened 2020-01-24
Apply terminal title name configuration

vinokurig:
### What does this PR do?
Add a configuration that controls default terminal title override:
ezgif-2-d5ad920320ee

### How to test?
Start a workspace from dev-file that contains a custom container that changes terminal title e.g: 
```
- alias: my-container
    type: dockerimage
    image: ivinokur/rhel8
    mountSources: true
    memoryLimit: 256M
```
fixes https://github.com/eclipse/che/issues/15242

### DIFF
--- a/che-theia-init-sources.yml
+++ b/che-theia-init-sources.yml
@@ -20,4 +20,5 @@ sources:
     - plugins/welcome-plugin
     - plugins/ssh-plugin
     - plugins/telemetry-plugin
+    - plugins/github-auth-plugin
   checkoutTo: master
--- a/cico_build_master.sh
+++ b/cico_build_master.sh
@@ -22,6 +22,9 @@ set -e
 
 parse "$@"
 
+export CDN_PREFIX=https://static.developers.redhat.com/che/theia_artifacts/
+export MONACO_CDN_PREFIX=https://cdn.jsdelivr.net/npm/
+
 install_deps
 load_jenkins_vars
 buildImages
--- a/cico_build_master.sh
+++ b/cico_build_master.sh
@@ -26,6 +26,8 @@ export CDN_PREFIX=https://static.developers.redhat.com/che/theia_artifacts/
 export MONACO_CDN_PREFIX=https://cdn.jsdelivr.net/npm/
 
 install_deps
+set +x
 load_jenkins_vars
+set -x
 buildImages
 publishImagesOnQuay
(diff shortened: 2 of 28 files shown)


vinokurig:
@azatsarynnyy The default terminal title name comes from the dev-file

benoitf:
Hello,

to me the property name is confusing. I expect that `override` is when container name is added (like the opposite)

BTW, as the naming of title/terminals is quite confusing in the Che context, shouldn't it be the default ?
By default, users opening terminals should see the container name as prefix without having to update some preferences.

benoitf:
@dmytro-ndp I tried with che-bot links to review but it seems it's not using the right images so I wasn't able to test by clicking.

```
  containers:
   - name: theia-ide
     image: "quay.io/eclipse/che-theia:next"
```

vinokurig:
@benoitf
> to me the property name is confusing. I expect that override is when container name is added (like the opposite)

How about `allowToChangeTitle` ?

benoitf:
I'm still wondering if we need to have a property or just prefix with container all the time ?

vinokurig:
I prefer to keep the property. Title with both container name and overridden title looks too long:
screenshot-che-che 192 168 99 118 nip io-2020 01 27-10_53_11
@slemeur @l0rd WDYT?

benoitf:
I think we could shorter the title
We already know we're inside a terminal so no need to have `terminal 0`, `terminal 1`
`container-short-name#1'`is enough + hover with full name (I'm unsure as well that first terminal should be 0 for humans)

vinokurig:
@benoitf I like your solution but we need to have agreement with @slemeur and @l0rd because the corresponding issue is about adding a property and viewing only default title or title from container: https://github.com/eclipse/che/issues/15242

l0rd:
@vinokurig I agree with @benoitf 
Moreover I am not sure if there is any value in having the property: it should always be overrideTitle=false.  As a user why should I be interested in setting `"terminal.overrideTitle": true`?  cc @AndrienkoAleksandr

AndrienkoAleksandr:
> Moreover I am not sure if there is any value in having the property: it should always be overrideTitle=false. As a user why should I be interested in setting "terminal.overrideTitle": true? cc @AndrienkoAleksandr

Show terminal title - it is a feature. I think most users don't need it. But I can not predict desire all users, that's why for me make sense to have an option.

vinokurig:
@benoitf @l0rd I have updated the PR to have both container name and container title in the terminal title:
screenshot-servere4jzwkrf-che-dev-server-3010 192 168 99 127 nip io-2020 01 30-10_48_07
Can we go with it?

PR #624, opened 2020-02-11
Implemented a possibility of obtaining client IP address for telemetry plugin

vparfonov:
Signed-off-by: Vitalii Parfonov <vparfono@redhat.com>

### What does this PR do?
Implement a possibility of obtaining client IP address for telemetry plugin

https://github.com/eclipse/che/issues/15624

### DIFF
--- a/extensions/eclipse-che-theia-plugin-ext/src/browser/che-telemetry-main.ts
+++ b/extensions/eclipse-che-theia-plugin-ext/src/browser/che-telemetry-main.ts
@@ -32,7 +32,8 @@ export class CheTelemetryMainImpl implements CheTelemetryMain {
 
     async $event(id: string, ownerId: string, properties: [string, string][]): Promise<void> {
         if (!this.ip) {
-            this.ip = (await this.getClientAddressInfo()).ip;
+            const client = await this.getClientAddressInfo();
+            this.ip = client.ip !== undefined ? client.ip : '';
         }
         let agent = '';
         let resolution = '';
@@ -53,7 +54,7 @@ export class CheTelemetryMainImpl implements CheTelemetryMain {
     }
 
     async $getClientAddressInfo(): Promise<ClientAddressInfo> {
-       return this.getClientAddressInfo();
+        return this.getClientAddressInfo();
     }
 
     async getClientAddressInfo(): Promise<ClientAddressInfo> {
@@ -61,6 +62,11 @@ export class CheTelemetryMainImpl implements CheTelemetryMain {
         if (response.status === 200) {
             return response.data;
         }
-        return {ip: 'not defined'};
+        console.log('Can`t obtain client adress information. Status: ' + response.status + 'Error message: ' + response.data);
+        return {
+            ip: undefined,
+            ipFamily: undefined,
+            port: undefined
+        };
     }
 }
--- a/extensions/eclipse-che-theia-plugin-ext/src/browser/che-telemetry-main.ts
+++ b/extensions/eclipse-che-theia-plugin-ext/src/browser/che-telemetry-main.ts
@@ -13,10 +13,12 @@ import { RPCProtocol } from '@theia/plugin-ext/lib/common/rpc-protocol';
 import { CheTelemetryMain, PLUGIN_RPC_CONTEXT, CheTelemetry } from '../common/che-protocol';
 import { CheApiService } from '../common/che-protocol';
 import { CommandRegistry } from '@theia/core';
+import * as axios from 'axios';
 
 export class CheTelemetryMainImpl implements CheTelemetryMain {
 
     private readonly cheApiService: CheApiService;
+    private ip: string;
 
     constructor(container: interfaces.Container, rpc: RPCProtocol) {
         const proxy: CheTelemetry = rpc.getProxy(PLUGIN_RPC_CONTEXT.CHE_TELEMETRY);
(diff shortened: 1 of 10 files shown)


extensions/eclipse-che-theia-plugin-ext/src/node/che-client-ip-service.ts
```diff
@@ -0,0 +1,27 @@
+/*********************************************************************
+ * Copyright (c) 2018 Red Hat, Inc.
```
RomanNikitenko: 2020?

extensions/eclipse-che-theia-plugin-ext/src/node/che-client-ip-service.ts
```diff
@@ -0,0 +1,27 @@
...
+import * as express from 'express';
+import { BackendApplicationContribution } from '@theia/core/lib/node/backend-application';
+import { injectable } from 'inversify';
+
+@injectable()
+export class CheClientIpServiceContribution implements BackendApplicationContribution {
+
+    configure(app: express.Application): void {
+        app.get('/che/client-ip', (req, res) => {
+            let clientIp = req.connection.remoteAddress;
```
ibuziuk: wondering will this work for both IPv4 & IPv6?

benoitf:
hello, should we move the getIp to be provided to all plug-ins and not used only internally by telemetry ?

extensions/eclipse-che-theia-plugin-ext/src/node/che-client-ip-service.ts
vparfonov: Yes, it will work correct for both https://nodejs.org/api/net.html#net_socket_remoteaddress

vparfonov:
> hello, should we move the getIp to be provided to all plug-ins and not used only internally by telemetry ?

@davidfestal WDYT about that?

davidfestal:
I assume I agree with @benoitf . Is it much work ?

vparfonov:
> as ip service is provided on top of `@eclipse-che/plugin` (with backend contribution) the IP visibility needs to be provided by `@eclipse-che/plugin` as well (so any Che plug-ins could use it)
@benoitf 
done, review again, please

extensions/eclipse-che-theia-plugin-ext/src/browser/che-telemetry-main.ts
```diff
@@ -51,4 +51,16 @@ export class CheTelemetryMainImpl implements CheTelemetryMain {
...
+    async $getClientAddressInfo(): Promise<ClientAddressInfo> {
+       return this.getClientAddressInfo();
+    }
+
+    async getClientAddressInfo(): Promise<ClientAddressInfo> {
+        const response = await axios.default.get('/che/client-ip');
+        if (response.status === 200) {
+            return response.data;
+        }
+        return {ip: 'not defined'};
```
RomanNikitenko: please fix formatting issue

extensions/eclipse-che-theia-plugin-ext/src/browser/che-telemetry-main.ts
```diff
@@ -51,4 +51,16 @@ export class CheTelemetryMainImpl implements CheTelemetryMain {
 
         return this.cheApiService.submitTelemetryEvent(id, ownerId, this.ip, agent, resolution, properties);
     }
+
+    async $getClientAddressInfo(): Promise<ClientAddressInfo> {
+       return this.getClientAddressInfo();
```
RomanNikitenko: please fix formatting issue

extensions/eclipse-che-theia-plugin/src/che-proposed.d.ts
```diff
@@ -70,10 +70,16 @@ declare module '@eclipse-che/plugin' {
...
+    export interface ClientAddressInfo {
+        ip: string,
+        port?: string
+        ipFamily?: string
+    }
     export namespace telemetry {
         export function event(id: string, ownerId: string, properties: [string, string][]): Promise<void>;
         /** Fires when a command will starts. */
         export function addCommandListener(commandId: string, listener: TelemetryListener): Promise<void>;
+        export function getClienAddressInfo(): Promise<ClientAddressInfo>;
```
RomanNikitenko: I think it makes sense to document that for some cases `'not defined` can be returned instead of actual `ip`

extensions/eclipse-che-theia-plugin-ext/src/node/che-client-ip-service.ts
```diff
@@ -0,0 +1,33 @@
...
+            if (!clientIp) {
+                clientIp = 'not defined';
```
benoitf: I would say `undefined` type rather than string then

extensions/eclipse-che-theia-plugin-ext/src/browser/che-telemetry-main.ts
```diff
@@ -46,6 +49,18 @@ export class CheTelemetryMainImpl implements CheTelemetryMain {
...
+    }
+
+    async getClientAddressInfo(): Promise<ClientAddressInfo> {
+        const response = await axios.default.get('/che/client-ip');
+        if (response.status === 200) {
+            return response.data;
+        }
+        return {ip: 'not defined'};
```
benoitf: it should be `undefined` type (rather than 'not defined' string)
Also might be interesting to log the error if we don't get correct status/error code (or include status in clientAddressInfo)

benoitf:
@vparfonov do you have a theia plug-in to test that get client IP ?

vparfonov:
crw-ci-test

PR #625, opened 2020-02-11
Add rsync tool to che-theia-dev container

vitaliy-guliy:

### What does this PR do?
Rsync tool is needed to synchronize sources between /projects and /tmp directories to be able to build them in /tmp directory on che.openshift.io


vitaliy-guliy:
> probably you'll need to enhance devfile as well with a new command ?

If you mean the devfile for che.openshift.io dogfooding, the command to rsync soures is already there.
For local che deployments we don't need any changes.

benoitf:
I would include devfile there https://github.com/eclipse/che-theia/tree/master/devfiles

vitaliy-guliy:
@benoitf ok, I will create another PR containing the devfile

PR #643, opened 2020-02-27
Remove deprecated Theia extensions. Use VS Code built-in ones.

azatsarynnyy:

### What does this PR do?
Removes the deprecated Theia extensions which will be deleted from upstream Theia.
Uses VS Code built-ins instead.

https://github.com/eclipse/che/issues/16133


azatsarynnyy:
crw-ci-test

azatsarynnyy:
crw-ci-test

generator/src/templates/assembly-package.mst
```diff
@@ -18,16 +18,13 @@
     "@theia/console": "^{{ version }}",
     "@theia/core": "^{{ version }}",
     "@theia/debug": "^{{ version }}",
-    "@theia/debug-nodejs": "^{{ version }}",
```
benoitf: just a note: should debug nodejs VS Code extension should be added somehow in the typescript che plugin ?

azatsarynnyy: @benoitf I think yes, tslint and nodejs VS Code extension should be included into typescript che plugin
I'll propose a PR to plugin registry

benoitf:
I think that yarn.lock should be bumped to theia 0.16.0 as well ?

generator/src/templates/assembly-package.mst
azatsarynnyy: provided PR https://github.com/eclipse/che-plugin-registry/pull/400

PR #644, opened 2020-02-27
fix RELEASE instructions so we don't end up...

nickboldt:
fix RELEASE instructions so we don't end up with an unbuildable branch

Change-Id: I8f396f12e40a6ab9f536bb5c2c033446f4b1ea48


nickboldt:
An even better process here would be to TAG the commit we want so we can fetch it as a tag, instead of pulling master and then looking for a specific commit in the past, some 40 ish commits ago.

nickboldt:
crw-ci-test

nickboldt:
another day, another "wait for CI to pass before I can update some documentation" :(

benoitf:
the issue is on the build part

nickboldt:
What about if they just produce a weekly tag, eg., an "integration build" ? Then it's generic and not product specific, but might make downstream adoption easier (again, it avoid "just grab a SHA").

Anyway I'll maintain the CRW fork for the purpose of having a tag. 

OR we remove the `--depth 1` or `--depth 40` flag entirely from any che-theia branch builds, so it doesn't fail as I was seeing when rebuilding 7.9.0 a week (and 44 commits) later.

If we do that I'm less fussed about using a tag, since master will have ALL the commits. It just makes the checkout process a lot slower as it's pulling to depth infinity.

PR #645, opened 2020-02-28
feat(plugin-remote): Allow to see plug-ins running in other containers

benoitf:
### What does this PR do?
See plug-ins across containers

Also it handles:
- disallow automatic dependencies download from extension referencing other extensions
- proxify exports from one container to other containers
- make exports calls synchronous by using deasync library
- Handle function() parameters on root exported objects. In that case it's handled as callbacks. So the function will be called remotely when needed.
- make deasync module a nodejs module (by reusing a custom pre-assembly of nodejs so that the theia endpoint runtime can be statically compiled
- change plugin's host process to be able to customize the plug-in manager to handle remote containers

Can be tested with https://che.openshift.io/f/?url=https://gist.githubusercontent.com/benoitf/618ba31e7812a35f8f6569b94e1672e4/raw/f002e493bb5a465119aeee9ebcd8928d8dab7d6a/devfile-across-containers.yaml

it'll deploy:
- one sidecar container with only `kubernetes vsix plugin`
- one sidecar container with only `vscode-yaml plugin`

When workspace is loaded, you may try to create a new yaml file
Paste content of https://kubernetes.io/docs/tasks/run-application/run-stateless-application-deployment/#creating-and-exploring-an-nginx-deployment 

```
apiVersion: apps/v1 # for versions before 1.9.0 use apps/v1beta2
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 2 # tells deployment to run 2 pods matching the template
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.7.9
        ports:
        - containerPort: 80
```

Some warnings should be there.
So kubernetes VS Code extension would have successfully register kubernetes schema into yaml VS Code extension.
image

Fixes https://github.com/eclipse/che/issues/15826
Fixes https://github.com/eclipse/che/issues/15844

Change-Id: I8134e4f8a704ac2993ce0580b11d153de8bb64ed

### DIFF
--- a/dockerfiles/theia-endpoint-runtime-binary/Dockerfile
+++ b/dockerfiles/theia-endpoint-runtime-binary/Dockerfile
@@ -8,14 +8,17 @@
 # Contributors:
 #   Red Hat, Inc. - initial API and implementation
 
+FROM quay.io/eclipse/che-custom-nodejs-deasync:10.19.0-latest as custom-nodejs
+
 FROM #{INCLUDE:docker/${BUILD_IMAGE_TARGET}/builder-from.dockerfile}
 
 WORKDIR /home/theia
 
 # Apply node libs installed globally to the PATH
 ENV PATH=${HOME}/.yarn/bin:${PATH}
-ENV NEXE_FLAGS="-t alpine-x64-10.16.0"
+ENV NEXE_FLAGS="--asset ${HOME}/pre-assembly-nodejs-static"
 
+COPY --from=custom-nodejs /pre-assembly-nodejs-static ${HOME}/pre-assembly-nodejs-static
 
 # setup extra stuff
 #{INCLUDE:docker/${BUILD_IMAGE_TARGET}/builder-setup.dockerfile}
--- a/extensions/eclipse-che-theia-plugin-remote/src/node/plugin-host-custom.ts
+++ b/extensions/eclipse-che-theia-plugin-remote/src/node/plugin-host-custom.ts
@@ -1,0 +1,103 @@
+/*********************************************************************
+ * Copyright (c) 2020 Red Hat, Inc.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ **********************************************************************/
+
+/**
+ * Copy of the upstream plugin-remote-init except that we add remote stuff
+ * const pluginRemoteBrowser = rpc.getProxy(MAIN_REMOTE_RPC_CONTEXT.PLUGIN_REMOTE_BROWSER);
+ * const pluginRemoteNodeImpl = new PluginRemoteNodeImpl(pluginRemoteBrowser);
+ * rpc.set(MAIN_REMOTE_RPC_CONTEXT.PLUGIN_REMOTE_NODE, pluginRemoteNodeImplt);
+ *
+ * and
+ *
+ * pluginRemoteNodeImpl.setPluginManager(pluginManager);
+ */
+import { Emitter } from '@theia/core/lib/common/event';
+import { RPCProtocolImpl } from '@theia/plugin-ext/lib/common/rpc-protocol';
+import { PluginHostRPC } from '@theia/plugin-ext/lib/hosted/node/plugin-host-rpc';
+import { PluginRemoteNodeImpl } from './plugin-remote-node-impl';
+import { MAIN_REMOTE_RPC_CONTEXT } from '../common/plugin-remote-rpc';
+console.log('PLUGIN_HOST_CHE_THEIA(' + process.pid + ') starting instance');
+
+// override exit() function, to do not allow plugin kill this node
+process.exit = function (code?: number): void {
+    const err = new Error('An plugin call process.exit() and it was prevented.');
+    console.warn(err.stack);
+} as (code?: number) => never;
+
+// same for 'crash'(works only in electron)
+// tslint:disable-next-line: no-any
+const proc = process as any;
+if (proc.crash) {
+    proc.crash = function (): void {
+        const err = new Error('An plugin call process.crash() and it was prevented.');
+        console.warn(err.stack);
+    };
+}
+
+process.on('uncaughtException', (err: Error) => {
+    console.error(err);
+});
+
+// tslint:disable-next-line: no-any
+const unhandledPromises: Promise<any>[] = [];
+
+// tslint:disable-next-line: no-any
+process.on('unhandledRejection', (reason: any, promise: Promise<any>) => {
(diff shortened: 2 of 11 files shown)


benoitf:
[ci-build]

benoitf:
[ci-build]

dockerfiles/theia-endpoint-runtime-binary/Dockerfile
```diff
@@ -8,14 +8,17 @@
 # Contributors:
 #   Red Hat, Inc. - initial API and implementation
 
+FROM quay.io/eclipse/che-custom-nodejs-deasync:10.19.0-latest as custom-nodejs
```
azatsarynnyy: @nickboldt I don't know whether it affects CRW build somehow but just FYI

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-node-impl.ts
```diff
@@ -0,0 +1,253 @@
...
+        const pluginManagerInternal = (this.pluginManager as any);
+
+        // add into the activatedPlugins stuff
+        // tslint:disable-next-line: no-any
+        const activatedPlugin: any = {};
+        // tslint:disable-next-line: no-any
+        const proxyExports: any = {};
+        activatedPlugin.exports = proxyExports;
+        const remoteBrowser = this.pluginRemoteBrowser;
+        const deasyncPromise = this.deasyncPromise;
```
tsmaeder: Not sure what this is for.

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-node-impl.ts
```diff
@@ -0,0 +1,253 @@
...
+        // add proxy
+        const callId = this.callId++;
+        proxyNames.forEach(entryName => {
+            console.log(`Proxyfing exports ${hostId} ${pluginId}/${entryName}`);
+            // tslint:disable-next-line: no-any
+            proxyExports[entryName] = (...args: any[]) => {
+                // need to keep arguments only if arguments have functions as functions can't be propagated remotely
+                const hasFunctions = args.some(arg => typeof arg === 'function');
```
tsmaeder: There are frameworks that we can use for remoting like https://github.com/stanford-oval/transparent-rpc It even handles function call on the objects you pass as arguments.

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-node-impl.ts
```diff
@@ -0,0 +1,253 @@
...
+        });
+        pluginManagerInternal.activatedPlugins.set(pluginId, activatedPlugin);
+        await pluginManagerInternal.activatePlugin(pluginId);
+    }
+
+    // tslint:disable-next-line: no-any
+    async $callLocalMethod(callId: number, index: number, ...args: any[]): Promise<any> {
+        const callInfo = this.calls.get(callId);
+        if (callInfo) {
+            const invocatioObject = callInfo.functions.get(index);
```
tsmaeder: I think "functionToInvoke" would be a better name. The object lifetime is longer than a single invocation.

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-node-impl.ts
```diff
@@ -0,0 +1,253 @@
...
+            if (!invocatioObject) {
+                throw new Error('Requesting a local-call method to an argument that was not stored');
```
tsmaeder: "...to a function that is not defined"?

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-node-impl.ts
benoitf: `this` will be a different object in inner objects.

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-node-impl.ts
benoitf: This framework doesn't do more than the RPC framework we already have on plugins. (BTW we have multi-plexing in theia RPC)
And functions are already handled by the underlying RPC protocol.

The real issue is for callbacks parameters that should not be sent: A function that returns a function

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-host-custom.ts
```diff
@@ -0,0 +1,93 @@
...
+
+process.on('uncaughtException', (err: Error) => {
+    console.error(err);
+});
+
+// tslint:disable-next-line: no-any
+const unhandledPromises: Promise<any>[] = [];
+
+// tslint:disable-next-line: no-any
+process.on('unhandledRejection', (reason: any, promise: Promise<any>) => {
```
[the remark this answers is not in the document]
benoitf: it's upstream code
https://github.com/eclipse-theia/theia/blob/bc16fc18ab16a34e8818ae525c6f8a20133baeae/packages/plugin-ext/src/hosted/node/plugin-host.ts#L46

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-host-custom.ts
```diff
@@ -0,0 +1,93 @@
...
+import { Emitter } from '@theia/core/lib/common/event';
+import { RPCProtocolImpl } from '@theia/plugin-ext/lib/common/rpc-protocol';
+import { PluginHostRPC } from '@theia/plugin-ext/lib/hosted/node/plugin-host-rpc';
+import { PluginRemoteNodeImpl } from './plugin-remote-node-impl';
+import { MAIN_REMOTE_RPC_CONTEXT } from '../common/plugin-remote-rpc';
+console.log('PLUGIN_HOST_CHE_THEIA(' + process.pid + ') starting instance');
+
+// override exit() function, to do not allow plugin kill this node
+process.exit = function (code?: number): void {
+    const err = new Error('An plugin call process.exit() and it was prevented.');
```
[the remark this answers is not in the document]
benoitf: it's upstream code
https://github.com/eclipse-theia/theia/blob/bc16fc18ab16a34e8818ae525c6f8a20133baeae/packages/plugin-ext/src/hosted/node/plugin-host.ts#L23

dockerfiles/theia-endpoint-runtime-binary/Dockerfile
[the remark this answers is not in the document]
benoitf: native modules can't be loaded dynamically with static nodejs
but we need static binary to work on alpine runtime.
so deasync is packaged as part of the nodejs runtime

PR #664, opened 2020-03-12
Patch main.js to proper handling hotkeys in webviews

vzhukovskii:
### What does this PR do?
Add patch that filters `CtrlCmd+O`, `CtrlCmd+P`, `CtrlCmd+S` and `F1` hotkeys to prevent their default execution in webviews.

https://github.com/eclipse/che/issues/15763

#### Release Notes
N/A

#### Docs PR
N/A


vzhukovskii:
@benoitf we can, but for this moment fix in upstream as I see should be more complex and may take much more time

vzhukovskii:
crw-ci-test

benoitf:
yeah but I don't think we should do patches for this ? as it's mainly an upstream issue

vzhukovskii:
Fixed in upstream

benoitf:
yes the fix you did upstream is definitely better ! thx

PR #668, opened 2020-03-16
Use the latest yeoman-plugin

vitaliy-guliy:
Signed-off-by: Vitaliy Guliy <vgulyy@redhat.com>

### What does this PR do?
Includes the latest version of `@theia/yeoman-plugin` to Che-Theia build.

It's cherry-picking of the commit from the PR https://github.com/eclipse/che-theia/pull/666

### What issues does this PR fix or reference?
https://github.com/eclipse/che/issues/16354

PR #666, opened 2020-03-13
Use the latest version of yeoman-plugin

vitaliy-guliy:
Signed-off-by: Vitaliy Guliy <vgulyy@redhat.com>

### What does this PR do?
Includes the latest version of `@theia/yeoman-plugin` to Che-Theia build.

### What issues does this PR fix or reference?
https://github.com/eclipse/che/issues/16354

PR #670, opened 2020-03-16
Use the latest yeoman-plugin (#668) (part...

nickboldt:
Use the latest yeoman-plugin (#668) (part 2)

Change-Id: I43fe01fbc2275b8d32219a54bbc36590af1711f6


nickboldt:
pushed this into #666 and the happy paths passed.

PR #672, opened 2020-03-18
Fix import of zip project in case the archive is hosted on the same cluster which uses self-signed certificate

mmorhun:

### What does this PR do?
Fixes import of zip project when it is hosted in the same cluster as Che and the cluster uses self-signed certificate.

### DIFF
--- a/plugins/workspace-plugin/src/theia-commands.ts
+++ b/plugins/workspace-plugin/src/theia-commands.ts
@@ -204,7 +204,7 @@ export class TheiaImportZipCommand implements TheiaImportCommand {
         const importZip = async (progress: theia.Progress<{ message?: string; increment?: number }>, token: theia.CancellationToken): Promise<void> => {
             try {
                 // download
-                const curlArgs = ['-L', '--output', this.zipfilePath];
+                const curlArgs = ['-sL', '--output', this.zipfilePath];
                 if (fs.existsSync(SS_CRT_PATH)) {
                     curlArgs.push('-k');
                 }
--- a/plugins/workspace-plugin/src/theia-commands.ts
+++ b/plugins/workspace-plugin/src/theia-commands.ts
@@ -17,6 +17,7 @@ import * as fileuri from './file-uri';
 import { execute } from './exec';
 import * as git from './git';
 
+const SS_CRT_PATH = '/tmp/che/secret/ca.crt';
 const CHE_TASK_TYPE = 'che';
 
 /**
@@ -204,6 +205,9 @@ export class TheiaImportZipCommand implements TheiaImportCommand {
             try {
                 // download
                 const wgetArgs = [this.locationURI!, '-O', this.zipfilePath];
+                if (fs.existsSync(SS_CRT_PATH)) {
+                    wgetArgs.push('--no-check-certificate');
+                }
                 await execute('wget', wgetArgs);
 
                 // expand
--- a/plugins/workspace-plugin/src/theia-commands.ts
+++ b/plugins/workspace-plugin/src/theia-commands.ts
@@ -204,11 +204,12 @@ export class TheiaImportZipCommand implements TheiaImportCommand {
         const importZip = async (progress: theia.Progress<{ message?: string; increment?: number }>, token: theia.CancellationToken): Promise<void> => {
             try {
                 // download
-                const wgetArgs = [this.locationURI!, '-O', this.zipfilePath];
+                const curlArgs = ['-L', '--output', this.zipfilePath];
                 if (fs.existsSync(SS_CRT_PATH)) {
-                    wgetArgs.push('--no-check-certificate');
+                    curlArgs.push('-k');
                 }
-                await execute('wget', wgetArgs);
+                curlArgs.push(this.locationURI!);
+                await execute('curl', curlArgs);
 
                 // expand
                 fs.mkdirSync(this.projectDir);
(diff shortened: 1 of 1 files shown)


mmorhun:
I would like reviewers to try to test it too.

plugins/workspace-plugin/src/theia-commands.ts
```diff
@@ -204,6 +205,9 @@ export class TheiaImportZipCommand implements TheiaImportCommand {
             try {
                 // download
                 const wgetArgs = [this.locationURI!, '-O', this.zipfilePath];
+                if (fs.existsSync(SS_CRT_PATH)) {
```
davidfestal: As commented in the [JIRA issue](https://issues.redhat.com/browse/CRW-728?focusedCommentId=14002634&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14002634), It seems that `wget` doesn't correctly support proxies.
I would suggest using `curl` instead of `wget`.
This way you would also be able to add the certificate authority.
I tested it with proxies + tls + self-signed certificates, inside a `che-theia` container of a Che installation on a disconnected OS 4.3 installation, and it works great.

This would mainly give:
```typescript
const curlArgs = [this.locationURI!, '-sL', '--output', this.zipfilePath];
if (fs.existsSync(SS_CRT_PATH)) {
    curlArgs.push('--cacert ');
    curlArgs.push(SS_CRT_PATH);
}
await execute('curl', curlArgs);
```

@mmorhun Do you think it would be worth updating the PR ?

cc @nickboldt who also proposed to use `curl` instead.

mmorhun: @davidfestal I do not mind using `curl` instead of `wget` and even like it if it fixes proxy issue.
But when I tested your suggestion it didn't work on my side. Probably, `curl` treats given certificate as only one trusted. So we should either be sure that the endpoint has ours self-signed certificate or just trust everyone. So, instead of `--cacert` we have to use `-k`.

davidfestal: Sure. I understand. I was mainly thinking of the case we import zip files from the devfile registry in the offline use-case. But sure, zip import should also work for external zip projects.
I'll test the `curl` command with `-k` on my side with the proxy to ensure it's OK, and will confirm here.

mmorhun:
Rebased

plugins/workspace-plugin/src/theia-commands.ts
davidfestal: So I tested, and the using:

```bash
curl <url> -kL --output <local file>
```

Also works great with proxies

davidfestal: BTW, why not add the `-k` in all cases, if we really want to be able to download any zip with self-signed certificate ?
However does it involve any security issue ?

davidfestal:
And would it be possible to have a corresponding PR on the `7.9.x` branch. That's quite important for the `CRW 2.1` ?

mmorhun:
@davidfestal yes, sure, I'll provide a backport.

plugins/workspace-plugin/src/theia-commands.ts
mmorhun: We may add `-k` for all cases, but actually we might need it only when self-signed certificate is used. So I just followed this logic.

mmorhun:
From the CI build logs:
```
@eclipse-che/theia-remote: [compile] src/node/plugin-remote-init.ts(377,7): error TS2420: Class 'PluginDeployerHandlerImpl' incorrectly implements interface 'PluginDeployerHandler'.
@eclipse-che/theia-remote: [compile]   Property 'undeployPlugin' is missing in type 'PluginDeployerHandlerImpl' but required in type 'PluginDeployerHandler'.
```
And it is not related to current PR...

azatsarynnyy:
> From the CI build logs:
> 
> ```
> @eclipse-che/theia-remote: [compile] src/node/plugin-remote-init.ts(377,7): error TS2420: Class 'PluginDeployerHandlerImpl' incorrectly implements interface 'PluginDeployerHandler'.
> @eclipse-che/theia-remote: [compile]   Property 'undeployPlugin' is missing in type 'PluginDeployerHandlerImpl' but required in type 'PluginDeployerHandler'.
> ```
> 
> And it is not related to current PR...

@mmorhun the fix is on the way https://github.com/eclipse/che-theia/pull/676

plugins/workspace-plugin/src/theia-commands.ts
davidfestal: got it

davidfestal:
@mmorhun do you also plan to remove the `-s` flag here (as done on the backported PR ?)

mmorhun:
@davidfestal yes, but wait when the CI build is fixed to not to build twice

nickboldt:
Instead of removing `-s` we can use `-sS` to be quiet but still throw errors if errors happen.

```
-S/--show-error
When used with -s it makes curl show an error message if it fails.
```

azatsarynnyy:
@mmorhun https://github.com/eclipse/che-theia/pull/676 is merged. It adapts Che Theia to the latest upstream Theia changes.
So, the build is fixed but the Happy Path tests are failing because of https://github.com/eclipse/che/issues/16403
QE Team is working on it.

PR #678, opened 2020-03-19
Fix import of zip project which hosted with cluster self-signed certificate

mmorhun:
### What does this PR do?
Backports https://github.com/eclipse/che-theia/pull/672 into 7.9.x branch

### DIFF
--- a/plugins/workspace-plugin/src/theia-commands.ts
+++ b/plugins/workspace-plugin/src/theia-commands.ts
@@ -204,7 +204,7 @@ export class TheiaImportZipCommand implements TheiaImportCommand {
         const importZip = async (progress: theia.Progress<{ message?: string; increment?: number }>, token: theia.CancellationToken): Promise<void> => {
             try {
                 // download
-                const curlArgs = ['-L', '--output', this.zipfilePath];
+                const curlArgs = ['-sSL', '--output', this.zipfilePath];
                 if (fs.existsSync(SS_CRT_PATH)) {
                     curlArgs.push('-k');
                 }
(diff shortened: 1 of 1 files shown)


davidfestal:
@nickboldt @dmytro-ndp Do you want to approve before merging this on the `7.9.x` branch ?

plugins/workspace-plugin/src/theia-commands.ts
```diff
@@ -203,8 +204,12 @@ export class TheiaImportZipCommand implements TheiaImportCommand {
         const importZip = async (progress: theia.Progress<{ message?: string; increment?: number }>, token: theia.CancellationToken): Promise<void> => {
             try {
                 // download
-                const wgetArgs = [this.locationURI!, '-O', this.zipfilePath];
-                await execute('wget', wgetArgs);
+                const curlArgs = ['-L', '--output', this.zipfilePath];
```
nickboldt: should we also add `-s` and `-S` flags?

nickboldt:
Seems legit, just one question about suppressing console stuff

plugins/workspace-plugin/src/theia-commands.ts
davidfestal: we decided not to add the silent flag, in order to keep the error message on the output in case of failure. Does it make sense to you ?

nickboldt:
```
-S/--show-error
When used with -s it makes curl show an error message if it fails.
```

davidfestal:
LGTM. Thanks @mmorhun !

davidfestal:
@nickboldt Maybe you could add a review comment on the main PR https://github.com/eclipse/che-theia/pull/672 to add the `sS` flags as well ?

nickboldt:
Sure -- https://github.com/eclipse/che-theia/pull/672/files#diff-a432c622941cff7771436a69c991ab3cR207

PR #698, opened 2020-04-07
fix(vscode-uri): Changes due to upstream changes of vscode-uri

benoitf:
### What does this PR do?
adapt code to upstream changes

https://github.com/eclipse-theia/theia/pull/7506

Change-Id: I8a580fe99eb3bdb17e69e5eeacbd523b712eaa0b

### DIFF
--- a/extensions/eclipse-che-theia-plugin-ext/src/plugin/che-sidecar-content-reader.ts
+++ b/extensions/eclipse-che-theia-plugin-ext/src/plugin/che-sidecar-content-reader.ts
@@ -10,7 +10,7 @@
 
 import { RPCProtocol } from '@theia/plugin-ext/lib/common/rpc-protocol';
 import * as fs from 'fs-extra';
-import URI from 'vscode-uri';
+import { URI } from 'vscode-uri';
 import { CheSideCarContentReader, PLUGIN_RPC_CONTEXT } from '../common/che-protocol';
 
 export class CheSideCarContentReaderImpl implements CheSideCarContentReader {
--- a/extensions/eclipse-che-theia-plugin-remote/src/node/che-content-aware-utils.ts
+++ b/extensions/eclipse-che-theia-plugin-remote/src/node/che-content-aware-utils.ts
@@ -9,7 +9,7 @@
  **********************************************************************/
 
 import * as theia from '@theia/plugin';
-import URI from 'vscode-uri';
+import { URI } from 'vscode-uri';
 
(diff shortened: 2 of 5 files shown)


benoitf:
@RomanNikitenko well the plug-in has its own dependency so it's fine.
Plug-ins are self contained and can use their own vscode-uri dependencies

RomanNikitenko:
> Plug-ins are self contained and can use their own vscode-uri dependencies

thanks!

I have another question related to dependencies.

For example, `CheSideCarContentReaderImpl` uses `vscode-uri` dependency.
I understand that the dependency comes anyway because `che-theia` depends on `theia` and `theia` uses `vscode-uri`.

But the question for me: 
should we define `vscode-uri` dependency explicitly in the corresponding `package.json` file of `@eclipse-che/theia-plugin-ext`?

Does it make sense?

RomanNikitenko:
> Plug-ins are self contained and can use their own vscode-uri dependencies

according to the answer we are not forced to update `vscode-ws-jsonrpc` and keep close to upstream (looks like it's defined explicitly only in `package.json` file of `task-plugin` ).

Just `nice to have` using more fresh versions of libs.
Right?

benoitf:
@RomanNikitenko yes we should have explicit dependencies for direct dependencies

I think there are some eslint/tslint rules and we should enable them.

benoitf:
>> Plug-ins are self contained and can use their own vscode-uri dependencies

> btw: according to the answer we are not forced to update vscode-ws-jsonrpc and keep close to
> upstream (looks like it's defined explicitly only in package.json file of task-plugin ).
> Just nice to have using more fresh versions of libs.
> Right?

yes would be nice to always keep to latest dependencies

RomanNikitenko:
@benoitf 
thank you very much for the answers, now it's more clear for me!

benoitf:
but happy path is not happy, need to investigate

azatsarynnyy:
> but happy path is not happy, need to investigate

it looks like it should be just rerun

azatsarynnyy:
crw-ci-test

RomanNikitenko:
crw-ci-test

RomanNikitenko:
crw-ci-test

benoitf:
crw-ci-test

benoitf:
crw-ci-test

benoitf:
I added a commit fix for nsfw (I can't reproduce it on my env and it was not the case as well for ci-build check) but it was failing on happy path

RomanNikitenko:
Theia has open issue for `nsfw build error`

PR #699, opened 2020-04-07
Fix security issues and bump some component versions

benoitf:
### What does this PR do?
you may look at each commit to get specific details

https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-7598 reported by github scan


generator/package.json
```diff
@@ -54,10 +53,6 @@
     "format-code": "tsfmt -r",
     "tslint-fix": "tslint --fix --project .",
     "tslint": "tslint --project .",
-    "exe": "yarn run exe:alpine && yarn run exe:mac && yarn run exe:linux",
```
ericwill, line 57: How come these were removed?

generator/package.json
```diff
@@ -54,10 +53,6 @@
...
-    "exe:alpine": "rimraf exe/alpine/ && mkdir -p exe/alpine && nexe dist/yargs.js -t alpine-x64-8.12.0 -o exe/alpine/che-theia",
-    "exe:mac": "rimraf exe/macOS/ && mkdir -p exe/macOS && nexe dist/yargs.js -t mac-x64-8.12.0 -o exe/macOS/che-theia",
-    "exe:linux": "rimraf exe/linux/ && mkdir -p exe/linux && nexe dist/yargs.js -t linux-x64-8.12.0 -o exe/linux/che-theia",
```
azatsarynnyy, line 60: If we aren't providing the generator as a native binary, then the doc has to be updated correspondingly
https://github.com/eclipse/che-theia/blob/aa38014db4e41afba47d08bbc6450a1fdfc14d1e/generator/CONTRIBUTING.md#native-executable

generator/package.json
benoitf: hello, it comes from the commit

```
feat(nexe): remove nexe from the generator as now it's built from the theia repository
```

as it seems that now we don't need anymore the native binary as the generator is already provided in che-theia repository

generator/package.json
benoitf: ok, agreed

maybe let's discuss if it's fine to remove it or not ?
I was the one bringing native binary but then I don't know if we still need it as it's now inside this repo

azatsarynnyy: I don't see why one would need Che Theia generator in the form of a native binary. I think it's good enough to provide a single way of installing/running Che Theia generator - as npm "binary" only.

benoitf:
@azatsarynnyy fixed commit with nexe and CONTRIBUTING.md  / native binaries

dmytro-ndp:
crw-ci-test

PR #700, opened 2020-04-07
Update dependencies

RomanNikitenko:

### What does this PR do?
- upgrade `vscode-uri` version to `^2.1.1` for `task-plugin`
- upgrade `vscode-ws-jsonrpc` to `^0.2.0` for `task-plugin`
- define explicitly `vscode-uri` and `vscode-ws-jsonrpc` in the corresponding `package.json` files  

https://github.com/eclipse/che/issues/16552


RomanNikitenko:
> maybe we should stick to fixed versions instead of caret versions ?

applied this approach

PR #701, opened 2020-04-07
Fix incorrect usage of THEIA_GIT_REFS env. variable

azatsarynnyy:
### What does this PR do?
Fixes incorrect usage of `THEIA_GIT_REFS` env. variable.
`THEIA_GIT_REFS` env. variable is used for invalidating docker cache when upstream Theia has been changed.
But there're some places when the variable is used as if it relates to `eclipse/che-theia` repo.

### DIFF
--- a/set_theia_version.sh
+++ b/set_theia_version.sh
@@ -71,8 +71,6 @@ change_change_che_theia_init_baranch () {
         fi
     done < "che-theia-init-sources.yml";
     printf "%b\n" "$content" > "che-theia-init-sources.yml";
-
-    sed -i 's~@api.github.com/repos/eclipse/che-theia/git/${GIT_REF}~@api.github.com/repos/eclipse/che-theia/git/refs/heads/'${branchName}'~' ./dockerfiles/theia/Dockerfile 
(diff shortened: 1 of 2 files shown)


set_theia_version.sh
```diff
@@ -71,8 +71,6 @@ change_change_che_theia_init_baranch () {
         fi
     done < "che-theia-init-sources.yml";
     printf "%b\n" "$content" > "che-theia-init-sources.yml";
-
-    sed -i 's~@api.github.com/repos/eclipse/che-theia/git/${GIT_REF}~@api.github.com/repos/eclipse/che-theia/git/refs/heads/'${branchName}'~' ./dockerfiles/theia/Dockerfile
```
azatsarynnyy, line 75: This edit actually does nothing as there no such command in the target file.
The correct value for `GIT_REF` env. variable comes from `build.include`, see the corresponding sed edit
https://github.com/eclipse/che-theia/blob/8fc54d6e8629d7fd405ea243498310c961e050b8/set_theia_version.sh#L128

PR #702, opened 2020-04-08
feat(eslint): switch extensions and plugins from tslint to eslint as for upstream theia

benoitf:
### What does this PR do?
Use of eslint instead of tslint as linter for che-theia extensions and plugins

https://github.com/eclipse/che/issues/16070


extensions/eclipse-che-theia-plugin-ext/src/browser/oauth-utils.ts
```diff
@@ -42,7 +42,7 @@ export class OauthUtils {
 
     async isAuthenticated(provider: string): Promise<boolean> {
         const token = await this.cheApiService.getOAuthToken(provider);
-        return token !== null;
+        return token !== undefined;
```
benoitf, line 45: @vinokurig is it ok to replace `!=== null` by undefined as reported by eslint ?

benoitf:
crw-ci-test

extensions/eclipse-che-theia-plugin-ext/src/browser/oauth-utils.ts
vinokurig: It's doesn't matter because I am going to rework it in this PR: https://github.com/eclipse/che-theia/pull/694/files#diff-c08dd0360535189a619213d76179e0e5R71-R78

dmytro-ndp:
crw-ci-test

PR #704, opened 2020-04-09
Add rules eslint extensions

benoitf:
### What does this PR do?
Add new rules reported by https://github.com/eclipse/che/issues/16070 that can be automated with eslint
Also apply some fixes manually

Fixes https://github.com/eclipse/che/issues/16070


benoitf:
I need to fix something for docker builds (not che-in-che) to make it fully working

PR #709, opened 2020-04-10
[WIP] Add non-interactive release script

azatsarynnyy:

### What does this PR do?
Replaces the existing release script with a new one that is executed in a non-interactive mode that should allow using it within [the automated Che Release Process](https://github.com/eclipse/che/issues/16499).

The script performs all the necessary operations for a particular type of a release (major/minor/bugfix), based on the given parameters.

closes https://github.com/eclipse/che/issues/16523


make-release.sh
```diff
@@ -0,0 +1,157 @@
+#!/bin/bash
```
benoitf, line 1: you may add copyright headers

nickboldt:
@azatsarynnyy  where do the values of `THEIA_VERSION` and `THEIA_COMMIT_SHA` come from? 

Can they be deduced from something upstream? Could we use a tag, eg., 7.11.1 or 7.12.0?

azatsarynnyy:
> @azatsarynnyy where do the values of `THEIA_VERSION` and `THEIA_COMMIT_SHA` come from?

@nickboldt usually, `THEIA_VERSION` is the latest `next`-version of Theia package published to npmjs registry. And `THEIA_COMMIT_SHA` is the commit from which that package was built.
I've described in [RELEASE.md](https://github.com/eclipse/che-theia/blob/7600331f7e09c4424721f8c2c9e126aa068a3d70/RELEASE.md#how-to-get-an-upstream-theia-version) how to get these two parameters.

> Can they be deduced from something upstream?

It would require fetching info from npmjs registry through some API. Not sure if it exists. And then, looking for a corresponding commit in the upstream repo.

> Could we use a tag, eg., 7.11.1 or 7.12.0?

The script can do both, minor and patch versions release.

make-release.sh
azatsarynnyy: added
thx!

nickboldt:
I was asking if instead of THEIA_COMMIT_SHA being a hard coded SHA, we could instead use a tag. That way we wouldn't need to manually compute the SHA for every release.

azatsarynnyy:
@nickboldt I've just realized that we don't have to use only a long SHA but a short one can be used as well. We can take it from an npm package version and provide it to `git checkout` command in dockerfile. So, `THEIA_COMMIT_SHA` parameter actually is not necessary.
I will rework it this week.

I like the idea of having `THEIA_VERSION` file instead of a cmd parameter. I'll try to apply it as well.

nickboldt:
> short one can be used as well

Yep! Git is smart enough to fetch from any SHA, provided that the SHA is unique. 

It's also smart enough that if you ask for a too-short value, you get the minimum length unique one:

```
$ git rev-parse --short=2 7e223ce3a026ffc3
7e22
```

benoitf:
If we go to a file with specific sha1/etc I think then we still need a parameter as well and  an updated build.sh script as most part of the script is not about release but about building.

What I meant is that we need a build script and then make-release can either read value from a file but users can still invoke build script with a parameter

BTW we have in the past discussed that we want to be able to build che-theia with a custom upstream PR (remote branch) to test early upstream PR so I'm unsure that it's fitting well with the new approach of using npmjs as well ?

make-release.sh
```diff
@@ -0,0 +1,154 @@
...
+apply_files_edits () {
+  # Che Theia release may depend on Theia next or latest
+  if [[ ${THEIA_VERSION} == *"-next."* ]]; then
+    THEIA_PATCHES_DIR=master
+    THEIA_BRANCH=master
+    THEIA_GIT_REFS=refs\\\\\\\\/heads\\\\\\\\/master
+  else 
+    THEIA_PATCHES_DIR=${THEIA_VERSION}
+    THEIA_BRANCH=v${THEIA_VERSION}
+    THEIA_GIT_REFS=refs\\\\\\\\/tags\\\\\\\\/v${THEIA_VERSION}
```
nickboldt, line 78: do you really need this many slashes??? 

If you run sed with the `-r` flag you probably don't have to use this many as it will use "extended regex"  and you shouldn't need to `\\` so many times. :D

make-release.sh
```diff
@@ -67,6 +66,12 @@ if [[ "${BASEBRANCH}" != "${BRANCH}" ]]; then
 fi
 
 apply_files_edits () {
+  read THEIA_VERSION < THEIA_VERSION
+  if [[ ! ${THEIA_VERSION} ]]; then
+    echo "THEIA_VERSION file is not found"; echo
```
nickboldt, line 71: I wonder if, to Florent's point, there's value in being able to override the value in the file with `--theia-version` flag?

azatsarynnyy:
> What I meant is that we need a build script and then make-release can either read value from a file but users can still invoke build script with a parameter

Sorry @benoitf but I didn't get your point here. The user still can invoke the build script. And this PR doesn't touch it.

> BTW we have in the past discussed that we want to be able to build che-theia with a custom upstream PR (remote branch) to test early upstream PR so I'm unsure that it's fitting well with the new approach of using npmjs as well ?

There's actually nothing new. That was the instructions that I've got with the previous release script and I documented it in the previous [RELEASE.md](https://github.com/eclipse/che-theia/blob/release/RELEASE.md#the-release-changes-to-depend-on-theia-commit).

make-release.sh
azatsarynnyy: it can be updated later if we'll need it

make-release.sh
azatsarynnyy: I quickly tried it but didn't work for me. So, let's keep it as is and improve later.

azatsarynnyy:
It was tested yesterday with 7.9.3 release. Everything went well :tada:
Other stuff/ideas can be improved/implemented later.
GH actions are failed with `no space left on device` :confused: 
So, I'm going to merge it once centos ci is happy.

benoitf:
@azatsarynnyy my point is that script should do 'release' stuff and move all 'build' stuff to build.sh script

because we could want to do custom builds by using nice stuff of this script but it's not related to making releases

benoitf:
FYI also .sh script is not passing on https://www.shellcheck.net/

benoitf:
and it doesn't work on macOS:

```shell
sed: -i may not be used with stdin
sed: -i may not be used with stdin
sed: 1: "./extensions/eclipse-ch ...": invalid command code .
sed: 1: "./plugins/github-auth-p ...": invalid command code .
sed: -i may not be used with stdin
sed: -i may not be used with stdin
sed: -i may not be used with stdin
```

azatsarynnyy:
> FYI also .sh script is not passing on https://www.shellcheck.net/

Nice tool!
I've created an issue https://github.com/eclipse/che/issues/16652

azatsarynnyy:
> and it doesn't work on macOS:
> 
> ```shell
> sed: -i may not be used with stdin
> sed: -i may not be used with stdin
> sed: 1: "./extensions/eclipse-ch ...": invalid command code .
> sed: 1: "./plugins/github-auth-p ...": invalid command code .
> sed: -i may not be used with stdin
> sed: -i may not be used with stdin
> sed: -i may not be used with stdin
> ```

reported https://github.com/eclipse/che/issues/16653

azatsarynnyy:
> @azatsarynnyy my point is that script should do 'release' stuff and move all 'build' stuff to build.sh script
> 
> because we could want to do custom builds by using nice stuff of this script but it's not related to making releases

I see. Thanks for the explanation! Created a separate task https://github.com/eclipse/che/issues/16654

PR #710, opened 2020-04-10
chore(github): Add GH action that allow to use parallel builds

benoitf:
### What does this PR do?
Introduce as experimental usage of github action for PR-build-check

It should be much more faster than ci.centos to perform the build check as we can use matrix jobs (and then build in parallel) and we don't need to setup the image/install software to run docker builds, etc.

if it works well, then we could switch to this check

N/A

Change-Id: I19f9db266294cd68e99a7d7c5b0e57ceaafae17a

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,31 @@
+#
+# Copyright (c) 2020 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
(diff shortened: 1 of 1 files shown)


benoitf:
not sure if github actions is enabled on this repository ?

azatsarynnyy:
> not sure if github actions is enabled on this repository ?

Should be enabled as we see `Actions` tab.
I've checked it on my fork: when disabled, the tab is absent.

benoitf:
ok, testing by merging it

benoitf:
seems to be triggered: https://github.com/eclipse/che-theia/actions/runs/75190544

PR #720, opened 2020-04-21
Fix 'is not a function' error

vinokurig:
### What does this PR do?
Add checks that fix `e.data.startsWith is not a function` error.

https://github.com/eclipse/che/issues/16630


extensions/eclipse-che-theia-plugin-ext/src/browser/oauth-utils.ts
```diff
@@ -35,7 +35,7 @@ export class OauthUtils {
             }
         });
         window.addEventListener('message', data => {
-            if (data.data.startsWith('token:') && this.oAuthPopup) {
+            if (this.oAuthPopup && data.data && (typeof data.data === 'string') && data.data.startsWith('token:')) {
```
azatsarynnyy, line 38: `data.data` check is redundant if the check for `typeof` is present

vinokurig: fixed

PR #722, opened 2020-04-22
feat(product): Add Eclipse Che as product name

benoitf:
### What does this PR do?
Use Eclipse Che as product name (like in dialog box)
Also with https://github.com/eclipse-theia/theia/pull/7642 it will allow to use that name inside webviews, etc.

Fixes https://github.com/eclipse/che/issues/16677

Change-Id: I0881b1749e69502265d89b4048e92eea3bd48f92


benoitf:
cc @scela

PR #724, opened 2020-04-23
chore(docker): Use --cache-from by default

benoitf:
### What does this PR do?
Use --cache-from by default when building images
basically, theia-dev image shouldn't be rebuilt if there is no change in generator or in Dockerfile of theia-dev

https://github.com/eclipse/che/issues/16463

Change-Id: Iad02615369764934460406d978879bff863195f9

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -41,4 +41,6 @@ jobs:
     - name: build
       run: |
         docker image prune -a -f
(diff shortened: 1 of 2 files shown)


benoitf:
@dmytro-ndp FYI this PR is breaking Happy Path PR check as it's changing default name of docker image so need to update happy path check job

PR #725, opened 2020-04-24
allow to provide yarn root options on build.sh + new job in github action

benoitf:
### What does this PR do?
allow to provide yarn root options on build.sh: it's allowing us to skip lot of post-scripts from the first `yarn` call
When building the images we don't need to build all modules and to grab and build all native dependencies.

Then, add a new job to perform this check separately

one job to check full nodejs build of che-theia
two parallel jobs: one for alpine, one for ubi8

### DIFF
--- a/build.include
+++ b/build.include
@@ -16,8 +16,16 @@ THEIA_BRANCH="master"
 THEIA_GIT_REFS="refs\\/heads\\/master"
 THEIA_DOCKER_IMAGE_VERSION=
 
+prepare_yarn_root_args() {
+    IFS=',' read -r -a YARN_ARGS_ARRAY <<< "$@"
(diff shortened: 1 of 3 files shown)


benoitf:
ci-build

build.include
```diff
@@ -27,6 +35,10 @@ parse() {
       --push)
         PUBLISH_IMAGES=true
       shift ;;
+      --root-yarn-opts*:*)
+        ROOT_YARN_CSV="${1#*:}"
```
azatsarynnyy, line 39: what does `CSV` mean?

benoitf: comma separated value
I copy it from there 
https://github.com/eclipse/che-theia/blob/master/dockerfiles/build.include#L102-L105

PR #729, opened 2020-04-29
Fix stop working terminal widgets when user clicks on the exited task widget.

AndrienkoAleksandr:
### What does this PR do?
Fix stop working terminal widgets when user clicks on the exited task widget.

https://github.com/eclipse/che/issues/16601

### DIFF
--- a/extensions/eclipse-che-theia-terminal/src/browser/terminal-widget/remote-terminal-widget.ts
+++ b/extensions/eclipse-che-theia-terminal/src/browser/terminal-widget/remote-terminal-widget.ts
@@ -63,6 +63,8 @@ export class RemoteTerminalWidget extends TerminalWidgetImpl {
 
     private isOpen: boolean = false;
     protected channel: OutputChannel;
+    protected closeOutputConnectionDisposable: Disposable;
+    protected processGone: boolean;
 
     @postConstruct()
     protected init(): void {
@@ -71,8 +73,11 @@ export class RemoteTerminalWidget extends TerminalWidgetImpl {
 
         this.toDispose.push(this.remoteTerminalWatcher.onTerminalExecExit(exitEvent => {
             if (this.terminalId === exitEvent.id) {
+                this.processGone = true;
                 if (this.options.closeWidgetOnExitOrError) {
                     this.dispose();
+                } else {
+                    this.closeOutputConnectionDisposable.dispose();
                 }
                 this.onTermDidClose.fire(this);
                 this.onTermDidClose.dispose();
@@ -89,8 +94,11 @@ export class RemoteTerminalWidget extends TerminalWidgetImpl {
         const missedPrivilegesErr = 'cannot create resource "pods/exec"';
         this.toDispose.push(this.remoteTerminalWatcher.onTerminalExecError(errEvent => {
             if (this.terminalId === errEvent.id) {
+                this.processGone = true;
                 if (this.options.closeWidgetOnExitOrError) {
                     this.dispose();
+                } else {
+                    this.closeOutputConnectionDisposable.dispose();
                 }
                 this.onTermDidClose.fire(this);
                 this.onTermDidClose.dispose();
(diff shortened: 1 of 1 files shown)


extensions/eclipse-che-theia-terminal/src/browser/terminal-widget/remote-terminal-widget.ts
```diff
@@ -272,7 +291,11 @@ export class RemoteTerminalWidget extends TerminalWidgetImpl {
     }
 
     protected resizeTerminalProcess(): void {
-        if (typeof this.terminalId !== 'number') {
+        if (typeof this.terminalId !== 'number' || this.processGone) {
```
RomanNikitenko, line 294: Looks like `typeof this.terminalId !== 'number'` is redundant, because on the next step we have `IBaseTerminalServer.validateId(this.terminalId)` which does [the same](https://github.com/eclipse-theia/theia/blob/dcb9c9c32f8be3412eaa8c9cf6da135b6f959a7f/packages/terminal/src/common/base-terminal-protocol.ts#L39)
wdyt?

Regarding `this.processGone`  I wonder if we have another indicator to avoid using boolean flag. I'll check and let you know.

RomanNikitenko: @AndrienkoAleksandr 
I see that you defined `this.processGone = true;` for:
- `onTerminalExecExit`
- `onTerminalExecError`
cases.

For these cases do we still have alive `this.termServer`? Can `this.termServer` be used like this `this.termServer!.check({ id: id }` for example?

AndrienkoAleksandr: > For these cases do we still have alive this.termServer?

Yes, we do.

AndrienkoAleksandr: But, I applied boolean flag, because I don't wan't to make request per each resize. Resize operation is expansive and we could affect performance if we will do to much requests.

RomanNikitenko: You are right about performance, I asked because I was looking for a solution without boolean flag.

At the moment we have [isOpen](https://github.com/eclipse/che-theia/blob/fc050e522870fad58802ae6f2ec75925c5c45d97/extensions/eclipse-che-theia-terminal/src/browser/terminal-widget/remote-terminal-widget.ts#L64) flag.  
It's switched to `true` when a socket [is open](https://github.com/eclipse/che-theia/blob/fc050e522870fad58802ae6f2ec75925c5c45d97/extensions/eclipse-che-theia-terminal/src/browser/terminal-widget/remote-terminal-widget.ts#L207), but is **NOT**  switched to `false` when:
-  [onError](https://github.com/eclipse/che-theia/blob/fc050e522870fad58802ae6f2ec75925c5c45d97/extensions/eclipse-che-theia-terminal/src/browser/terminal-widget/remote-terminal-widget.ts#L212) event
- [onClose](https://github.com/eclipse/che-theia/blob/fc050e522870fad58802ae6f2ec75925c5c45d97/extensions/eclipse-che-theia-terminal/src/browser/terminal-widget/remote-terminal-widget.ts#L219) event 
- [at disposing](https://github.com/eclipse/che-theia/blob/fc050e522870fad58802ae6f2ec75925c5c45d97/extensions/eclipse-che-theia-terminal/src/browser/terminal-widget/remote-terminal-widget.ts#L204) for example.

If I understand correctly `this.closeOutputConnectionDisposable.dispose()` is executed for `onTerminalExecExit` and `onTerminalExecError` events, so next a socket will be closed.
If so, can we:
- use `isOpen` flag instead the new flag 
or
- have `socket` as field and do `this.socket = undefined` when a socket is closed. For this case maybe we could remove both boolean flags and use check if socket exist.

I don't know this part of code very well, so sorry if I generate bad ideas :-)
I don't have a goal to avoid using boolean flags at all costs and it's OK to me merge the PR as is if we don't have a better solution at the moment.

Thanks in advance!

AndrienkoAleksandr: Hello @RomanNikitenko, thanks a lot for feedback. I removed flag `isOpen` it was really confused flag with bad name. About this.socket - I think we should not use it to determine process alive or it is gone, because this socket is responsible to send/receive data to the exec, but it is not responsible to make resize.

RomanNikitenko:
@AndrienkoAleksandr 
btw, there is some problem with ci docker build, please take a look.
maybe restart will help...

AndrienkoAleksandr:
crw-ci-test

AndrienkoAleksandr:
crw-ci-build

PR #737, opened 2020-05-05
Make it possible to build Che Theia based on upstream Theia commit

azatsarynnyy:

### What does this PR do?
With the proposed changes it's possible to build Che Theia based on a particular commit of Theia.

By default, the build works without any changes - it takes the latest commit of Theia in master branch.
`THEIA_COMMIT_SHA` argument can be passed through the [build.sh](https://github.com/eclipse/che-theia/blob/master/dockerfiles/theia/build.sh) `build-args`.

Also, it's not needed anymore to set Theia version to `THEIA_VERSION` file before doing a release. It will be fetched from npmjs.org.

closes https://github.com/eclipse/che/issues/16691
closes https://github.com/eclipse/che/issues/16654


benoitf:
I'm still wondering about THEIA_COMMIT_SHA, how does it change the yarn.lock dependencies of plug-ins, etc to use that version as well ? looks like it will still try to compile extensions/plugins with yarn.lock settings
it seems it's only managed by release script, not the build script

azatsarynnyy:
`THEIA_COMMIT_SHA` doesn't affect node build of the extensions/plugins.
It affects only upstream Theia build with our extensions/plugins included.

PR #739, opened 2020-05-06
15540 user level tasks

tsmaeder:
### What does this PR do?
This PR fixes:
* Support Che tasks for multi-root workspaces https://github.com/eclipse/che/issues/15540
* Task: Changing a command in the devfile of an existing workspace is not taken in account https://github.com/eclipse/che/issues/13994

The approach is to not replicate tasks to tasks.json but instead contributing them to the "tasks" key of the user preferences.

Update: switched to replicating the tasks, but to the user level.

### DIFF
--- a/plugins/containers-plugin/src/containers-tree-data-provider.ts
+++ b/plugins/containers-plugin/src/containers-tree-data-provider.ts
@@ -10,6 +10,7 @@
 
 import * as theia from '@theia/plugin';
 import { IContainer } from './containers-service';
+import { TaskScope } from '@eclipse-che/plugin';
 
 interface ITreeNodeItem {
     id: string;
@@ -120,7 +121,7 @@ export class ContainersTreeDataProvider implements theia.TreeDataProvider<ITreeN
                         iconPath: 'fa-cogs medium-yellow',
                         command: {
                             id: CONTAINERS_PLUGIN_RUN_TASK_COMMAND_ID,
-                            arguments: [this.getRootUri().toString(), command.commandName, container.name]
+                            arguments: [command.commandName, container.name]
                         }
                     });
                 });
@@ -235,14 +236,6 @@ export class ContainersTreeDataProvider implements theia.TreeDataProvider<ITreeN
         this.onDidChangeTreeDataEmitter.fire();
     }
 
-    private getRootUri(): theia.Uri {
-        const workspaceFolders = theia.workspace.workspaceFolders;
-        if (!workspaceFolders || workspaceFolders.length < 1) {
-            return theia.Uri.file('/projects');
-        }
-        return workspaceFolders[0].uri;
-    }
-
     private getRandId(): string {
         let uniqueId = '';
         for (let counter = 0; counter < 1000; counter++) {
@@ -295,10 +288,10 @@ export const CONTAINERS_PLUGIN_RUN_TASK_COMMAND_ID = 'containers-plugin-run-task
  * This is needed if a command doesn't have container to be run in specified.
  * In such case we run the command in the container under which this command was clicked.
  */
-export async function containersTreeTaskLauncherCommandHandler(source: string, label: string, containerName: string): Promise<void> {
+export async function containersTreeTaskLauncherCommandHandler(label: string, containerName: string): Promise<void> {
     const tasks: theia.Task[] = await theia.tasks.fetchTasks({ type: 'che' });
     for (const task of tasks) {
-        if (task.name === label && task.source === source) {
+        if (task.name === label) { // task labels are unique in the workspace
             if (!task.definition.target) {
                 task.definition.target = {};
             }
@@ -310,5 +303,5 @@ export async function containersTreeTaskLauncherCommandHandler(source: string, l
     }
 
     // Shouldn't happen. Fallback to default behaviour.
-    theia.commands.executeCommand('task:run', source, label);
+    theia.commands.executeCommand('task:run', 'che', label, TaskScope.Global);
 }
--- a/plugins/task-plugin/src/export/export-configs-manager.ts
+++ b/plugins/task-plugin/src/export/export-configs-manager.ts
@@ -10,7 +10,6 @@
 
 import { injectable, inject, multiInject } from 'inversify';
 import { CheWorkspaceClient } from '../che-workspace-client';
-import * as theia from '@theia/plugin';
 import { che as cheApi } from '@eclipse-che/api';
 
 export const ConfigurationsExporter = Symbol('ConfigurationsExporter');
@@ -23,7 +22,7 @@ export interface ConfigurationsExporter {
      * @param workspaceFolder workspace folder for exporting configs in the config file
      * @param commands commands with configurations for export
      */
-    export(workspaceFolder: theia.WorkspaceFolder, commands: cheApi.workspace.Command[]): Promise<void>;
+    export(commands: cheApi.workspace.Command[]): Promise<void>;
 }
 /** Contains configurations as array of object and as raw content and is used at getting configurations from config file for example */
 export interface Configurations<T> {
@@ -46,25 +45,16 @@ export class ExportConfigurationsManager {
     protected readonly exporters: ConfigurationsExporter[];
 
     async export(): Promise<void> {
-        const workspaceFolders = theia.workspace.workspaceFolders;
-        if (!workspaceFolders || workspaceFolders.length < 1) {
-            return;
-        }
-
         const exportPromises = [];
         const cheCommands = await this.cheWorkspaceClient.getCommands();
         for (const exporter of this.exporters) {
-            exportPromises.push(this.doExport(workspaceFolders, cheCommands, exporter));
+            exportPromises.push(this.doExport(cheCommands, exporter));
         }
 
         await Promise.all(exportPromises);
     }
 
-    private async doExport(workspaceFolders: theia.WorkspaceFolder[], cheCommands: cheApi.workspace.Command[], exporter: ConfigurationsExporter): Promise<void> {
-        const exportConfigsPromises = [];
-        for (const workspaceFolder of workspaceFolders) {
-            exportConfigsPromises.push(exporter.export(workspaceFolder, cheCommands));
-        }
-        await Promise.all(exportConfigsPromises);
+    private async doExport(cheCommands: cheApi.workspace.Command[], exporter: ConfigurationsExporter): Promise<void> {
+        return exporter.export(cheCommands);
     }
 }
--- a/plugins/task-plugin/src/export/launch-configs-exporter.ts
+++ b/plugins/task-plugin/src/export/launch-configs-exporter.ts
@@ -31,7 +31,20 @@ export class LaunchConfigurationsExporter implements ConfigurationsExporter {
     @inject(VsCodeLaunchConfigsExtractor)
     protected readonly vsCodeLaunchConfigsExtractor: VsCodeLaunchConfigsExtractor;
 
-    async export(workspaceFolder: theia.WorkspaceFolder, commands: cheApi.workspace.Command[]): Promise<void> {
+    async export(commands: cheApi.workspace.Command[]): Promise<void> {
+        if (!theia.workspace.workspaceFolders) {
+            return;
+        }
+
+        const exportConfigsPromises: Promise<void>[] = [];
+
+        for (const workspaceFolder of theia.workspace.workspaceFolders) {
+            exportConfigsPromises.push(this.doExport(workspaceFolder, commands));
+        }
+        await Promise.all(exportConfigsPromises);
+    }
+
+    async doExport(workspaceFolder: theia.WorkspaceFolder, commands: cheApi.workspace.Command[]): Promise<void> {
         const launchConfigFileUri = this.getConfigFileUri(workspaceFolder.uri.path);
         const configFileConfigs = this.configFileLaunchConfigsExtractor.extract(launchConfigFileUri);
         const vsCodeConfigs = this.vsCodeLaunchConfigsExtractor.extract(commands);
@@ -67,7 +80,6 @@ export class LaunchConfigurationsExporter implements ConfigurationsExporter {
 
             conflictHandler(conflict, config2);
         }
-
         return result;
     }
 
--- a/plugins/task-plugin/src/export/task-configs-exporter.ts
+++ b/plugins/task-plugin/src/export/task-configs-exporter.ts
@@ -9,7 +9,6 @@
  **********************************************************************/
 
 import { injectable, inject } from 'inversify';
-import * as theia from '@theia/plugin';
 import * as startPoint from '../task-plugin-backend';
 import { che as cheApi } from '@eclipse-che/api';
 import { TaskConfiguration } from '@eclipse-che/plugin';
@@ -20,12 +19,17 @@ import { VsCodeTaskConfigsExtractor } from '../extract/vscode-task-configs-extra
 import { ConfigurationsExporter } from './export-configs-manager';
 import { ConfigFileTasksExtractor } from '../extract/config-file-task-configs-extractor';
 import { BackwardCompatibilityResolver } from '../task/backward-compatibility';
+import { homedir } from 'os';
 
 const CONFIG_DIR = '.theia';
 const TASK_CONFIG_FILE = 'tasks.json';
 const formattingOptions = { tabSize: 4, insertSpaces: true, eol: '' };
 
 export const VSCODE_TASK_TYPE = 'vscode-task';
+
+// this really should be handled through some theia API, but there is none available in plugins
+// it only works on linux and only if this plugin runs in the theia container
+export const THEIA_USER_TASKS_PATH = resolve(homedir(), CONFIG_DIR, TASK_CONFIG_FILE);
 
 /** Exports configurations of tasks in the config file. */
 @injectable()
@@ -43,9 +47,8 @@ export class TaskConfigurationsExporter implements ConfigurationsExporter {
     @inject(BackwardCompatibilityResolver)
     protected readonly backwardCompatibilityResolver: BackwardCompatibilityResolver;
 
-    async export(workspaceFolder: theia.WorkspaceFolder, commands: cheApi.workspace.Command[]): Promise<void> {
-        const tasksConfigFileUri = this.getConfigFileUri(workspaceFolder.uri.path);
-        const configFileTasks = this.configFileTasksExtractor.extract(tasksConfigFileUri);
+    async export(commands: cheApi.workspace.Command[]): Promise<void> {
+        const configFileTasks = this.configFileTasksExtractor.extract(THEIA_USER_TASKS_PATH);
 
         const cheTasks = this.cheTaskConfigsExtractor.extract(commands);
         const vsCodeTasks = this.vsCodeTaskConfigsExtractor.extract(commands);
@@ -54,18 +57,18 @@ export class TaskConfigurationsExporter implements ConfigurationsExporter {
 
         const configFileContent = configFileTasks.content;
         if (configFileContent) {
-            this.saveConfigs(tasksConfigFileUri, configFileContent, this.merge(configFileConfigs, devfileConfigs, this.getConsoleConflictLogger()));
+            this.saveConfigs(THEIA_USER_TASKS_PATH, configFileContent, this.merge(configFileConfigs, devfileConfigs, this.getConsoleConflictLogger()));
             return;
         }
 
         const vsCodeTasksContent = vsCodeTasks.content;
         if (vsCodeTasksContent) {
-            this.saveConfigs(tasksConfigFileUri, vsCodeTasksContent, devfileConfigs);
+            this.saveConfigs(THEIA_USER_TASKS_PATH, vsCodeTasksContent, devfileConfigs);
             return;
         }
 
         if (cheTasks) {
-            this.saveConfigs(tasksConfigFileUri, '', cheTasks);
+            this.saveConfigs(THEIA_USER_TASKS_PATH, '', cheTasks);
         }
     }
 
@@ -101,10 +104,6 @@ export class TaskConfigurationsExporter implements ConfigurationsExporter {
         }
 
         return JSON.stringify(properties1) === JSON.stringify(properties2);
-    }
-
-    private getConfigFileUri(rootDir: string): string {
-        return resolve(rootDir.toString(), CONFIG_DIR, TASK_CONFIG_FILE);
     }
 
     private saveConfigs(tasksConfigFileUri: string, content: string, configurations: TaskConfiguration[]): void {
--- a/plugins/workspace-plugin/src/theia-commands.ts
+++ b/plugins/workspace-plugin/src/theia-commands.ts
@@ -16,6 +16,7 @@ import { che as cheApi } from '@eclipse-che/api';
 import * as fileuri from './file-uri';
 import { execute } from './exec';
 import * as git from './git';
+import { TaskScope } from '@eclipse-che/plugin';
 
 const SS_CRT_PATH = '/tmp/che/secret/ca.crt';
 const CHE_TASK_TYPE = 'che';
@@ -265,7 +266,7 @@ export class TheiaCommand {
 
         if (this.id === ActionId.RUN_COMMAND) {
             if (this.properties) {
-                return theia.commands.executeCommand('task:run', CHE_TASK_TYPE, this.properties.name)
+                return theia.commands.executeCommand('task:run', CHE_TASK_TYPE, this.properties.name, TaskScope.Global)
                     .then(() => {
                         theia.window.showInformationMessage('Executed che command succesfully');
                     }, e => {


tsmaeder:
crw-ci-test

RomanNikitenko:
@tsmaeder 
I would like to test your changes, but I found that your fork and branch is outdated
outdated

Could you align your branch with master?

RomanNikitenko:
@tsmaeder 
The compilation errors are fixed for me and finally, I got the images with your changes. 
Could you explain what's the true way now to edit a configuration for `che` task?

tsmaeder:
> @tsmaeder
> Could you explain what's the true way now to edit a configuration for `che` task?

If it's defined in the devfile, you edit it in the devfile, if it's defined in a tasks.json, you edit it in that tasks.json.

tsmaeder:
> @tsmaeder
> For me `che` tasks are always in running state with your changes.

Can you please provide the configuration of the task you're using? I can't easily copy it from the animated gif.

RomanNikitenko:
@tsmaeder 
> Can you please provide the configuration of the task you're using? I can't easily copy it from the animated gif.

I defined in my devfile:
 ```
commands:
  - name: 'test:che:task'
    actions:
      - workdir: /projects/che-theia
        type: exec
        command: |
          echo test che task
        component: che-dev
```

**Comment by RomanNikitenko** [comment]
@tsmaeder 
> If it's defined in the devfile, you edit it in the devfile

so, to edit a `che` task user should:
- go to dashboard
- stop workspace
- edit devfile
- start workspace

Correct?
I'm asking to understand the new behavior and continue test your changes

**Comment by tsmaeder** [comment]
@RomanNikitenko correct. The long term "evil plan" is to listen to changes in the devfile (https://github.com/eclipse/che/issues/16859) and to send "preference changed" when the devfile is changed. We also could represent the relevant sections of the devfile as a non-file resource inside Theia and write updates back to the devfile via the Che API.

**Comment by tsmaeder** [comment]
@RomanNikitenko I added your task to a workspace based on the PR and the case works for me:

https://youtu.be/GFLInRwpG1Q

**Comment by RomanNikitenko** [comment]
@tsmaeder 
>  I added your task to a workspace based on the PR and the case works for me

thanks
But on your video I don't see that the icon for the corresponding tab was changed when the task was completed...

I'm testing your changes on che.openshift.io, so maybe some delays on this infrastructure...
Could you provide the corresponding section for the devfile with your changes?
I could try it on che.openshift.io...

**Comment by tsmaeder** [comment]
Here's what I'm using:
```
metadata:
  name: petclinic-dev-environment
projects:
  - name: petclinic
    source:
      location: 'https://github.com/spring-projects/spring-petclinic.git'
      type: git
      branch: master
components:
  - selector:
      app.kubernetes.io/component: webapp
    type: kubernetes
    entrypoints:
      - command:
          - tail
        args:
          - '-f'
          - /dev/null
        containerName: spring-boot
    reference: >-
      https://raw.githubusercontent.com/eclipse/che/master/tests/e2e/files/happy-path/containers-happy-path.yaml
    alias: petclinic-web
  - selector:
      app.kubernetes.io/component: database
    type: kubernetes
    reference: >-
      https://raw.githubusercontent.com/eclipse/che/master/tests/e2e/files/happy-path/containers-happy-path.yaml
    alias: petclinic-db
  - mountSources: true
    memoryLimit: 3Gi
    type: dockerimage
    alias: maven-container
    image: 'quay.io/eclipse/happy-path:nightly'
    env:
      - value: /home/user/.m2
        name: MAVEN_CONFIG
  - id: redhat/java/latest
    type: chePlugin
  - id: redhat/vscode-yaml/latest
    type: chePlugin
  - type: cheEditor
    reference: >-
      https://raw.githubusercontent.com/chepullreq4/pr-check-files/master/che-theia/pr-739/simple/che-theia-editor.yaml
apiVersion: 1.0.0
commands:
  - name: build
    actions:
      - workdir: /projects/petclinic
        type: exec
        command: >-
          mvn clean package | tee build.txt && tail -n 40
          /projects/petclinic/build.txt | grep 'BUILD SUCCESS' >
          /projects/petclinic/result-build.txt
        component: maven-container
  - name: build-file-output
    actions:
      - type: exec
        command: >-
          cd /projects/petclinic && mvn package | tee build-output.txt && tail
          -n 40 /projects/petclinic/build-output.txt | grep 'BUILD SUCCESS' >
          /projects/petclinic/result-build-output.txt
        component: maven-container
  - name: run
    actions:
      - workdir: /projects/petclinic/target
        type: exec
        command: >-
          java -jar spring-petclinic-2.3.0.BUILD-SNAPSHOT.jar
          --spring.profiles.active=mysql
        component: maven-container
  - name: run-with-changes
    actions:
      - workdir: /projects/petclinic/target
        type: exec
        command: >-
          java -jar spring-petclinic-2.3.0.BUILD-SNAPSHOT.jar
          --spring.profiles.active=mysql
        component: maven-container
  - name: run-debug
    actions:
      - workdir: /projects/petclinic/target
        type: exec
        command: >-
          java -jar -Xdebug
          -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=5005
          spring-petclinic-2.3.0.BUILD-SNAPSHOT.jar
          --spring.profiles.active=mysql
        component: maven-container
  - name: Debug remote java application
    actions:
      - referenceContent: |
          {
          "version": "0.2.0",
          "configurations": [
            {
              "type": "java",
              "name": "Debug (Attach) - Remote",
              "request": "attach",
              "hostName": "localhost",
              "port": 5005
            }]
          }
        type: vscode-launch
  - name: 'test:che:task'
    actions:
      - workdir: /projects/che-theia
        type: exec
        command: |
          echo test che task
        component: maven-container
```

**Comment by RomanNikitenko** [comment]
@tsmaeder 
I tried your `cheEditor` component on che.openshift.io and had the same experience - the task didn't look as terminated, also I didn't see that the icon was changed when the task was completed.

running_user_lavel_bug

I have another question: why the task system doesn't remember user's choice at running a `che` task? AFAIK it works for `che` tasks for `master` branch...

users_choice

**Comment by tsmaeder** [comment]
@RomanNikitenko the only thing I can reproduce is that ctrl-c is not working. Let's do a BJ on monday so we can determine what you and I are doing differently.

**Comment by RomanNikitenko** [comment]
@tsmaeder 
Could you try your changes on che.openshift.io before our BJ session?
All my video I recorded on che.openshift.io, maybe that's why we have the different behavior...

**Comment by tsmaeder** [comment]
@RomanNikitenko I tried with your devfile both on crc and openshift.io and tasks terminate fine for me. The only differences I can think of is that I use Chromium, not Chrome, and that I'm on Windows, not Linux 🤷

**Comment by RomanNikitenko** [comment]
@tsmaeder
OK, I hope it works well for other people.

Could you check that a status is changed when a `che` task is completed.
I recorded how it works for master branch
status_icons_master

File: extensions/eclipse-che-theia-preferences-provider-extension/src/browser/async-preference-provider.ts [hunk 1]
```diff
@@ -0,0 +1,100 @@
+/********************************************************************************
+ * Copyright (C) 2018 Ericsson and others.
```
RomanNikitenko, line 2: Does it mean that `AsyncPreferenceProvider` things were copied from upstream?
Should we have a CQ then?

extensions/eclipse-che-theia-plugin-ext/src/common/converter.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+import * as che from '@eclipse-che/plugin';
+import * as theia from '@theia/task/lib/common';
+
```
RomanNikitenko, line 18: If I understand correctly we have these functions because `scope`field only in task config.
Can we align the corresponding object to upstream and avoid adding/using these functions?

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -271,7 +271,7 @@ const uploadPrivateKey = async (sshkeyManager: SshKeyManager) => {
     }
 
     unlinkSync(privateKeyPath.path);
-    rmdirSync(tempDir, { recursive: true });
+    removeSync(tempDir);
```
RomanNikitenko, line 274: Are changes for `ssh-plugin` included by mistake in this PR?

extensions/eclipse-che-theia-preferences-provider-extension/src/browser/che-tasks-preferences.ts
```diff
@@ -0,0 +1,138 @@
...
+        return this.wrappedProvider.getConfigUri();
+    }
+
+    getDomain(): string[] | undefined {
+        return this.wrappedProvider.getDomain();
+    }
+}
+
+function fromVSCodeTaskCommand(command: cheApi.workspace.Command, handler: (task: TaskCustomization) => void): void {
+    if (command && command.attributes && command.attributes.actionReferenceContent) {
```
RomanNikitenko, line 82: Sorry, I don't understand why `TaskCustomization` is used here. 
Can we use `TaskConfiguration` instead?

extensions/eclipse-che-theia-preferences-provider-extension/src/browser/async-preference-provider.ts
tsmaeder: No, for two reasons:
1. I just copied the header, the rest is at most "inspiried by"
2. We're already shipping all of theia with our product, so we already have approval to use that code.

plugins/ssh-plugin/src/ssh-plugin-backend.ts
tsmaeder: No, otherwise, the code does not compile

extensions/eclipse-che-theia-preferences-provider-extension/src/browser/che-tasks-preferences.ts
tsmaeder: Because TaskConfiguration have a mandatory scope, which vs code task customizations do not have.

tsmaeder: Generally, I think we should get rid of TaskConfiguration when talking about non-contributed things. I am not aware of an instance where it the differentiation makes sense.

plugins/ssh-plugin/src/ssh-plugin-backend.ts
tsmaeder: Or it didn't until Florent merge his recent PR.

extensions/eclipse-che-theia-plugin-ext/src/common/converter.ts
tsmaeder: No, I don't think so: I'd have to start modifying the execution part of the tasks plugin, which I don't want to do at this time. Also, using theia extension types for our main/ext interface is not a good thing, IMO: main/ext usually uses their own types. This is more work, but makes the code less susceptible to changes in Theia. Otherwise, we'd have to use the same types everywhere (including the plugin side), which would break layering/modularity.

tsmaeder:
The thing you noticed with the icon on terminals not changing to a checkmark upon task completion is also present in Theia and in commits prior to the one introducing the user level tasks. So I'm concluding this has nothing to do with this PR. It's as broken as before.

extensions/eclipse-che-theia-preferences-provider-extension/src/browser/che-tasks-preferences.ts
RomanNikitenko: You can see how `TaskCustomization` is used on `theia` side.
For example, `task-service` [has](https://github.com/eclipse-theia/theia/blob/bb43e9ea7508687932e386a344f2afc0e915e451/packages/task/src/browser/task-service.ts#L885):
 `getTaskCustomization(task: TaskConfiguration): Promise<TaskCustomization>`

The method is used to get `TaskCustomization` based on given `TaskConfiguration`:
- at running a task we have `TaskConfiguration` for running and [get](https://github.com/eclipse-theia/theia/blob/bb43e9ea7508687932e386a344f2afc0e915e451/packages/task/src/browser/task-service.ts#L488)  `customizationObject` to resolve `problemMatchers` for example

- [it's impossible](https://github.com/eclipse-theia/theia/blob/bb43e9ea7508687932e386a344f2afc0e915e451/packages/task/src/browser/task-configurations.ts#L250) to get `TaskCustomization` if it's **NOT** `detected` task

So I don't understand how we can consider a `che` task as `TaskCustomization`. Customization for what?

RomanNikitenko:
> The thing you noticed with the icon on terminals not changing to a checkmark upon task completion is also present in Theia and in commits prior to the one introducing the user level tasks. So I'm concluding this has nothing to do with this PR. It's as broken as before.

@tsmaeder 
I checked it for `che-theia` `master` branch for **che** tasks - it works for me.
Does it work for you for `che` tasks for `master` branch of `che-theia`?

extensions/eclipse-che-theia-preferences-provider-extension/src/browser/che-task-schema.ts
```diff
@@ -63,3 +67,14 @@ export const CHE_TASK_SCHEMA: che.TaskJSONSchema = {
     },
     additionalProperties: true
 };
+
+@injectable()
+export class SchemaContribution implements FrontendApplicationContribution {
```
vzhukovskii, line 72: It would be better to move this class outside this file. To `che-task-schema-contribution.ts` or smth like that. Cause it's not obvious why an extra class exists in schema description.

tsmaeder:
> I checked it for che-theia master branch for che tasks

The commit used in che-theia master is something like two weeks old. However in the commit immediately preceding my change in Theia, it already did not work. So it was broken somewhere between those two. Still, nothing this PR introduced.

RomanNikitenko:
At the moment user has ability to apply `Configure task` action for any configuration from the list, then `tasks.json` is open and the corresponding configuration is available for editing. 
configure_master

I tested this behavior for current PR and can see that empty `tasks.json` file is open - the corresponding configuration is not available for editing.
configure_task_action

I understand that according to the new behavior user should go to dashboard, edit the devfile and restart a wroskpace, but I think it can be really confusing for user to see just empty `tasks.json` file after applying `Configure task` action - looks like a bug...

Do you have an idea how we can improve this place?
Maybe some notification with the corresponding info what user should do to edit a `che` task.

extensions/eclipse-che-theia-preferences-provider-extension/src/browser/che-tasks-preferences.ts
tsmaeder: Yes, but TaskCustomization and TaskConfiguration are interfaces: they are simply promises that an object has certain properties of certain types: so every TaskConfiguration is also a TaskCustomization. So the object we're returning here does not have the mandatory "scope" of the appropriate type. So it does not conform to the TaskConfiguration interface.

> it's impossible to get TaskCustomization if it's NOT detected task

I don't understand what you're trying to say here: could you explain differently?

In general, I don't think it ever makes sense to talk about a TaskConfiguration for anything that is not a detected task.

tsmaeder:
> At the moment user has ability to apply Configure task action for any configuration from the list, then tasks.json is open and the corresponding configuration is available for editing.

Ok, that is a real problem: I simply did not expect the case where we need to configure the tasks outside of what is available in the devfile: the problem matchers are purely a Theia concept and therefore not available for Che commands.

extensions/eclipse-che-theia-preferences-provider-extension/src/browser/che-tasks-preferences.ts
RomanNikitenko: Do you consider `che` task as `detected` or as `configured` task?

RomanNikitenko:
@tsmaeder 
> the problem matchers are purely a Theia concept and therefore not available for Che commands

I think it's VS Code concept.

For `che` commands at the moment it's possible to select `Never scan the task output` once and task system remember user's choice(please see my video [in the comment](https://github.com/eclipse/che-theia/pull/739#issuecomment-629262613)) and doesn't ask about it every time when user wants to run the task.

btw: `Configure task` action is **NOT** ability to define `problemMatcher` for a task. In VS Code/Theia it's ability to get `customization` for a task, to `override` some fields, to add some fields for `base` configuration which was provided by the corresponding provider from plugin side. You can try to add `presentation` field for `VS Code` task and manage how output is displayed for the current task: which terminal is used and so on.

But my question is more general:
The behavior which I described [above](https://github.com/eclipse/che-theia/pull/739#issuecomment-630206856) is definitely a bug on `theia` side.
Let's assume it's fixed. So, after applying `Configure task` action a configuration for `che` task in `tasks.json` on user level and user edits it. 
Next we have to sync it in `devfile`?
or?
what's the plan?

extensions/eclipse-che-theia-preferences-provider-extension/src/browser/che-tasks-preferences.ts
tsmaeder: The way it is implemented now they are configured tasks (aka TaskCustomizations). Maybe that _is_ the wrong approach and we should present them as detected tasks. At this point, we could call "configure" on the che tasks and would be able to set stuff like "problemMatcher" in the user-level "tasks.json". What do you think.

tsmaeder:
@RomanNikitenko it seems the "checkbox" on the terminal tab is a che-terminal feature only? In current nightly, it only works for che tasks, not shell tasks, for example. Can you confirm that?

RomanNikitenko:
@tsmaeder 
> it seems the "checkbox" on the terminal tab is a che-terminal feature only? In current nightly, it only works for che tasks, not shell tasks, for example.

I mentioned in [this comment](https://github.com/eclipse/che-theia/pull/739#issuecomment-630113538) and [this one](https://github.com/eclipse/che-theia/pull/739#issuecomment-630182043) that status icons work for `che` tasks for `master` branch of `che-theia`

About `theia`/`vs code` tasks - it was broken and we have an issue for that https://github.com/eclipse/che/issues/16672

RomanNikitenko:
@tsmaeder @ericwill @azatsarynnyy 
About our meeting yesterday.

I asked what happens if user clicks on `gear wheel` icon in `VS Code` - so applying `Configure Task` action for a `User` level task **in `VS Code`**.
Today I tested it a little and can see that:
- `tasks.json` file on `User` level is open with the corresponding configurations when user clicks on `gear wheel` icon
- user is able to edit any configuration
- task system remembers user's choice for `Never scan the task output`, I guess it's possible to define any problem matcher or other fields

user_level_vs_code

**Why I raise this question at the moment?**

If I understand correctly, the idea was to provide `che` tasks in `read` mode on `editor` side, but have ability to edit them from `dashboard` and have `devFile` as one source for editing.
But from my testing of VS Code behavior I see that it's possible to edit `User` level task and at the moment we just have a bug for `Theia` which will be fixed sooner or later. That's why I prefer to think about it right now - if it's fixed - what our next step?

@tsmaeder 
please correct me if I'm wrong - I don't have an experience related to `User` level tasks in `VS Code`.

RomanNikitenko:
@tsmaeder @ericwill @azatsarynnyy
About our meeting yesterday.

Another topic which we discussed is Tasks and Workspace roots.
I tested current behavior for **master** branch and can see that `che` tasks at the moment are `Global` tasks.
What I mean: I started a workspace for `che-in-che` development, it contains two projects (theia and che-theia) and commands to build these projects.

After starting the workspace `/projects` is a workspace folder which contains both projects, 
I can:
- run the task to build `theia` project
- run the task to build `che-theia` project

Then I open only `theia` project as workspace folder - all `che` commands are still available for running: `theia:build` and `che-theia:build`.

Then I open only `che-theia` project as workspace folder - the same behavior - both `theia:build` and `che-theia:build` are still available for running.
  
I know that for some projects it's important to have a different workspace root (not `/projects`). 
But taking into account my testing above at the moment I see more problems from `devFile` side than from task system side. 
Let's assume that it's possible on `devFile` level to define `/che-theia` as workspace folder - it will do automatically the same that I did manually in my test above - it will open `/che-theia` when the workspace is started and then all `che` tasks are available for running.

What I don't understand in our strategy - why `theia:build` and `che-theia:build` tasks are global.
One of them should belongs to `theia` project and another one to `che-theia` project. At the same time user can have global task `build` like:
```
        {
            "type": "che",
            "label": "build",
            "command": "yarn\n",
            "target": {
                "workingDir": "${workspaceFolder}",
                "component": "che-dev"
            },
            "problemMatcher": []
        },
``` 
I don't want prevent or change something in the current PR behavior, but the goal is to clarify (at least for myself) some things in our strategy, in our vision. 
I think User level / Global tasks are cool feature and it gives us more opportunities for tasks system. Maybe we should rise this topic again when the PR is merged and we get ability to define `che` tasks as `User level` tasks.

tsmaeder:
crw-ci-test

RomanNikitenko:
@tsmaeder 
I tested your changes after update and can see some issues.

1. Running a `che` task from `Workspace` panel is not working
run_workspace_panel

2. Resolving variables for `command` field of `che` tasks is not working
3. Resolving variables for `workingDir` field of `che` tasks is not working

**Expected behavior (master branch):**
expected_resolve_workDir

**Actual behavior (current branch):**
resolving_che_actual

4. `Che` tasks are not displayed in the list for `Configure Tasks` action
configure_che_master

5.  Task system doesn't remember user's choice `Never scan the task output`
user_choice_never

RomanNikitenko:
I used the following configurations for testing variables:
```
      {
            "type": "che",
            "label": "test resolving Che command",
            "command": "echo ${file}",
            "target": {
                "workingDir": "${CHE_PROJECTS_ROOT}/console-java-simple",
                "component": "maven"
            }
        },
        {
            "type": "che",
            "label": "test resolving Che workingDir",
            "command": "ls",
            "target": {
                "workingDir": "${workspaceFolder}/console-java-simple",
                "component": "maven"
            }
        },
```

tsmaeder:
I've filed https://github.com/eclipse-theia/theia/issues/7916

tsmaeder:
@RomanNikitenko 


> Resolving variables for command field of che tasks is not working
Resolving of CHE_PROJECTS_DIR works just fine for me. Are you sure this is not a problem with the terminal? Could you try `ls -la` instead?

> Resolving variables for workingDir field of che tasks is not working

You have a file outside of the workspace open (~/.theia/tasks.json). In this state, `workspaceFolder` resolves to nothing. This is the intended behavior. If we wanted to change this behaviour, we'd have to fix the implementation of the "workspaceFolder" variable. If you open a file from the workspace, the resolution works as you intended.

tsmaeder:
For 5) I've filed https://github.com/eclipse-theia/theia/issues/7927 in Theia

RomanNikitenko:
@tsmaeder 


> @RomanNikitenko
> 
> > Resolving variables for command field of che tasks is not working
> 
> Resolving of CHE_PROJECTS_DIR works just fine for me. Are you sure this is not a problem with the terminal? Could you try `ls -la` instead?

Why did you quote  `Resolving variables for command field of che tasks is not working` (please pay attention **command field**), but your answer about  **CHE_PROJECTS_DIR**? 
The configuration for testing contains:
- label: test resolving Che **command**
- command with variable: **${file}**

> You have a file outside of the workspace open (~/.theia/tasks.json). In this state, workspaceFolder resolves to nothing. This is the intended behavior. 

Do you mean that the cause of issue with resolving variable ${workspaceFolder} is - I just had  opened `~/.theia/tasks.json` file? 
But I had opened `tasks.json` file for master branch as well on my video. 
I didn't investigate it, but looks strange, that ${workspaceFolder} can not be resolved if `tasks.json` file is open.

RomanNikitenko:
@tsmaeder 
You are right - I can reproduce the bug related to variables resolving only when I have open `tasks.json` file on user level. It works well if I have open `tasks.json` file on workspace level. 

I tested it for your branch and for `theia` as well. At least the following variables are affected:
- ${workspaceFolder}
- ${cwd}
- ${file}

I don't think it's expected behavior - it works well for VS Code, so it's a bug on `Theia` side, I created the corresponding issue. Thank you for pointing the root cause of the issue!

tsmaeder:
@RomanNikitenko so what do we need for you to give me a +1 on this PR?

RomanNikitenko:
> @RomanNikitenko so what do we need for you to give me a +1 on this PR?

@tsmaeder 
I didn't rejected your changes, so you can merge this PR any time when you feel it's good enough.

The goal of my comments here is: detect and provide you info about regressions using my knowledge about task system.  

I reviewed and tested your last changes in upstream, so I think 2 problems from [this comment](https://github.com/eclipse/che-theia/pull/739#issuecomment-634551814) should be fixed now.
I could check it for `che` tasks on `che-theia` side, but the corresponding images are outdated at the moment, so I don't have ability to test it again.

tsmaeder:
> At least 1 approving review is required by reviewers with write access

@RomanNikitenko

RomanNikitenko:
@tsmaeder 
As I mentioned above the corresponding images are outdated at the moment, so I don't have ability to test it again. 

Can you confirm that the following problems were fixed for `che` tasks:
- Che tasks are not displayed in the list for `Configure Tasks` action
- Task system doesn't remember user's choice `Never scan the task output`
 
I guess with these changes we still have the following problems:
- Running a che task from `Workspace` panel is not working
- Resolving variables for `che` tasks is not working at editing a configuration (user has open `tasks.json` file)

The first problem is quite critical from my point of view. 
Can we fix at least `Running a che task from Workspace panel ` within current PR?
Did you investigate the problem?

tsmaeder:
crw-ci-test

nickboldt:
crw-ci-test

tsmaeder:
@RomanNikitenko finally managed to fix the problem with executing from the "My Workspace" view. If you would be so kind to review that last commit?

tsmaeder:
crw-ci-test

RomanNikitenko:
btw: it makes sense to provide some changes to the doc https://www.eclipse.org/che/docs/che-7/defining-custom-commands-for-che-theia/ to inform user that `che` tasks configurations from the devfile are exported to user level `tasks.json` file (instead of `/workspace/.theia` directory)

PR #740, opened 2020-05-06
move COPY to...

nickboldt:
move COPY to {HOME}/eclipse-che-theia-generator into post-env.dockerfile stage; move yarn global add steps into new add-che-theia-generator.dockerfile

Change-Id: I23e440429e9d8819ccbc17e7f7b5e1fc14184094


benoitf:
hi, why should we do that ?

nickboldt:
> we're just COPYing a folder, it's something that should work everywhere

But it doesn't in the ubi-brew version. Maybe the offline flag changes the behaviour somehow?

benoitf:
closing as won't fix

PR #745, opened 2020-05-13
fix(requirements): stick to nodejs 10.0 api in plugin

benoitf:
### What does this PR do?
ssh plugin was using a nodejs 12.x api.
Remove it, ensure that node API is stuck to 10.x and switch from fs to fs-extra to avoid to use sync methods

Change-Id: Ia2a63363e532566e612c0ab3e6c9064f8c655163

### DIFF
--- a/plugins/ssh-plugin/package.json
+++ b/plugins/ssh-plugin/package.json
@@ -32,6 +32,7 @@
         "@eclipse-che/plugin": "0.0.1"
       },
       "devDependencies": {
+        "@types/node": "^10.0.0",
         "@eclipse-che/api": "latest",
         "@eclipse-che/plugin": "0.0.1",
         "@theia/plugin": "next",
--- a/plugins/ssh-plugin/src/ssh-plugin-backend.ts
+++ b/plugins/ssh-plugin/src/ssh-plugin-backend.ts
@@ -13,9 +13,8 @@ import * as che from '@eclipse-che/plugin';
 import { RemoteSshKeyManager, SshKeyManager } from './node/ssh-key-manager';
 import { che as cheApi } from '@eclipse-che/api';
 import { resolve, join } from 'path';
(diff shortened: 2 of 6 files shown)


benoitf:
crw-ci-test

plugins/ssh-plugin/package.json
```diff
@@ -32,6 +32,7 @@
         "@eclipse-che/plugin": "0.0.1"
       },
       "devDependencies": {
+        "@types/node": "^10.17.21",
```
tsmaeder, line 35: Why do we need a different version here?

plugins/ssh-plugin/src/ssh-plugin-backend.ts
```diff
@@ -246,7 +245,7 @@ const uploadPrivateKey = async (sshkeyManager: SshKeyManager) => {
         hostName = `default-${Date.now()}`;
     }
 
-    const tempDir = mkdtempSync(join(os.tmpdir(), 'private-key-'));
+    const tempDir = await mkdtemp(join(os.tmpdir(), 'private-key-'));
```
tsmaeder, line 248: Not a fan of removing the "sync" here. I don't know how to test this change and making something async _can_ lead to concurrency problems.

benoitf: what is the difference between
`await somethingASync` vs `somethingSync` ?

plugins/ssh-plugin/package.json
benoitf: in any case all 10.0.0 resolve to 10.17.21 or later

plugins/ssh-plugin/src/ssh-plugin-backend.ts
benoitf: also I don't see what concurrency problems would be there

benoitf:
crw-ci-test

PR #746, opened 2020-05-14
Support encrypted SSH private keys

vinokurig:
### What does this PR do?
When uploading an encrypted SSH private key, ask a password for the key and register the key by executing ssh-add command. On the SSH plugin start iterate through the registered ssh keys and if a private key is encrypted read the passphrase from the key and register it with the help of ssh-add.

fixes https://github.com/eclipse/che/issues/16517


ericwill:
PR check failed because it can't seem to install `sshpass`.

PR #751, opened 2020-05-20
switch to use ubi8/nodejs-10:latest instead...

nickboldt:
switch to use ubi8/nodejs-10:latest instead of locking to a tag

Intent is to work around OSBS problems we're having right now (unable to resolve tags), and keep the builds always using the latest ubi base images without *ME* having to keep submitting PRs to update it. 

We can go back to tag locking later but for now it's a nuisance / extra maintenance burden.

Note that these files are already outdated: latest nodejs-10 tag is https://access.redhat.com/containers/?architecture=AMD64#/registry.access.redhat.com/ubi8/nodejs-10/images/1-82.1589298623 not 1-81. Kinda proves my point :D 

Change-Id: I922cd0248e2383131437831f00b459565b880964

### DIFF
--- a/dockerfiles/theia-dev/docker/ubi8/builder-image-from.dockerfile
+++ b/dockerfiles/theia-dev/docker/ubi8/builder-image-from.dockerfile
@@ -1,2 +1,1 @@
-# https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8/nodejs-10
-FROM registry.access.redhat.com/ubi8/nodejs-10:1-81
+FROM registry.access.redhat.com/ubi8/nodejs-10:latest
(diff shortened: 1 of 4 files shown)


benoitf:
could you tell more in the description ?
AFAIK we want to have reproducible builds, using latest won't help there

I think that if you drop tags, you should use manifest then, but not use latest as we may have random issues

nickboldt:
Intent is to work around OSBS problems we're having right now, and keep the builds always using the latest ubi base images without ME having to keep submitting PRs to update it :D

benoitf:
yes but as we want reproducible builds, need to use digests (or like tags) but not latest.

nickboldt:
You don't trust node to provide smooth updates? :D I'm shocked! :)

PR #752, opened 2020-05-20
CRW-910 :: new script to automatically keep dockerfiles...

nickboldt:
new script to automatically keep dockerfiles up to date

Change-Id: I284d2c1c48b9c6c6b4812d22dc808e035cbb38c0

### DIFF
--- a/build/scripts/update-from.sh
+++ b/build/scripts/update-from.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+#
+# Copyright (c) 2019-2020 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+# update FROM lines in Dockerfiles to the latest base images
+
+# get script
+if [[ ! -f /tmp/updateBaseImages.sh ]]; then 
+    pushd /tmp >/dev/null || exit
+    curl -sSLO https://raw.githubusercontent.com/redhat-developer/codeready-workspaces/master/product/updateBaseImages.sh
+    chmod +x updateBaseImages.sh
+    popd >/dev/null || exit
+fi
+
+# this will attempt to push the change directly, and if it fails, will then 
+# try to generate a PR using hub if installed, or simply create a PR branch in origin
+now="$(date +%s)"
+PR_BRANCH="pr-master-update-base-images-${now}"
+BRANCHUSED=master
+for d in $(find $(pwd)/dockerfiles -name "*ockerfile" | grep ubi); do 
+    if [[ $(grep access.redhat.com/containers $d) ]]; then 
+        f=${d##*/}
+        dir=${d%/*}
+        /tmp/updateBaseImages.sh -b ${BRANCHUSED} -w ${dir} -maxdepth 1 -f $f -prb "${PR_BRANCH}" --no-push
+    fi
+done
+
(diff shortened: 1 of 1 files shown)


build/scripts/update-from.sh
```diff
@@ -0,0 +1,30 @@
...
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+# update FROM lines in Dockerfiles to the latest base images
+
+# get script
+if [[ ! -f /tmp/updateBaseImages.sh ]]; then 
+    pushd /tmp >/dev/null || exit
+    curl -sSLO https://raw.githubusercontent.com/redhat-developer/codeready-workspaces/master/product/updateBaseImages.sh
```
benoitf, line 16: Hi, FYI Che should not depend on codeready workspaces

Also using master is not nice as it can break later

build/scripts/update-from.sh
```diff
@@ -0,0 +1,30 @@
...
+    chmod +x updateBaseImages.sh
+    popd >/dev/null || exit
+fi
+
+# this will attempt to push the change directly, and if it fails, will then 
+# try to generate a PR using hub if installed, or simply create a PR branch in origin
+now="$(date +%s)"
+for d in $(find $(pwd)/dockerfiles -name "*ockerfile" | grep ubi); do 
+    if [[ $(grep access.redhat.com/containers $d) ]]; then
```
benoitf, line 25: IMHO there is something wrong to check redhat repo in scripts

build/scripts/update-from.sh
nickboldt: well we could move this script to another repo if you'd like, and then have CRW depend on it. Maybe put it in che-machine-exec?

build/scripts/update-from.sh
nickboldt: no, this is because the logic for staying current requires a comment like this:

```
# https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8-minimal
FROM registry.access.redhat.com/ubi8-minimal:8.2-267 as runtime
```

We could extend this so it supports other registries but then we also need to make sure the skopeo inspection works for those repos...

build/scripts/update-from.sh
nickboldt: which branch would be better?

nickboldt:
honestly I'm happy to move this somewhere else but I would love to have it be part of the release process so every 3 weeks we update to newer base images. that's why I wanted to put this wrapper script in the theia repo, so that we can incorporate it into the make-release.sh script.

I would also love to move the cico and release related stuff into a build/ folder so they're not cluttering the root.

benoitf:
IMHO you should provide a github action for the script and trigger it with cron for example

nickboldt:
> github action for the script and trigger it with cron for example

That would make sense if we want to check few newer base images more often than once a week (with each .0 or .z release).

build/scripts/update-from.sh
nickboldt: psst. where would you like this script moved/copied?

nickboldt:
crw-ci-test

nickboldt:
Closing as looks like it's hit a brick wall. :(

benoitf:
I think the way to go is to file PR when there is an update on images

nickboldt:
You're advocating for a manual solution? Don't you mean to *generate* a PR when there's an update available?

benoitf:
yes, I meant automatically generate PR if an update is required

nickboldt:
OK, then https://issues.redhat.com/browse/CRW-598 and https://issues.redhat.com/browse/CRW-910 track that

PR #765, opened 2020-06-04
Temporary skip one of the cypress tests which is not stable

azatsarynnyy:

### What does this PR do?
Temporary, disable one of the cypress tests which is not stable.
Sometimes, it requires to just rerun the CI job several times for passing the test.
For example:
https://ci.centos.org/job/devtools-che-theia-che-release/ - jobs 156, 157
https://ci.centos.org/job/devtools-che-theia-che-release/ - jobs 24, 25

As HappyPath tests are passing successfully, I propose to skip this test temporarily while investigating how to make it more stable.

### DIFF
--- a/dockerfiles/theia/e2e/cypress/integration/theia/typescript.spec.ts
+++ b/dockerfiles/theia/e2e/cypress/integration/theia/typescript.spec.ts
@@ -21,7 +21,7 @@ context('TypeScript', () => {
     });
 
     // Create a typescript file and check we can use the editor
-    it('Check Invalid Syntax', () => {
+    it.skip('Check Invalid Syntax', () => {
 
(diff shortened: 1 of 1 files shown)


vzhukovskii:
Do we need a separate issue to investigate the problem?

azatsarynnyy:
> Do we need a separate issue to investigate the problem?

sure, I'll create once the PR is merged

benoitf:
you could just skip tests during the release.

azatsarynnyy:
> you could just skip tests during the release.

@benoitf sure, we can but it's not about the releases only
this test false positive sometimes during any build: locally, on centos.ci, in GH workflows

azatsarynnyy:
the follow-up issue https://github.com/eclipse/che/issues/17095

PR #776, opened 2020-06-18
Fix release script to keep the same Theia versions for bugfix releases

azatsarynnyy:

### What does this PR do?
Fixes release script to set the latest Theia version when doing x.y.0 releases only.
But when doing other bugfix releases (e.g. x.y.1), Theia version must be the same as for corresponding y-release.

https://github.com/eclipse/che/issues/17188


azatsarynnyy:
crw-ci-test

PR #778, opened 2020-06-23
Open Che workspace with multiple workspace root

benoitf:
### What does this PR do?
Handle multiple workspace roots
Checks if there is a che.theia-workspace file.
  if there is one, returns that file as being a workspace root
  if none, generate the file by getting all clonedPath/path of the devfile projects

enhancements:
 - [x] make it optional for now (like if there is a special attribute in devfile, use that mode)
 - [ ] track new folders in /projects and add them to the che.theia-workspace file automatically (and if we delete folders)
 - [ ] allow to customize/provide different projects to be added in workspace root. Could be useful for a maven project to see only sub-folders/modules
 
 I think we could resolve those items ^^ in a separate issue: by default a workspace starts in `OFF` multi-root mode - so no breaking changes for the `master` branch
 
The current state of the PR - a workspace starts in `OFF` multi-root mode - so the behavior should be exactly as for `master` branch.
A user should add the following section to a `devfile` to turn on multi-root mode:
 ```
attributes:
  multiRoot: 'on'

 ```
 Please let me know if you have a better idea for the attribute and value as marker for multi-root mode - it's easy to change on the current step.

Change-Id: Idfb9370dc503bcfb5504ba37446f7f1ab171b515

https://github.com/eclipse/che/issues/17212

### TODO
- [x] Adding a workspace folder directly **after** cloning a project (not before, not during)
- [x] Align the corresponding plugin API of Theia with VS Code to provide the ability to display progress for views from plugin side (the PR on `Theia` side)
- [x] Minor: Replace CheApiService by WorkspaceService
- [x] Import a debug configuration from the devfile to a multi-root workspace https://github.com/eclipse/che-theia/pull/778#issuecomment-740552451
- [x] Fix config storage paths https://github.com/eclipse/che/issues/18548
- [x] Adapt handling of `Readme` files to multi-root workspace changes 
- [x] A debug configuration is not available for running
- [x] Add workspace folders in turn (there is a problem related to plugin API at adding project as workspace folder - a project is not displayed in `Projects` view when there are few projects are adding as workspace folders at the same time)  
- [x] Investigate if it's possible to export configurations on `Workspace` level https://github.com/eclipse/che-theia/pull/778#issuecomment-768936441
- [x] `Open Configurations` and `Add configuration` actions are not working for a multi-root workspace
- [x] Adapt Happy Path Tests to multi-root workspace changes
- [x] An ability to switch (turn on and turn off ) multi-root mode on a `devfile` level

Try it with two go projects, one with go modules, another without:
http://che.openshift.io/f/?url=https://gist.githubusercontent.com/benoitf/57ffa7e39e2e16a6a2f6756a44693e43/raw/5afea100c012fe05e325358f0a58577821a3cb99/golang-example.devfile.yaml

### Happy Path Channel

HAPPY_PATH_CHANNEL=next

### DIFF
--- a/extensions/eclipse-che-theia-plugin-ext/src/browser/che-debug-configuration-manager.ts
+++ b/extensions/eclipse-che-theia-plugin-ext/src/browser/che-debug-configuration-manager.ts
@@ -1,0 +1,32 @@
+/**********************************************************************
+ * Copyright (c) 2021 Red Hat, Inc.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ ***********************************************************************/
+
+import { injectable, postConstruct } from 'inversify';
+
+import { DebugConfigurationManager } from '@theia/debug/lib/browser/debug-configuration-manager';
+
+@injectable()
+export class CheDebugConfigurationManager extends DebugConfigurationManager {
+  @postConstruct()
+  protected async init(): Promise<void> {
+    super.init();
+
+    /**
+     * Theia creates a DebugConfigurationModel for each workspace folder in a workspace at starting the IDE.
+     * For the CHE multi-root workspace there no workspace folders at that step:
+     * CHE clones projects at starting the IDE and adds a workspace folder directly after cloning a project.
+     * That's why we need the following logic -
+     * DebugConfigurationManager should create the corresponding model when a workspace is changed (a workspace folder is added)
+     */
+    this.workspaceService.onWorkspaceChanged(() => {
+      this.updateModels();
+    });
+  }
+}
--- a/extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
+++ b/extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
@@ -46,12 +46,22 @@ import pluginExtBackendModule from '@theia/plugin-ext/lib/plugin-ext-backend-mod
 import pluginRemoteBackendModule from './plugin-remote-backend-module';
 import pluginVscodeBackendModule from '@theia/plugin-ext-vscode/lib/node/plugin-vscode-backend-module';
 
+const DEFAULT_THEIA_HOME_DIR = '/home/theia/';
+const DEFAULT_THEIA_DEV_HOME_DIR = '/home/theia-dev';
+
 interface CheckAliveWS extends ws {
   alive: boolean;
 }
 
-function modifyPathToLocal(origPath: string): string {
-  return path.join(os.homedir(), origPath.substr(0, '/home/theia/'.length));
+function modifyPathToLocal(originalPath: string): string {
+  if (originalPath.startsWith(DEFAULT_THEIA_HOME_DIR)) {
+    return path.join(os.homedir(), originalPath.substring(DEFAULT_THEIA_HOME_DIR.length));
+  }
+
+  if (originalPath.startsWith(DEFAULT_THEIA_DEV_HOME_DIR)) {
+    return path.join(os.homedir(), originalPath.substring(DEFAULT_THEIA_DEV_HOME_DIR.length));
+  }
+  return originalPath;
 }
 
 @injectable()
@@ -127,10 +137,22 @@ export class PluginRemoteInit {
     const originalStart = PluginManagerExtImpl.prototype.$start;
     PluginManagerExtImpl.prototype.$start = async function (params: PluginManagerStartParams): Promise<void> {
       const { hostLogPath, hostStoragePath, hostGlobalStoragePath } = params.configStorage;
+
+      const overriddenLogPath = modifyPathToLocal(hostLogPath);
+      await fs.ensureDir(overriddenLogPath);
+
+      const overriddenStoragePath = hostStoragePath ? modifyPathToLocal(hostStoragePath) : undefined;
+      if (overriddenStoragePath) {
+        await fs.ensureDir(overriddenStoragePath);
+      }
+
+      const overriddenGlobalStoragePath = modifyPathToLocal(hostGlobalStoragePath);
+      await fs.ensureDir(overriddenGlobalStoragePath);
+
       params.configStorage = {
-        hostLogPath: modifyPathToLocal(hostLogPath),
-        hostStoragePath: hostStoragePath ? modifyPathToLocal(hostStoragePath) : undefined,
-        hostGlobalStoragePath: modifyPathToLocal(hostGlobalStoragePath),
+        hostLogPath: overriddenLogPath,
+        hostStoragePath: overriddenStoragePath,
+        hostGlobalStoragePath: overriddenGlobalStoragePath,
       };
       // call original method
       return originalStart.call(this, params);
--- a/extensions/eclipse-che-theia-workspace/src/browser/che-navigator-widget.tsx
+++ b/extensions/eclipse-che-theia-workspace/src/browser/che-navigator-widget.tsx
@@ -1,0 +1,26 @@
+/**********************************************************************
+ * Copyright (c) 2021 Red Hat, Inc.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ ***********************************************************************/
+
+import * as React from 'react';
+
+import { injectable } from 'inversify';
+
+import { FileNavigatorWidget } from '@theia/navigator/lib/browser/navigator-widget';
+
+@injectable()
+export class CheFileNavigatorWidget extends FileNavigatorWidget {
+  protected renderEmptyMultiRootWorkspace(): React.ReactNode {
+    return (
+      <div className="theia-navigator-container">
+        <div className="center">No projects in the workspace yet</div>
+      </div>
+    );
+  }
+}
--- a/extensions/eclipse-che-theia-workspace/src/node/che-workspace-server.ts
+++ b/extensions/eclipse-che-theia-workspace/src/node/che-workspace-server.ts
@@ -8,71 +8,72 @@
  * SPDX-License-Identifier: EPL-2.0
  ***********************************************************************/
 
-import { injectable, inject } from 'inversify';
-import { DefaultWorkspaceServer } from '@theia/workspace/lib/node/default-workspace-server';
-import { CheApiService } from '@eclipse-che/theia-plugin-ext/lib/common/che-protocol';
-import { FileUri } from '@theia/core/lib/node';
 import * as fs from 'fs-extra';
 import * as path from 'path';
 
+import { inject, injectable } from 'inversify';
+
+import { DefaultWorkspaceServer } from '@theia/workspace/lib/node/default-workspace-server';
+import { FileUri } from '@theia/core/lib/node';
+import { WorkspaceService } from '@eclipse-che/theia-remote-api/lib/common/workspace-service';
+
 interface TheiaWorkspace {
-    folders: TheiaWorkspacePath[]
-};
+  folders: TheiaWorkspacePath[];
+}
 
 interface TheiaWorkspacePath {
-    path: string
-};
+  path: string;
+}
 
 @injectable()
 export class CheWorkspaceServer extends DefaultWorkspaceServer {
+  @inject(WorkspaceService)
+  protected workspaceService: WorkspaceService;
 
-    @inject(CheApiService)
-    private cheApiService: CheApiService;
-
-    // override any workspace that could have been defined through CLI and use entries from the devfile
-    // if not possible, use default method
-    protected async getRoot(): Promise<string | undefined> {
-
-        let projectsRoot: string;
-        if (process.env.CHE_PROJECTS_ROOT) {
-            projectsRoot = process.env.CHE_PROJECTS_ROOT;
-        } else {
-            projectsRoot = '/projects';
-        }
-
-        // first, check if we have a che.theia-workspace file
-        const cheTheiaWorkspaceFile = path.resolve(projectsRoot, 'che.theia-workspace');
-        const exists = await fs.pathExists(cheTheiaWorkspaceFile);
-        if (exists) {
-            return FileUri.create(cheTheiaWorkspaceFile).toString();
-        }
-
-        // no, then create the file
-
-        const workspace = await this.cheApiService.currentWorkspace();
-        const devfile = workspace.devfile;
-        if (devfile) {
-            const projects = devfile.projects;
-            if (projects) {
-                // create a struc for each project
-                const theiaWorkspace: TheiaWorkspace = { folders: [] };
-                for (const project of projects) {
-                    const projectPath = project.clonePath ? path.join(projectsRoot, project.clonePath) : path.join(projectsRoot, project.name!);
-                    // check parent folder exists
-                    const parentDir = path.resolve(projectPath, '..');
-                    await fs.ensureDir(parentDir);
-                    theiaWorkspace.folders.push({ path: FileUri.create(projectPath).toString() });
-                }
-
-                // now, need to write the content of this file
-                await fs.writeFile(cheTheiaWorkspaceFile, JSON.stringify(theiaWorkspace), { encoding: 'utf8' });
-
-                // return this content
-                return FileUri.create(cheTheiaWorkspaceFile).toString();
-            }
-        }
-
-        return super.getRoot();
+  // override any workspace that could have been defined through CLI and use entries from the devfile
+  // if not possible, use default method
+  protected async getRoot(): Promise<string | undefined> {
+    let projectsRoot: string;
+    if (process.env.CHE_PROJECTS_ROOT) {
+      projectsRoot = process.env.CHE_PROJECTS_ROOT;
+    } else {
+      projectsRoot = '/projects';
     }
 
+    // first, check if we have a che.theia-workspace file
+    const cheTheiaWorkspaceFile = path.resolve(projectsRoot, 'che.theia-workspace');
+    const exists = await fs.pathExists(cheTheiaWorkspaceFile);
+    if (exists) {
+      return FileUri.create(cheTheiaWorkspaceFile).toString();
+    }
+
+    // no, then create the file
+
+    const workspace = await this.workspaceService.currentWorkspace();
+    const devfile = workspace.devfile;
+    if (devfile) {
+      const projects = devfile.projects;
+      if (projects) {
+        // create a struc for each project
+        const theiaWorkspace: TheiaWorkspace = { folders: [] };
+        for (const project of projects) {
+          const projectPath = project.clonePath
+            ? path.join(projectsRoot, project.clonePath)
+            : path.join(projectsRoot, project.name!);
+          // check parent folder exists
+          const parentDir = path.resolve(projectPath, '..');
+          await fs.ensureDir(parentDir);
+          theiaWorkspace.folders.push({ path: FileUri.create(projectPath).toString() });
+        }
+
+        // now, need to write the content of this file
+        await fs.writeFile(cheTheiaWorkspaceFile, JSON.stringify(theiaWorkspace), { encoding: 'utf8' });
+
+        // return this content
+        return FileUri.create(cheTheiaWorkspaceFile).toString();
+      }
+    }
+
+    return super.getRoot();
+  }
 }
--- a/extensions/eclipse-che-theia-workspace/src/node/che-workspace-server.ts
+++ b/extensions/eclipse-che-theia-workspace/src/node/che-workspace-server.ts
@@ -1,0 +1,78 @@
+/**********************************************************************
+ * Copyright (c) 2021 Red Hat, Inc.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ ***********************************************************************/
+
+import { injectable, inject } from 'inversify';
+import { DefaultWorkspaceServer } from '@theia/workspace/lib/node/default-workspace-server';
+import { CheApiService } from '@eclipse-che/theia-plugin-ext/lib/common/che-protocol';
+import { FileUri } from '@theia/core/lib/node';
+import * as fs from 'fs-extra';
+import * as path from 'path';
+
+interface TheiaWorkspace {
+    folders: TheiaWorkspacePath[]
+};
+
+interface TheiaWorkspacePath {
+    path: string
+};
+
+@injectable()
+export class CheWorkspaceServer extends DefaultWorkspaceServer {
+
+    @inject(CheApiService)
+    private cheApiService: CheApiService;
+
+    // override any workspace that could have been defined through CLI and use entries from the devfile
+    // if not possible, use default method
+    protected async getRoot(): Promise<string | undefined> {
+
+        let projectsRoot: string;
+        if (process.env.CHE_PROJECTS_ROOT) {
+            projectsRoot = process.env.CHE_PROJECTS_ROOT;
+        } else {
+            projectsRoot = '/projects';
+        }
+
+        // first, check if we have a che.theia-workspace file
+        const cheTheiaWorkspaceFile = path.resolve(projectsRoot, 'che.theia-workspace');
+        const exists = await fs.pathExists(cheTheiaWorkspaceFile);
+        if (exists) {
+            return FileUri.create(cheTheiaWorkspaceFile).toString();
+        }
+
+        // no, then create the file
+
+        const workspace = await this.cheApiService.currentWorkspace();
+        const devfile = workspace.devfile;
+        if (devfile) {
+            const projects = devfile.projects;
+            if (projects) {
+                // create a struc for each project
+                const theiaWorkspace: TheiaWorkspace = { folders: [] };
+                for (const project of projects) {
+                    const projectPath = project.clonePath ? path.join(projectsRoot, project.clonePath) : path.join(projectsRoot, project.name!);
+                    // check parent folder exists
+                    const parentDir = path.resolve(projectPath, '..');
+                    await fs.ensureDir(parentDir);
+                    theiaWorkspace.folders.push({ path: FileUri.create(projectPath).toString() });
+                }
+
+                // now, need to write the content of this file
+                await fs.writeFile(cheTheiaWorkspaceFile, JSON.stringify(theiaWorkspace), { encoding: 'utf8' });
+
+                // return this content
+                return FileUri.create(cheTheiaWorkspaceFile).toString();
+            }
+        }
+
+        return super.getRoot();
+    }
+
+}
--- a/extensions/eclipse-che-theia-workspace/src/node/che-workspace-server.ts
+++ b/extensions/eclipse-che-theia-workspace/src/node/che-workspace-server.ts
@@ -54,22 +54,8 @@ export class CheWorkspaceServer extends DefaultWorkspaceServer {
     if (devfile) {
       const projects = devfile.projects;
       if (projects) {
-        // create a struc for each project
         const theiaWorkspace: TheiaWorkspace = { folders: [] };
-        for (const project of projects) {
-          const projectPath = project.clonePath
-            ? path.join(projectsRoot, project.clonePath)
-            : path.join(projectsRoot, project.name!);
-          // check parent folder exists
-          const parentDir = path.resolve(projectPath, '..');
-          await fs.ensureDir(parentDir);
-          theiaWorkspace.folders.push({ path: FileUri.create(projectPath).toString() });
-        }
-
-        // now, need to write the content of this file
         await fs.writeFile(cheTheiaWorkspaceFile, JSON.stringify(theiaWorkspace), { encoding: 'utf8' });
-
-        // return this content
         return FileUri.create(cheTheiaWorkspaceFile).toString();
       }
     }
--- a/extensions/eclipse-che-theia-workspace/src/node/che-workspace-server.ts
+++ b/extensions/eclipse-che-theia-workspace/src/node/che-workspace-server.ts
@@ -11,11 +11,11 @@
 import * as fs from 'fs-extra';
 import * as path from 'path';
 
+import { Workspace, WorkspaceService } from '@eclipse-che/theia-remote-api/lib/common/workspace-service';
 import { inject, injectable } from 'inversify';
 
 import { DefaultWorkspaceServer } from '@theia/workspace/lib/node/default-workspace-server';
 import { FileUri } from '@theia/core/lib/node';
-import { WorkspaceService } from '@eclipse-che/theia-remote-api/lib/common/workspace-service';
 
 interface TheiaWorkspace {
   folders: TheiaWorkspacePath[];
@@ -33,33 +33,29 @@ export class CheWorkspaceServer extends DefaultWorkspaceServer {
   // override any workspace that could have been defined through CLI and use entries from the devfile
   // if not possible, use default method
   protected async getRoot(): Promise<string | undefined> {
-    let projectsRoot: string;
-    if (process.env.CHE_PROJECTS_ROOT) {
-      projectsRoot = process.env.CHE_PROJECTS_ROOT;
-    } else {
-      projectsRoot = '/projects';
+    const workspace = await this.workspaceService.currentWorkspace();
+    if (!isMultiRoot(workspace)) {
+      return super.getRoot();
     }
+
+    const projectsRootEnvVariable = process.env.CHE_PROJECTS_ROOT;
+    const projectsRoot = projectsRootEnvVariable ? projectsRootEnvVariable : '/projects';
 
     // first, check if we have a che.theia-workspace file
     const cheTheiaWorkspaceFile = path.resolve(projectsRoot, 'che.theia-workspace');
+    const cheTheiaWorkspaceFileUri = FileUri.create(cheTheiaWorkspaceFile);
     const exists = await fs.pathExists(cheTheiaWorkspaceFile);
-    if (exists) {
-      return FileUri.create(cheTheiaWorkspaceFile).toString();
+    if (!exists) {
+      // no, then create the file
+      const theiaWorkspace: TheiaWorkspace = { folders: [] };
+      await fs.writeFile(cheTheiaWorkspaceFile, JSON.stringify(theiaWorkspace), { encoding: 'utf8' });
     }
 
-    // no, then create the file
-
-    const workspace = await this.workspaceService.currentWorkspace();
-    const devfile = workspace.devfile;
-    if (devfile) {
-      const projects = devfile.projects;
-      if (projects) {
-        const theiaWorkspace: TheiaWorkspace = { folders: [] };
-        await fs.writeFile(cheTheiaWorkspaceFile, JSON.stringify(theiaWorkspace), { encoding: 'utf8' });
-        return FileUri.create(cheTheiaWorkspaceFile).toString();
-      }
-    }
-
-    return super.getRoot();
+    return cheTheiaWorkspaceFileUri.toString();
   }
+}
+
+function isMultiRoot(workspace: Workspace): boolean {
+  const devfile = workspace.devfile;
+  return !!devfile && !!devfile.attributes && !!devfile.attributes.multiRoot && devfile.attributes.multiRoot === 'on';
 }
--- a/plugins/task-plugin/src/export/export-configs-manager.ts
+++ b/plugins/task-plugin/src/export/export-configs-manager.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (c) 2019-2020 Red Hat, Inc.
+ * Copyright (c) 2019-2021 Red Hat, Inc.
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -8,12 +8,10 @@
  * SPDX-License-Identifier: EPL-2.0
  ***********************************************************************/
 
-import * as startPoint from '../task-plugin-backend';
-import * as theia from '@theia/plugin';
-
 import { inject, injectable, multiInject } from 'inversify';
 
 import { CheWorkspaceClient } from '../che-workspace-client';
+import { LaunchConfigurationsExporter } from './launch-configs-exporter';
 import { che as cheApi } from '@eclipse-che/api';
 
 export const ConfigurationsExporter = Symbol('ConfigurationsExporter');
@@ -45,26 +43,21 @@ export class ExportConfigurationsManager {
   @multiInject(ConfigurationsExporter)
   protected readonly exporters: ConfigurationsExporter[];
 
-  init(): void {
-    this.export();
+  @inject(LaunchConfigurationsExporter)
+  protected readonly launchConfigurationsExporter: LaunchConfigurationsExporter;
 
-    theia.workspace.onDidChangeWorkspaceFolders(
-      event => {
-        const workspaceFolders = event.added;
-        if (workspaceFolders && workspaceFolders.length > 0) {
-          this.export();
-        }
-      },
-      undefined,
-      startPoint.getSubscriptions()
-    );
+  protected cheCommands: cheApi.workspace.Command[] = [];
+
+  async init(): Promise<void> {
+    this.cheCommands = await this.cheWorkspaceClient.getCommands();
+    this.launchConfigurationsExporter.init(this.cheCommands);
+    this.export();
   }
 
   async export(): Promise<void> {
     const exportPromises = [];
-    const cheCommands = await this.cheWorkspaceClient.getCommands();
     for (const exporter of this.exporters) {
-      exportPromises.push(this.doExport(cheCommands, exporter));
+      exportPromises.push(this.doExport(this.cheCommands, exporter));
     }
 
     await Promise.all(exportPromises);
--- a/plugins/task-plugin/src/export/export-configs-manager.ts
+++ b/plugins/task-plugin/src/export/export-configs-manager.ts
@@ -8,6 +8,9 @@
  * SPDX-License-Identifier: EPL-2.0
  ***********************************************************************/
 
+import * as startPoint from '../task-plugin-backend';
+import * as theia from '@theia/plugin';
+
 import { inject, injectable, multiInject } from 'inversify';
 
 import { CheWorkspaceClient } from '../che-workspace-client';
@@ -41,6 +44,21 @@ export class ExportConfigurationsManager {
 
   @multiInject(ConfigurationsExporter)
   protected readonly exporters: ConfigurationsExporter[];
+
+  init(): void {
+    this.export();
+
+    theia.workspace.onDidChangeWorkspaceFolders(
+      event => {
+        const workspaceFolders = event.added;
+        if (workspaceFolders && workspaceFolders.length > 0) {
+          this.export();
+        }
+      },
+      undefined,
+      startPoint.getSubscriptions()
+    );
+  }
 
   async export(): Promise<void> {
     const exportPromises = [];
--- a/plugins/task-plugin/src/export/launch-configs-exporter.ts
+++ b/plugins/task-plugin/src/export/launch-configs-exporter.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (c) 2019-2020 Red Hat, Inc.
+ * Copyright (c) 2019-2021 Red Hat, Inc.
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -8,6 +8,7 @@
  * SPDX-License-Identifier: EPL-2.0
  ***********************************************************************/
 
+import * as startPoint from '../task-plugin-backend';
 import * as theia from '@theia/plugin';
 
 import { ensureDirExists, modify, writeFile } from '../utils';
@@ -32,14 +33,28 @@ export class LaunchConfigurationsExporter implements ConfigurationsExporter {
   @inject(VsCodeLaunchConfigsExtractor)
   protected readonly vsCodeLaunchConfigsExtractor: VsCodeLaunchConfigsExtractor;
 
-  async export(commands: cheApi.workspace.Command[]): Promise<void> {
-    if (!theia.workspace.workspaceFolders) {
+  async init(commands: cheApi.workspace.Command[]): Promise<void> {
+    theia.workspace.onDidChangeWorkspaceFolders(
+      event => {
+        const workspaceFolders: theia.WorkspaceFolder[] | undefined = event.added;
+        if (workspaceFolders && workspaceFolders.length > 0) {
+          this.export(commands, workspaceFolders);
+        }
+      },
+      undefined,
+      startPoint.getSubscriptions()
+    );
+  }
+
+  async export(commands: cheApi.workspace.Command[], workspaceFolders?: theia.WorkspaceFolder[]): Promise<void> {
+    workspaceFolders = workspaceFolders ? workspaceFolders : theia.workspace.workspaceFolders;
+    if (!workspaceFolders) {
       return;
     }
 
     const exportConfigsPromises: Promise<void>[] = [];
 
-    for (const workspaceFolder of theia.workspace.workspaceFolders) {
+    for (const workspaceFolder of workspaceFolders) {
       exportConfigsPromises.push(this.doExport(workspaceFolder, commands));
     }
     await Promise.all(exportConfigsPromises);
--- a/plugins/task-plugin/src/export/launch-configs-exporter.ts
+++ b/plugins/task-plugin/src/export/launch-configs-exporter.ts
@@ -10,8 +10,8 @@
 
 import * as theia from '@theia/plugin';
 
+import { ensureDirExists, modify, writeFile } from '../utils';
 import { inject, injectable } from 'inversify';
-import { modify, writeFileSync } from '../utils';
 
 import { ConfigFileLaunchConfigsExtractor } from '../extract/config-file-launch-configs-extractor';
 import { ConfigurationsExporter } from './export-configs-manager';
@@ -46,23 +46,23 @@ export class LaunchConfigurationsExporter implements ConfigurationsExporter {
   }
 
   async doExport(workspaceFolder: theia.WorkspaceFolder, commands: cheApi.workspace.Command[]): Promise<void> {
-    const launchConfigFileUri = this.getConfigFileUri(workspaceFolder.uri.path);
-    const configFileConfigs = this.configFileLaunchConfigsExtractor.extract(launchConfigFileUri);
+    const workspaceFolderPath = workspaceFolder.uri.path;
+    const launchConfigFilePath = resolve(workspaceFolderPath, CONFIG_DIR, LAUNCH_CONFIG_FILE);
+    const configFileConfigs = await this.configFileLaunchConfigsExtractor.extract(launchConfigFilePath);
     const vsCodeConfigs = this.vsCodeLaunchConfigsExtractor.extract(commands);
 
     const configFileContent = configFileConfigs.content;
     if (configFileContent) {
-      this.saveConfigs(
-        launchConfigFileUri,
+      return this.saveConfigs(
+        workspaceFolderPath,
         configFileContent,
         this.merge(configFileConfigs.configs, vsCodeConfigs.configs, this.getConsoleConflictLogger())
       );
-      return;
     }
 
     const vsCodeConfigsContent = vsCodeConfigs.content;
     if (vsCodeConfigsContent) {
-      this.saveConfigs(launchConfigFileUri, vsCodeConfigsContent, vsCodeConfigs.configs);
+      return this.saveConfigs(workspaceFolderPath, vsCodeConfigsContent, vsCodeConfigs.configs);
     }
   }
 
@@ -100,13 +100,32 @@ export class LaunchConfigurationsExporter implements ConfigurationsExporter {
     return JSON.stringify(properties1) === JSON.stringify(properties2);
   }
 
-  private getConfigFileUri(rootDir: string): string {
-    return resolve(rootDir.toString(), CONFIG_DIR, LAUNCH_CONFIG_FILE);
-  }
+  private async saveConfigs(
+    workspaceFolderPath: string,
+    content: string,
+    configurations: theia.DebugConfiguration[]
+  ): Promise<void> {
+    /*
+        There is an issue related to file watchers: the watcher only reports the first directory when creating recursively directories.
+        For example:
+            - we would like to create /projects/someProject/.theia/launch.json recursively
+            - /projects/someProject directory already exists
+            - .theia directory and launch.json file should be created
+            - as result file watcher fires an event that .theia directory was created, there is no an event about launch.json file
 
-  private saveConfigs(launchConfigFileUri: string, content: string, configurations: theia.DebugConfiguration[]): void {
+        The issue is reproduced not permanently.
+
+        We had to use the workaround to avoid the issue: first we create the directory and then - config file
+    */
+
+    const configDirPath = resolve(workspaceFolderPath, CONFIG_DIR);
+    await ensureDirExists(configDirPath);
+
+    const launchConfigFilePath = resolve(configDirPath, LAUNCH_CONFIG_FILE);
+    await ensureDirExists(launchConfigFilePath);
+
     const result = modify(content, ['configurations'], configurations, formattingOptions);
-    writeFileSync(launchConfigFileUri, result);
+    return writeFile(launchConfigFilePath, result);
   }
 
   private getConsoleConflictLogger(): (config1: theia.DebugConfiguration, config2: theia.DebugConfiguration) => void {
--- a/plugins/task-plugin/src/utils.ts
+++ b/plugins/task-plugin/src/utils.ts
@@ -8,6 +8,7 @@
  * SPDX-License-Identifier: EPL-2.0
  ***********************************************************************/
 
+import * as fs from 'fs-extra';
 import * as jsoncparser from 'jsonc-parser';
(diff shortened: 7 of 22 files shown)


tsmaeder:
I think an approach that computes the workspace roots at the same time as doing the cloning would make it easier keeping projects and workspace roots in sync as we evolve the code. We can also make sure language servers are not seeing half-cloned projects that way. https://github.com/eclipse/che-theia/pull/408 implements that approach and also cleans up the UI a bit. I would recommend we base our work on that PR.

sunix:
From what i see in the exemple, it's only possible to define one workspace root per project.
But in some usecase, we would like to be able to use 1+ roots in subfolders (a maven project + nodejs in the same git project)
What about that proposal: 
https://github.com/eclipse/che/issues/15347#issue-529480447

benoitf:
@sunix: not yet as it requires a new devfile format

> Also why adding a intermediate file? che.theia-workspace file

to make it persistent across workspace restart.

sunix:
> to make it persistent across workspace restart.

Ok, I feel like this is all temporary ... waiting for a new devfile format. What are the next steps?

RomanNikitenko:
Faced with the issue related to cloning a project, investigating...

cloning_error

benoitf:
@RomanNikitenko looks like it was related to my hack "commit"

RomanNikitenko:
@benoitf 
hack "commit" was on task plugin side (if i'm not mistaken)
current problem is related to cloning
but probably removing the commit somehow broke the behavior...

it's also strange for me because yesterday I tested after removing the commit using my another devfile with `theia` and `che-theia` projects and it worked well
today I took a devfile from happy path tests and faced  with the issue...

benoitf:
@RomanNikitenko as you've a launch.json in the workspace root folder, theia try to clone project but it can't as there is already some files there

RomanNikitenko:
> I think an approach that computes the workspace roots at the same time as doing the cloning would make it easier keeping projects and workspace roots in sync as we evolve the code. We can also make sure language servers are not seeing half-cloned projects that way. #408 implements that approach and also cleans up the UI a bit. I would recommend we base our work on that PR.

@tsmaeder 
I tried to apply your idea, so adding a workspace folder directly **after** cloning a project.
I'm worried about the behavior related to project tree with this approach.
`You have not yet added a folder to the workspace` message and `Add Folder` button is displayed while a project is cloning.
Sure in the right bottom area the notification about cloning project process is displayed at the same time.
But I guess the behavior can confuse an user, especially for case of slow connection or a big project. 

importing

tsmaeder:
@RomanNikitenko I agree this can be confusing when there is a long wait before cloning starts. A couple of ideas:
1. Unbind the UI that shows the "Open Workspace" button.
2. Maybe open some (closable) modal UI at startup that tells the user cloning will start shortly?

RomanNikitenko:
@tsmaeder 
>  I agree this can be confusing when there is a long wait **before cloning starts**. 

To clarify: in my implementation a workspace folder is added when **cloning is completed**, because you mentioned that:
> We can also make sure language servers are not seeing **half-cloned projects** that way.

tsmaeder:
> 
> To clarify: in my implementation a workspace folder is added when **cloning is completed**, because you mentioned that:

I understand that...but when the user sees that projects are being added, she knows something is going on; when the IDE just sits there, it's much worse.

RomanNikitenko:
> I understand that...but when the user sees that projects are being added, she knows something is going on; when the IDE just sits there, it's much worse.

I agree. I just highlighted some details in my previous comment to be sure that we speak about the same things.

azatsarynnyy:
Let's try the solution we've agreed on during our standup call, yesterday.
Customize the Explorer in Che Theia to get rid of the existing text + button and display some progress instead, while cloning.

To not spend much time creating something new, we can try to reuse the existing progress indicator - this fast-moving line for example:
Peek 2020-10-28 13-08

... and some text, e.g. `Cloning the sources...`

RomanNikitenko:
I aligned the corresponding plugin API of `Theia` with `VS Code` and provided the ability to display progress for views from plugin side, so now we can improve the behavior described [here](https://github.com/eclipse/che-theia/pull/778#issuecomment-717191869) and apply [this approach](https://github.com/eclipse/che-theia/pull/778#issuecomment-717867792)

RomanNikitenko:
[crw-ci-test]

benoitf:
[crw-ci-test]

benoitf:
trying with `next` Happy Path workflow

RomanNikitenko:
@benoitf 
maybe some adaptation is required on happy path tests side - the behavior was changed: `projects` as root ---> Che Workspace
I'll take a look...

benoitf:
yes it searches for
```
1) Validation of workspace start
10:13:09        Wait for workspace readiness:
10:13:09      TimeoutError: Exceeded maximum gettin of the 'class' attribute attempts, from the 'By(css selector, div[data-node-id='/projects/petclinic:/projects/petclinic'])' element
```

but the node is now `/petclinic`

benoitf:
trying the "open PR" link it seems it's still written in the css node /projects/something but there is no colon
also there is a refresh as it's not finding something but I don't know in the usual case if a refresh is performed or not so yes, need to investigate

and dom seems to be valid
```
<div class="theia-TreeNode theia-CompositeTreeNode theia-ExpandableTreeNode theia-FileStatNode theia-DirNode" draggable="true" title="/projects/petclinic" style="padding-left: 0px;"><div class="theia-TreeNodeContent"><div data-node-id="/projects/petclinic" class="theia-TreeNodeSegment theia-ExpansionToggle theia-mod-collapsed">
```

RomanNikitenko:
I'm going to create the corresponding branch in `che` project to investigate/play with tests

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
I adapted some tests for the current PR changes within https://github.com/eclipse/che/pull/18417. 
Now happy path tests unhappy because of waiting `Activating Java Test Runner` process.

If I understand correctly, happy path tests wait for this process when `java` file is open. But with the current PR changes `Activating Java Test Runner` process starts when a workspace folder is open and sure it's completed when `java` file is open.
So some adaptation for it ^^ on happy path tests side is required.
I'll check it tomorrow.

RomanNikitenko:
[crw-ci-test]

dmytro-ndp:
[crw-ci-test]

dmytro-ndp:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
Looks like 2 last failures of happy path tests are related to https://github.com/eclipse/che/issues/18034

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
I detected some problems related to import a `debug` configuration from `devfile` to workspace.

At switching to multi-root workspace the behavior (without any changes - so as is from `master` branch) will be: 
- a configuration is imported from the `devfile` to the corresponding config file `someProject/.theia/launch.json` for each project of the workspace.
- as result we have duplicates:
duplicates

Another problem: after starting a workspace user gets modified file -  `someProject/.theia/launch.json` - for each imported project.

The current behavior for `master`:  
- a configuration is imported from the `devfile` to the corresponding config file `launch.json` in the `/workspaceFolder/.theia` directory, so `/projects/.theia/launch.json`, as default `workspaceFolder` for `master` branch is `/projects`.
- then the configuration is available for running for all projects placed in the workspace folder - so `/projects` directory in our case

`Theia` behavior for multi-root workspace: each project/workspace folder has own `launch.json` file with configurations

`VS Code` behavior for multi-root workspace:
- each project/workspace folder has own `launch.json` file with configurations
- it's possible to set a configuration inside `User` settings - [Global launch configuration](https://code.visualstudio.com/docs/editor/debugging#_global-launch-configuration)
- it's possible to set a configuration inside `Workspace` settings - [Workspace launch configuration](https://code.visualstudio.com/docs/editor/multi-root-workspaces#_workspace-launch-configurations) 

I think we could consider the following approaches to resolve the problems:
- import a configuration to `Workspace` settings
- import a configuration to `User` settings

It should be mentioned that at the moment `Theia` doesn't support  `Global` and `Workspace` launch configurations.

For both cases we will have the configuration defined as global, so this one will be available for running for all projects of a workspace. If our goal is `project` specific configuration - then I think we should provide some mechanism on `devfile` level for mapping configuration-project.

azatsarynnyy:
I'd say it's ok as for the first stage. In the second step, we can make the debug configurations user-scoped.

azatsarynnyy:
After testing it with go Workspace I see strange behavior.
I used Che-provided go stack with the replaced editor to enable multi-root, by default.

<details>
<pre>
apiVersion: 1.0.0
metadata:
  name: golang-k09os
attributes:
  persistVolumes: 'false'
projects:
-
  name: example
  source:
    type: git
    location: https://github.com/golang/example.git
  clonePath: src/github.com/golang/example/
-
  name: golang-echo-realworld-example-app
  source:
    type: git
    location: https://github.com/xesina/golang-echo-realworld-example-app.git
  clonePath: src/github.com/xesina/golang-echo-realworld-example-app
components:
  - memoryLimit: 768Mi
    type: cheEditor
    reference: 'https://raw.githubusercontent.com/chepullreq4/pr-check-files/master/che-theia/pr-778/che_theia_meta.yaml'
    alias: che-theia
  - id: golang/go/latest
    memoryLimit: 512Mi
    preferences:
      go.lintTool: golangci-lint
      go.useLanguageServer: true
      go.lintFlags: '--fast'
    type: chePlugin
    alias: go-plugin
    env:
      - value: 'off'
        name: GO111MODULE
  - mountSources: true
    endpoints:
      - name: 8080-tcp
        port: 8080
    memoryLimit: 2Gi
    type: dockerimage
    image: 'quay.io/eclipse/che-golang-1.14:7.22.0'
    alias: go-cli
    env:
      - value: $(CHE_PROJECTS_ROOT)
        name: GOPATH
      - value: /tmp/.cache
        name: GOCACHE
commands:
  - name: 1.1 Run outyet
    actions:
      - workdir: '${CHE_PROJECTS_ROOT}/src/github.com/golang/example/outyet'
        type: exec
        command: go get -d && go run main.go
        component: go-cli
  - name: 1.2 Stop outyet
    actions:
      - type: exec
        command: kill $(pidof go)
        component: go-cli
  - name: 1.3 Test outyet
    actions:
      - workdir: '${CHE_PROJECTS_ROOT}/src/github.com/golang/example/outyet'
        type: exec
        command: go test
        component: go-cli
  - name: '2.1 xenisa :: install dependencies'
    actions:
      - workdir: '${GOPATH}/src/github.com/xesina/golang-echo-realworld-example-app'
        type: exec
        command: go mod download
        component: go-cli
  - name: '2.2 xenisa :: run'
    actions:
      - workdir: '${GOPATH}/src/github.com/xesina/golang-echo-realworld-example-app'
        type: exec
        command: go run main.go
        component: go-cli
  - name: '2.3 xenisa :: build'
    actions:
      - workdir: '${GOPATH}/src/github.com/xesina/golang-echo-realworld-example-app'
        type: exec
        command: go build
        component: go-cli
  - name: '2.4 xenisa :: test'
    actions:
      - workdir: '${GOPATH}/src/github.com/xesina/golang-echo-realworld-example-app'
        type: exec
        command: go test ./...
        component: go-cli
  - name: Run current file
    actions:
      - workdir: '${fileDirname}'
        type: exec
        command: 'go get -d && go run ${file}'
        component: go-cli
  - name: Debug current file
    actions:
      - referenceContent: |
          {
            "version": "0.2.0",
            "configurations": [
              {
                "name": "Debug current file",
                "type": "go",
                "request": "launch",
                "mode": "auto",
                "program": "${fileDirname}"
              }
            ]
          }
        type: vscode-launch
</pre>
</details>

Only one workspace root is displayed, after cloning the projects. But Devflie defines two.

image

I thought maybe it was because of `clonePath`. But after removing `clonePath` behavior is the same.

RomanNikitenko:
@azatsarynnyy 
thanks for the testing, I'm testing the use case using provided `devfile`...

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test --rebuild]

benoitf:
@RomanNikitenko should I remove 'draft status' of this PR ?

RomanNikitenko:
@benoitf 
I'm waiting for happy path tests to check https://github.com/eclipse/che-theia/pull/778#issuecomment-741571730
Looks like @azatsarynnyy tested my changes without last commit as happy path tests were not working some time, so the images don't contain my last changes.

I'll test the changes again when happy path tests provide updated images and open the PR for review if everything works well

tsmaeder:
> 
> 
> I'd say it's ok as for the first stage. In the second step, we can make the debug configurations user-scoped.

There is no user scope for debug configurations in Theia, as far as I know.

RomanNikitenko:
> > I'd say it's ok as for the first stage. In the second step, we can make the debug configurations user-scoped.
> 
> There is no user scope for debug configurations in Theia, as far as I know.

yes, I mentioned about it 

> It should be mentioned that at the moment Theia doesn't support Global and Workspace launch configurations.

RomanNikitenko:
[crw-ci-test --rebuild]

azatsarynnyy:
[crw-ci-test --rebuild]

azatsarynnyy:
[crw-ci-test --rebuild]

azatsarynnyy:
[crw-ci-test --rebuild]

azatsarynnyy:
[crw-ci-test --rebuild]

azatsarynnyy:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
Something wrong with happy path tests again :disappointed_relieved:
I'm not able even to get tests report
 
tests_report

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
The tests fail because of bad credentials:
```
09:41:11 Step 10/39 : ADD https://${GITHUB_TOKEN}:x-oauth-basic@api.github.com/repos/${THEIA_GITHUB_REPO}/git/refs/heads/master /tmp/branch_info.json
09:41:11 ADD failed: failed to GET https://****:x-oauth-basic@api.github.com/repos/eclipse-theia/theia/git/refs/heads/master with status 401 Unauthorized: {"message":"Bad credentials","documentation_url":"https://docs.github.com/rest"}
```

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test --rebuild]

dmytro-ndp:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

dmytro-ndp:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
The current state of the PR: happy path tests fail at running a debug configuration.
On the video from happy path tests report I see that a configuration from the `devfile` is exported to the `launch.json` file. But at opening debug view there is no configurations for running.

Unfortunately I'm not able to reproduce it manually, so it's very hard to debug and find the cause of the problem. At the same time happy path tests reflect this issue permanently.

I'm going to come back to the issue after my vacation.

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test]

benoitf:
🎉  E2E Happy path tests succeed  !

RomanNikitenko:
@benoitf 
Unfortunately it was by chance...

Happy path tests fail permanently at running a debug configuration. I'm not able to reproduce it manually, so I switched the assembly on my custom branch in `theia` and at the moment I'm debugging the state:
- add logs in my custom branch in `theia`
- rerun the happy path tests

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

azatsarynnyy:
Currently, testing of the PR is blocked by https://github.com/eclipse/che/issues/18828

RomanNikitenko:
Looks like the cause of the failed tests is the same as for https://github.com/eclipse/che/issues/18826
On the video provided by tests I see that there is no project when a workspace is started

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
I checked why the tests were unhappy for the previous build ^^

connection_interrupted

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
After my last changes I have a lot of successful reports from happy path tests and 2 failed reports.
One of the failed report was related to  https://github.com/eclipse/che/issues/18826, please see details https://github.com/eclipse/che-theia/pull/778#issuecomment-762851972
Another one was related to interrupted connection: https://github.com/eclipse/che-theia/pull/778#issuecomment-763379324

So, I think that happy path tests are pretty stable for the current PR.

I was switched to another urgent issue. I'm going to come back soon to do manual testing for different use cases related to multi root workspaces and open the PR for review.

RomanNikitenko:
I found that there is an ability in `theia` to set launch configurations on workspace level.
Also there is a PR under review in `theia` which should improve the behavior and fix some bugs related to running launch configs from workspace level.

I think we could export launch configurations from `devfile to workspace config file.

Using such approach we could avoid:
 - unstaged git changes for each imported project,
- configurations duplication for each imported project

I created the corresponding issue: https://github.com/eclipse/che/issues/18866

@benoitf @azatsarynnyy 
please let me know your opinion

benoitf:
suggestion looks fine to me

azatsarynnyy:
As we have repeatedly discussed during our stand up calls, I think exporting launch configurations from devfile to workspace config file is the required step to get proper multi-root support and have a consistent debug flow in both single- and multi-root workspaces.

RomanNikitenko:
> I found that there is an ability in theia to set launch configurations on workspace level.
Also there is a PR under review in theia which should improve the behavior and fix some bugs related to running launch configs from workspace level. I'm testing that PR for the Che assembly based on the current PR to detect possible problems.

The result of testing/investigation is:
- `theia` really supports Workspace scoped launch configurations. At the moment the configurations should be placed inside of `settings` of workspace config file. There is a PR in `theia` which should align the behavior with VS Code (please see VS Code behavior [here]( https://code.visualstudio.com/docs/editor/multi-root-workspaces#_workspace-launch-configurations)).    
- moreover: `theia` supports User level launch configs
- the problem is: there are several bugs related to launch configs. 
- the most critical is - if user's project has `.theia/launch.json` or `.vscode/launch.json` with configs - `User` level and `Workspace` scoped configurations are not available for running - they are just not displayed in the `Debug` panel.
- another critical bug related to adding/updating/removing launch configurations using plugin API - at the moment it works incorrectly and leads to state that it's impossible to run a configuration provided through plugin API. It's critical for `che-theia` as launch related logic is placed on plugin side, so I was going to export launch configs from devfile on Workspace level using API. 
- minor bug - `User` level and `Workspace` scoped configurations are displayed in the `Debug` panel as `folder` scoped, not as `Workspace` or `User`, so they are duplicated in the dropdown list of the `Debug` panel for each project.

The good news: the bug related to API (adding/updating/removing launch configurations using plugin API) will be fixed by the mentioned above the PR in `theia` which I'm testing/reviewing. The PR also allows to place `Wrokspace` launch configs outside of settings of the workspace config file (the same as in VS Code).   

But there are still 2 bugs which I described above. I created the corresponding issues in `theia`.

RomanNikitenko:
[crw-ci-test --rebuild]

RomanNikitenko:
I have some difference in `Explorer` view for `Go` sample.
Please see my devfile for testing:
<details>

<summary>Devfile</summary>

```
apiVersion: 1.0.0
metadata:
  name: multi-root-go
projects:
  - name: example
    source:
      location: 'https://github.com/golang/example.git'
      type: git
    clonePath: src/github.com/golang/example/
  - name: golang-echo-realworld-example-app
    source:
      location: 'https://github.com/xesina/golang-echo-realworld-example-app.git'
      type: git
      branch: master
    clonePath: src/github.com/xesina/golang-echo-realworld-example-app
components:
  - id: golang/go/latest
    memoryLimit: 512Mi
    preferences:
      go.lintTool: golangci-lint
      go.useLanguageServer: true
      go.lintFlags: '--fast'
    type: chePlugin
    alias: go-plugin
    env:
      - value: 'off'
        name: GO111MODULE
  - mountSources: true
    endpoints:
      - name: 8080-tcp
        port: 8080
    memoryLimit: 2Gi
    type: dockerimage
    image: 'quay.io/eclipse/che-golang-1.14:7.22.0'
    alias: go-cli
    env:
      - value: $(CHE_PROJECTS_ROOT)
        name: GOPATH
      - value: /tmp/.cache
        name: GOCACHE
  - memoryLimit: 512Mi
    type: cheEditor
    reference: 'https://raw.githubusercontent.com/chepullreq4/pr-check-files/master/che-theia/pr-778/che_theia_meta.yaml'
    alias: che-theia
  - id: redhat/vscode-yaml/latest
    type: chePlugin
commands:
  - name: 1.1 Run outyet
    actions:
      - workdir: '${CHE_PROJECTS_ROOT}/src/github.com/golang/example/outyet'
        type: exec
        command: go get -d && go run main.go
        component: go-cli
  - name: 1.2 Stop outyet
    actions:
      - type: exec
        command: kill $(pidof go)
        component: go-cli
  - name: 1.3 Test outyet
    actions:
      - workdir: '${CHE_PROJECTS_ROOT}/src/github.com/golang/example/outyet'
        type: exec
        command: go test
        component: go-cli
  - name: '2.1 xenisa :: install dependencies'
    actions:
      - workdir: '${GOPATH}/src/github.com/xesina/golang-echo-realworld-example-app'
        type: exec
        command: go mod download
        component: go-cli
  - name: '2.2 xenisa :: run'
    actions:
      - workdir: '${GOPATH}/src/github.com/xesina/golang-echo-realworld-example-app'
        type: exec
        command: go run main.go
        component: go-cli
  - name: '2.3 xenisa :: build'
    actions:
      - workdir: '${GOPATH}/src/github.com/xesina/golang-echo-realworld-example-app'
        type: exec
        command: go build
        component: go-cli
  - name: '2.4 xenisa :: test'
    actions:
      - workdir: '${GOPATH}/src/github.com/xesina/golang-echo-realworld-example-app'
        type: exec
        command: go test ./...
        component: go-cli
  - name: Run current file
    actions:
      - workdir: '${fileDirname}'
        type: exec
        command: 'go get -d && go run ${file}'
        component: go-cli
  - name: Debug current file
    actions:
      - referenceContent: |
          {
            "version": "0.2.0",
            "configurations": [
              {
                "name": "Debug current file",
                "type": "go",
                "request": "launch",
                "mode": "auto",
                "program": "${fileDirname}"
              }
            ]
          }
        type: vscode-launch
```

</details>

The project structure for the `next` version of the editor looks like:

go-current

The project structure for the current PR's version of the editor looks like:
 
multi-root-go

@benoitf 
could you take a look ^^ - do you think it's expected structure for `Go` within a multi-root workspace?
Also I have to replace the value of `GO111MODULE` variable by `on` to avoid the errors which you can see in `Problems` view on the screenshot above.

thanks in advance for your opinion!

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
Looks like the tests were unhappy due to https://github.com/eclipse/che/issues/18876

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
The tests are unhappy due to https://github.com/eclipse/che/issues/19004

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
The tests are still unhappy due to https://github.com/eclipse/che/issues/19004
But I think we can start a process of review and manual testing
So I'm marking the PR as `ready for review`.

@vinokurig 
could you test some base `git` functionality for a multi-root project, please, when you have a chance.

@benoitf 
I see in the PR description:
> make it optional for now (like if there is a special attribute in devfile, use that mode)

could you share more details about it - do we have such special attribute in the `devfile`?

benoitf:
@RomanNikitenko there is no such attribute but it's if we want to revert per workspace this mode (stay as today without multi workspace root)

RomanNikitenko:
@benoitf 
> it's if we want to revert per workspace this mode (stay as today without multi workspace root)

yes, I see. I tired to add a filed to devfile as a marker - should we turn on `multi-root` mode or not
I got an error related to devfile schema validation.

So the current state of the PR - the multi-root mode is turned on by default.

benoitf:
@RomanNikitenko you tried with a devfile attribute ?

RomanNikitenko:
@benoitf 
no, just as additional field :-)

Tried as the following:
```
attributes:
  multi-root: 'on'
```
It works! Thanks!

OK, so we should provide some mechanism to have an ability switch `multi-root` mode

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
[crw-ci-test]

RomanNikitenko:
I provided a workaround on `Happy Path` tests side to avoid https://github.com/eclipse/che/issues/19004 and get check for my changes using the tests suits.
The workaround is: a delay before typing a text in an editor.

I'm going to remove the workaround as soon as QE team fixes the issue.

Anyway I think we can start a process of review and manual testing.

The current state of the PR - a workspace starts in `OFF` multi-root mode - so the behavior should be exactly as for `master` branch.
A user should add the following section to a `devfile` to turn on multi-root mode:
 ```
attributes:
  multiRoot: 'on'

 ```
 Please let me know if you have a better idea for the attribute and value as marker for multi-root mode - it's easy to change on the current step.

azatsarynnyy:
Thank you Roman for the update!
I plan to review/test it today.

benoitf:
we're at the beginning of the sprint so it's a good date to merge it

sunix:
Question: how do you deal with workspace settings defined in a devfile ?

sunix:
I don't want to be alarming but I cannot see a single line of test.

sunix:
> Question: how do you deal with workspace settings defined in a devfile ?
> Like https://github.com/eclipse/che-devfile-registry/blob/dcfdacf3954e30a25a8af3c7bd497b193d24f14a/devfiles/quarkus/devfile.yaml#L59-L62

maybe covered with https://github.com/eclipse/che-theia/blob/master/extensions/eclipse-che-theia-preferences-provider-extension/src/browser/prefs-provider.ts
It needs to be tested

RomanNikitenko:
> Question: how do you deal with workspace settings defined in a devfile ?
> Like https://github.com/eclipse/che-devfile-registry/blob/dcfdacf3954e30a25a8af3c7bd497b193d24f14a/devfiles/quarkus/devfile.yaml#L59-L62

I didn't provide anything special for this case within my PR, but I checked that with my changes the preference is available in `Workspace` scope:
multi-root-preferences

plugins/workspace-plugin/src/workspace-folder-updater.ts
```diff
@@ -0,0 +1,91 @@
...
+    return new Promise<void>((resolve, reject) => {
+      const disposable = theia.workspace.onDidChangeWorkspaceFolders(event => {
+        const existedWrokspaceFolders = theia.workspace.workspaceFolders;
+
+        const existedFoldersPaths = existedWrokspaceFolders
+          ? existedWrokspaceFolders.map(folder => folder.uri.path)
+          : [];
+        const addedFoldersPaths = event.added.map(folder => folder.uri.path);
+
+        if (addedFoldersPaths.includes(projectPath) || existedFoldersPaths.includes(projectPath)) {
```
tsmaeder, line 60: Why would we check the existing folders? The existing folders may be updated before the event is sent and we have no guarantee that processing is complete just because the folder is in "existingFolders". Why not just check for the added folders? Is the event not reliable?

plugins/workspace-plugin/src/workspace-folder-updater.ts
```diff
@@ -0,0 +1,91 @@
...
+ ***********************************************************************/
+
+import * as theia from '@theia/plugin';
+const UPDATE_WORKSPACE_FOLDER_TIMEOUT = 5000;
+
+export class WorkspaceFolderUpdater {
+  private pendingFolders: string[] = [];
+  private addingWorkspaceFolderPromise: Promise<void> | undefined;
+
+  async addWorkspaceFolder(path: string): Promise<void> {
```
tsmaeder, line 18: Maybe I'm missing somethings, but couldn't we replace this method with this:
```
  private addingWorkspaceFolderPromise: Promise<void> = Promise.resolve();

  async addWorkspaceFolder(path: string): Promise<void> {
    const workspaceFolderPath = this.toValidWorkspaceFolderPath(path);

    return this.addingWorkspaceFolderPromise.then(()=>this.addFolder(workspaceFolderPath));
  }
```

plugins/workspace-plugin/src/workspace-folder-updater.ts
```diff
@@ -0,0 +1,91 @@
...
+      const next = this.pendingFolders.shift();
+      if (next) {
+        this.addWorkspaceFolder(next);
+      }
+    }
+    return Promise.resolve();
+  }
+
+  protected addFolder(projectPath: string): Promise<void> {
+    const workspaceFolders = theia.workspace.workspaceFolders;
```
tsmaeder, line 45: Again, I think we could simplify this method like so:

```
    const isProjectFolder = (folder: theia.WorkspaceFolder) => folder.uri.path === projectPath;
    const workspaceFolders = theia.workspace.workspaceFolders || [];
    if (workspaceFolders.some(isProjectFolder)) {
      return Promise.resolve(undefined);
    }

    return new Promise<void>((resolve, reject) => {
      const disposable = theia.workspace.onDidChangeWorkspaceFolders(event => {

        if (event.added.some(isProjectFolder)) {
          clearTimeout(addFolderTimeout);

          disposable.dispose();

          resolve();
        }
      });

      const addFolderTimeout = setTimeout(() => {
        disposable.dispose();

        reject(
          new Error(
            `Adding workspace folder ${projectPath} was cancelled by timeout ${UPDATE_WORKSPACE_FOLDER_TIMEOUT} ms`
          )
        );
      }, UPDATE_WORKSPACE_FOLDER_TIMEOUT);

      theia.workspace.updateWorkspaceFolders(workspaceFolders ? workspaceFolders.length : 0, undefined, {
        uri: theia.Uri.file(projectPath),
      });
    });
```

plugins/workspace-plugin/src/workspace-folder-updater.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    if (this.addingWorkspaceFolderPromise) {
+      this.pendingFolders.push(workspaceFolderPath);
+    } else {
+      try {
+        this.addingWorkspaceFolderPromise = this.addFolder(workspaceFolderPath);
+        await this.addingWorkspaceFolderPromise;
+      } catch (error) {
+        console.error(error);
```
tsmaeder, line 31: Why do we need to handle this here and not let the error propagate?

plugins/workspace-plugin/src/workspace-folder-updater.ts
```diff
@@ -0,0 +1,91 @@
...
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ ***********************************************************************/
+
+import * as theia from '@theia/plugin';
+const UPDATE_WORKSPACE_FOLDER_TIMEOUT = 5000;
+
+export class WorkspaceFolderUpdater {
```
tsmaeder, line 14: All this code just because VS Code can't be bothered to return a Promise from updateWorkspaceFolders. Grr...

plugins/workspace-plugin/src/workspace-projects-manager.ts
```diff
@@ -33,37 +34,56 @@ export function handleWorkspaceProjects(pluginContext: theia.PluginContext, proj
   });
 }
 
+type Project = cheApi.workspace.devfile.Project | cheApi.workspace.ProjectConfig;
+
 abstract class WorkspaceProjectsManager {
```
tsmaeder, line 39: Do we seriously still support non-devfile workspaces?

plugins/workspace-plugin/src/workspace-projects-manager.ts
```diff
@@ -33,37 +34,56 @@ export function handleWorkspaceProjects(pluginContext: theia.PluginContext, proj
...
     }
-    const cloneCommandList = await this.selectProjectToCloneCommands(workspace);
-    await this.executeCloneCommands(cloneCommandList);
 
+    const cloningPromise = this.executeCloneCommands(workspace);
+
+    theia.window.withProgress({ location: { viewId: 'explorer' } }, () => cloningPromise);
+
+    await cloningPromise;
     await this.startSyncWorkspaceProjects();
```
tsmaeder, line 61: I don't thingk "startSyncWorkspaceProjects" is really async. Let's get rid of async/await.

tsmaeder (commented):
I ❤️ . As always, I have a couple of questions and suggestions.

azatsarynnyy (commented):
I've tested it with the different stacks.

All the basic functionality worked well:
- sparse checkout
- tasks running
- debugging
- importing the plug-ins` preferences

RomanNikitenko:
@azatsarynnyy @tsmaeder 
thank you very much for the reviewing/testing

I'm working on the addressing the comments, will update the PR ASAP.

plugins/workspace-plugin/src/workspace-projects-manager.ts
RomanNikitenko: I create the issue for the refactoring workspace creation related logic: https://github.com/eclipse/che/issues/19082

plugins/workspace-plugin/src/workspace-projects-manager.ts
RomanNikitenko: I create the issue for the refactoring workspace creation related logic: https://github.com/eclipse/che/issues/19082

RomanNikitenko: Please see https://github.com/eclipse/che/issues/15249#issuecomment-780369425

extensions/eclipse-che-theia-workspace/src/node/che-workspace-server.ts
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    const projectsRootEnvVariable = process.env.CHE_PROJECTS_ROOT;
+    const projectsRoot = projectsRootEnvVariable ? projectsRootEnvVariable : '/projects';
+
+    // first, check if we have a che.theia-workspace file
+    const cheTheiaWorkspaceFile = path.resolve(projectsRoot, 'che.theia-workspace');
+    const cheTheiaWorkspaceFileUri = FileUri.create(cheTheiaWorkspaceFile);
+    const exists = await fs.pathExists(cheTheiaWorkspaceFile);
+    if (exists) {
```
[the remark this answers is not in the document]
RomanNikitenko: Done, thanks!

extensions/eclipse-che-theia-workspace/src/browser/che-navigator-widget.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { injectable } from 'inversify';
+
+import { FileNavigatorWidget } from '@theia/navigator/lib/browser/navigator-widget';
+
+@injectable()
+export class CheFileNavigatorWidget extends FileNavigatorWidget {
+  protected renderEmptyMultiRootWorkspace(): React.ReactNode {
+    return (
+      <div className="theia-navigator-container">
+        <div className="center">Cloning projects....</div>
```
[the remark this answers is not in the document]
RomanNikitenko: Done, thanks!

plugins/task-plugin/src/export/export-configs-manager.ts
```diff
@@ -42,11 +43,21 @@ export class ExportConfigurationsManager {
...
+    this.launchConfigurationsExporter.init(this.cheCommands);
+    this.export();
+  }
+
   async export(): Promise<void> {
     const exportPromises = [];
-    const cheCommands = await this.cheWorkspaceClient.getCommands();
     for (const exporter of this.exporters) {
-      exportPromises.push(this.doExport(cheCommands, exporter));
+      exportPromises.push(this.doExport(this.cheCommands, exporter));
```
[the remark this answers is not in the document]
RomanNikitenko: `Che` commands are fetched at starting a workspace.
AFAIK there is no way to edit a `devfile` and apply the changes without a workspace restarting.
So - why we should do an extra request?

plugins/welcome-plugin/src/welcome-plugin.ts
```diff
@@ -122,18 +132,37 @@ export function start(context: theia.PluginContext): void {
...
-      if (workspacePlugin) {
-        workspacePlugin.exports.onDidCloneSources(() => handleReadmeFiles());
-      } else {
-        handleReadmeFiles();
-      }
-    }, 100);
+  addPanel(context);
+
+  let cloneSourcesDisposable: theia.Disposable | undefined = undefined;
+  const workspacePlugin = theia.plugins.getPlugin('Eclipse Che.@eclipse-che/workspace-plugin');
```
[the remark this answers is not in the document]
tsmaeder: In that case, we can remove the code once we've transitioned to "multi-root" only. I would add a comment saying that.

plugins/task-plugin/src/export/export-configs-manager.ts
```diff
@@ -42,11 +43,21 @@ export class ExportConfigurationsManager {
...
+
+  protected cheCommands: cheApi.workspace.Command[] = [];
+
+  async init(): Promise<void> {
+    this.cheCommands = await this.cheWorkspaceClient.getCommands();
+    this.launchConfigurationsExporter.init(this.cheCommands);
+    this.export();
+  }
+
   async export(): Promise<void> {
```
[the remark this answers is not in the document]
RomanNikitenko: I tried to apply it and found that:
- to do it fair enough I should provide the refactoring for all entities related to the export configurations process, so `interface ConfigurationsExporter`, `TaskConfigurationsExporter`, `LaunchConfigurationsExporter`, ...
- from the other hand the core point of these entities - export of configs from a `devfile` to the corresponding config file

So, unless you insist, I would leave it there.
thanks in advance!

plugins/task-plugin/src/export/launch-configs-exporter.ts
```diff
@@ -100,13 +116,32 @@ export class LaunchConfigurationsExporter implements ConfigurationsExporter {
...
+    const configDirPath = resolve(workspaceFolderPath, CONFIG_DIR);
+    await ensureDirExistenceAsync(configDirPath);
+
+    const launchConfigFilePath = resolve(configDirPath, LAUNCH_CONFIG_FILE);
+    await ensureDirExistenceAsync(launchConfigFilePath);
 
-  private saveConfigs(launchConfigFileUri: string, content: string, configurations: theia.DebugConfiguration[]): void {
     const result = modify(content, ['configurations'], configurations, formattingOptions);
-    writeFileSync(launchConfigFileUri, result);
+    return writeFile(launchConfigFilePath, result);
```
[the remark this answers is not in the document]
RomanNikitenko: Done - added `return this.saveConfigs()` to `doExport()`
thanks!

plugins/task-plugin/src/utils.ts
```diff
@@ -116,3 +134,12 @@ export function ensureDirExistence(filePath: string): void {
   }
   fs.mkdirSync(dirName, { recursive: true });
 }
+
+/** Creates a directory to the file if they don't exist */
```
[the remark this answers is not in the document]
RomanNikitenko: Changed, the new comment is: `Creates a directory containing the file if they don't exist`

plugins/task-plugin/src/utils.ts
```diff
@@ -116,3 +134,12 @@ export function ensureDirExistence(filePath: string): void {
...
+export async function ensureDirExistenceAsync(filePath: string): Promise<void> {
```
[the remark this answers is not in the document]
RomanNikitenko: Renamed to `ensureDirExists`

plugins/welcome-plugin/src/welcome-plugin.ts
```diff
@@ -53,23 +53,33 @@ async function getHtmlForWebview(context: theia.PluginContext): Promise<string>
...
-  // In case of only one workspace
-  if (roots && roots.length === 1) {
-    const children = await theia.workspace.findFiles('README.md', 'node_modules/**', 1);
-    const updatedChildren = children.filter((child: theia.Uri) => {
-      if (child.fsPath.indexOf('node_modules') === -1) {
-        return child;
-      }
-    });
+export async function handleReadmeFiles(
+  openReadmeHandler?: () => void,
```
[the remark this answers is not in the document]
RomanNikitenko: Renamed by `readmeHandledCallback`

plugins/welcome-plugin/src/welcome-plugin.ts
```diff
@@ -122,18 +132,37 @@ export function start(context: theia.PluginContext): void {
...
-      } else {
-        handleReadmeFiles();
-      }
-    }, 100);
+  addPanel(context);
+
+  let cloneSourcesDisposable: theia.Disposable | undefined = undefined;
+  const workspacePlugin = theia.plugins.getPlugin('Eclipse Che.@eclipse-che/workspace-plugin');
+  if (workspacePlugin) {
+    cloneSourcesDisposable = workspacePlugin.exports.onDidCloneSources(
```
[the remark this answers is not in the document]
RomanNikitenko: Actually it's not enough to `invoke once`.
It should dispose both events when we find and open a readme file.
So, if an event is fired, but a readme file is NOT found - we continue to listen events - the next event can bring a workspace folder with a readme file.

extensions/eclipse-che-theia-plugin-ext/src/browser/che-debug-configuration-manager.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { DebugConfigurationManager } from '@theia/debug/lib/browser/debug-configuration-manager';
+
+@injectable()
+export class CheDebugConfigurationManager extends DebugConfigurationManager {
+  @postConstruct()
+  protected async init(): Promise<void> {
+    super.init();
+    this.workspaceService.onWorkspaceChanged(() => {
+      this.updateModels();
```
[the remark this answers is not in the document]
tsmaeder: Don't tell me, put it in a comment ;-)

dmytro-ndp:
[crw-ci-test]

RomanNikitenko:
@dmytro-ndp 
I see that you rerun the tests for my PR, just FYI:

- the tests are run from the custom branch https://github.com/eclipse/che/compare/devfile-workspace-root
- I'm going to update it today to remove my workaround and pick up your fix.

dmytro-ndp:
@RomanNikitenko: thanks for information. I am stopping Happy path test execution then.

plugins/welcome-plugin/src/welcome-plugin.ts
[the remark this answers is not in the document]
tsmaeder: So whichever event happens first cancels execution of the other event handler as well?

plugins/workspace-plugin/src/theia-commands.ts
```diff
@@ -44,6 +44,7 @@ function isDevfileProjectConfig(
 
 export interface TheiaImportCommand {
   execute(): PromiseLike<void>;
+  getProjectPath(): string;
```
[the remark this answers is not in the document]
RomanNikitenko: I refactored the code according to your comment: https://github.com/eclipse/che-theia/pull/778#discussion_r576204400
So, the method was removed within the refactoring.
thanks!

plugins/workspace-plugin/src/theia-commands.ts
```diff
@@ -235,6 +240,10 @@ export class TheiaImportZipCommand implements TheiaImportCommand {
     }
   }
 
+  getProjectPath(): string {
+    return this.projectDir;
```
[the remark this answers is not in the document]
RomanNikitenko: I refactored the code according to your comment: https://github.com/eclipse/che-theia/pull/778#discussion_r576204400
So, the method was removed within the refactoring.
thanks!

plugins/workspace-plugin/src/theia-commands.ts
```diff
@@ -235,6 +240,10 @@ export class TheiaImportZipCommand implements TheiaImportCommand {
...
+  }
+
   execute(): PromiseLike<void> {
```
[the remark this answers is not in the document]
RomanNikitenko: I refactored the code according to your suggestion.
thanks!

extensions/eclipse-che-theia-plugin-ext/src/browser/che-debug-configuration-manager.ts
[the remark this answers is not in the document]
RomanNikitenko: Added the comment

plugins/workspace-plugin/src/workspace-folder-updater.ts
RomanNikitenko: it's because `updateWorkspaceFolders()` is [too tricky](https://github.com/eclipse-theia/theia/blob/4e1c1d429226280e6f0eff49658c733a38b50c4a/packages/plugin/src/theia.d.ts#L5428). I played a lot with it to avoid bugs within working on the issue...

plugins/workspace-plugin/src/workspace-folder-updater.ts
RomanNikitenko: Let's consider the case - we subscribed on the event too late - the folder was already added
If we don't check `existingWorkspaceFolders` - we never find the folder among `added` folders as we already missed the corresponding event. We can find it only among `existingWorkspaceFolders`.  

```
existingWorkspaceFolders = theia.workspace.workspaceFolders
```
so - we should do nothing - the folder was already added

plugins/workspace-plugin/src/workspace-folder-updater.ts
RomanNikitenko: `updateWorkspaceFolders()` is [too tricky](https://github.com/eclipse-theia/theia/blob/4e1c1d429226280e6f0eff49658c733a38b50c4a/packages/plugin/src/theia.d.ts#L5428).

Adding a workspace folder using API without waiting for `onDidChangeWorkspaceFolders` leads to bugs - I played a lot with it and faced with bugs related to `Project tree` and so on.

Let's consider the case:
- we are adding a workspace folder `111` using API
- at the same time a request for adding `222` is coming`
- 300 milliseconds later - request to add `333`
- 500 milliseconds later - request to add `444`

The API works well if we add folders step by step, waiting for `onDidChangeWorkspaceFolders` is the required condition for each step.

RomanNikitenko: We could avoid all this complex logic by adding all workspace folders in one API request when all projects have cloned.
But it leads to bad UX - `Project tree` will be empty all time while projects are cloned. It's too critical when at least one project is big enough and cloning process takes a lot of time...

plugins/workspace-plugin/src/workspace-folder-updater.ts
RomanNikitenko: If a request for adding a folder failed - we should continue to add other folders.

plugins/workspace-plugin/src/workspace-folder-updater.ts
RomanNikitenko: Applied your approach as close as possible - thank you - it really does the code more readable.

plugins/workspace-plugin/src/workspace-folder-updater.ts
```diff
@@ -0,0 +1,91 @@
...
+    const workspaceFoldersPaths = workspaceFolders ? workspaceFolders.map(folder => folder.uri.path) : [];
+    if (workspaceFoldersPaths.includes(projectPath)) {
+      return Promise.resolve(undefined);
+    }
+
+    return new Promise<void>((resolve, reject) => {
+      const disposable = theia.workspace.onDidChangeWorkspaceFolders(event => {
+        const existedWrokspaceFolders = theia.workspace.workspaceFolders;
+
+        const existedFoldersPaths = existedWrokspaceFolders
```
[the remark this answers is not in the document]
RomanNikitenko: fixed, thanks!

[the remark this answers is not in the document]
RomanNikitenko: fixed

plugins/workspace-plugin/src/workspace-folder-updater.ts
```diff
@@ -0,0 +1,91 @@
...
+  protected addFolder(projectPath: string): Promise<void> {
+    const workspaceFolders = theia.workspace.workspaceFolders;
+    const workspaceFoldersPaths = workspaceFolders ? workspaceFolders.map(folder => folder.uri.path) : [];
+    if (workspaceFoldersPaths.includes(projectPath)) {
+      return Promise.resolve(undefined);
+    }
+
+    return new Promise<void>((resolve, reject) => {
+      const disposable = theia.workspace.onDidChangeWorkspaceFolders(event => {
+        const existedWrokspaceFolders = theia.workspace.workspaceFolders;
```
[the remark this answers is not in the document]
RomanNikitenko: Done!

plugins/welcome-plugin/src/welcome-plugin.ts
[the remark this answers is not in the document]
RomanNikitenko: true with one remark:
> if we opened a readme file

azatsarynnyy (approved):
After several rounds of testing, I don't see any issues anymore.
I think it's ready now and all other enhancements we could make in further iterations.

plugins/welcome-plugin/src/welcome-plugin.ts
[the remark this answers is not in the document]
tsmaeder: We could do something like 
```
function onlyOnce(event, handler, previous?) {
      toDispose= event((t)=> {
        toDispose.dispose();
        if (previous) {
           previous.dispose()
        }
        handler(t);
     }
  }
}
```

But that wouldn't really be simpler anymore. Resolving.

plugins/workspace-plugin/src/workspace-folder-updater.ts
tsmaeder: But we would only start executing the addFolder promise for `222` once the promise for '111` is resolved. Doing 
```
await promise;
doStuff();
```
is strictl the same as doing 
```
promise.then(()=>doStuff()),
```

RomanNikitenko: Sorry, I didn't catch the idea - what about `333` and `444` at the same time?

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-init.ts
```diff
@@ -46,12 +46,17 @@ import pluginExtBackendModule from '@theia/plugin-ext/lib/plugin-ext-backend-mod
...
+const DEFAULT_HOME_DIR = '/home/theia/';
+
 interface CheckAliveWS extends ws {
   alive: boolean;
 }
 
-function modifyPathToLocal(origPath: string): string {
-  return path.join(os.homedir(), origPath.substr(0, '/home/theia/'.length));
+function modifyPathToLocal(originalPath: string): string {
+  if (originalPath.startsWith(DEFAULT_HOME_DIR)) {
```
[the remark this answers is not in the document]
RomanNikitenko: I added handling of `/home/theia-dev` case, please tale a look the changes https://github.com/eclipse/che-theia/pull/778/commits/ecbb10b6edf31f345b44740d8ffa72ae7ee80ce2
Does it look good from your point of view? 
thanks!

RomanNikitenko:
I have rebased the branch and added the handling of `/home/theia-dev` case according to https://github.com/eclipse/che-theia/pull/778#discussion_r576090139

@tsmaeder 
please take a look at the changes https://github.com/eclipse/che-theia/pull/778/commits/ecbb10b6edf31f345b44740d8ffa72ae7ee80ce2

thanks in advance!

benoitf:
I can't approve the PR as I created it :-)

azatsarynnyy:
> I can't approve the PR as I created it :-)

@benoitf We'll take this ^ as your approvement :smile:

RomanNikitenko:
I resolved the conflicts (the changes for some files were too close to the merged recently PR) and rebased my branch.
I'm going to merge the changes tomorrow morning if there are no objections and the tests are happy.

plugins/workspace-plugin/src/workspace-folder-updater.ts
tsmaeder: Yes, just ranting ;-)

plugins/workspace-plugin/src/workspace-folder-updater.ts
tsmaeder: You're trying to delay the processing of folder 222 until after the processing of folder 111 has finished, right? But if you do 

processFolder111Promise.then(processFolder222Promise) that is exactly the same as when you write

```
await processFolder111Promise;
startProcessFolder222Promise;
```
It literally compiles to the same javascript. The try catch can also be handled like so: https://stackoverflow.com/questions/36596925/promise-ignore-catch-and-return-to-chain

In the end, both versions work, just that mine would be easier to understand and shorter, IMO.

RomanNikitenko: I see, thank you!
Yes, I understand that I could use something like `promise.then(()=>doStuff()),`

I thought your way allows to avoid keeping a set with pending folders for adding...
So, sorry - first I didn't catch your idea. Now I see what it was about...

tsmaeder: I think you acutally _can_ get rid of that set: just check the workspaces array whether the folder you're adding is in it before executing the `updateWorkspace`. The code is executed only a couple of times and the workspace folder is usually 2-3 entries at most, so I would choose clarity over performance here.

tsmaeder: You might have a couple of extra thenables on the stack, but who cares.

PR #781, opened 2020-06-30
Support proxy configuration for tasks related functionality

RomanNikitenko:

### What does this PR do?
Support proxy configuration for tasks related functionality

### DIFF
--- a/plugins/task-plugin/src/machine/websocket.ts
+++ b/plugins/task-plugin/src/machine/websocket.ts
@@ -10,6 +10,9 @@
 
 import * as WS from 'ws';
 import * as fs from 'fs';
+import * as tunnel from 'tunnel';
+import * as url from 'url';
+import * as https from 'https';
 import { IWebSocket, ConsoleLogger, createWebSocketConnection, Logger, MessageConnection } from 'vscode-ws-jsonrpc';
 
 const SS_CRT_PATH = '/tmp/che/secret/ca.crt';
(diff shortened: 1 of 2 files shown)


azatsarynnyy:
After merging this PR we need to backport it to master and to 7.15.x

davidfestal:
Maybe this code should be extracted in a common component and reused for both the various places where it is used (http client in `che-workspace-client` for example) ? This would avoid some code duplication.
I'm not saying that for this PR of course, but maybe it could be planned for on master ?
wdyt @azatsarynnyy ?

RomanNikitenko:
@davidfestal helped me to do testing the changes - thanks a lot!

Looks like the changes fixes the issue.
I'm going to do more testing tomorrow morning.

azatsarynnyy:
> Maybe this code should be extracted in a common component and reused for both the various places where it is used (http client in `che-workspace-client` for example) ? This would avoid some code duplication.
> I'm not saying that for this PR of course, but maybe it could be planned for on master ?
> wdyt @azatsarynnyy ?

@davidfestal definitely, it's a good idea. We'll merge the patch into 7.14.x, 7.15.x, and master.
Then we'll create an issue for extracting the proxy related functionality into a separate library/project. To be able to reuse it in both Extensions and Plugins.

plugins/task-plugin/src/machine/websocket.ts
```diff
@@ -28,22 +31,102 @@ export class ReconnectingWebSocket {
...
+                const httpOverHttpsAgent = tunnel.httpOverHttps({ proxy: httpsProxyOptions });
+                const httpsOverHttpAgent = tunnel.httpsOverHttp({
+                    proxy: mainProxyOptions,
+                    ca: certificateAuthority ? [certificateAuthority] : undefined
+                });
+                const httpsOverHttpsAgent = tunnel.httpsOverHttps({
+                    proxy: httpsProxyOptions,
+                    ca: certificateAuthority ? [certificateAuthority] : undefined
+                });
+                const urlIsHttps = (parsedBaseUrl.protocol || 'http:').startsWith('https:');
```
tsmaeder, line 63: parsedUrl.protocol && parsedUrl.startsWith('https') would be easier to understand, IMO.

plugins/task-plugin/src/machine/websocket.ts
```diff
@@ -28,22 +31,102 @@ export class ReconnectingWebSocket {
...
+            const parsedBaseUrl = url.parse(baseUrl);
+            if (parsedBaseUrl.hostname && this.shouldProxy(parsedBaseUrl.hostname)) {
+                const parsedProxyUrl = url.parse(proxyUrl);
+                const mainProxyOptions = this.getMainProxyOptions(parsedProxyUrl);
+                const httpsProxyOptions = this.getHttpsProxyOptions(mainProxyOptions, parsedBaseUrl.hostname, certificateAuthority);
+                const httpOverHttpAgent = tunnel.httpOverHttp({ proxy: mainProxyOptions });
+                const httpOverHttpsAgent = tunnel.httpOverHttps({ proxy: httpsProxyOptions });
+                const httpsOverHttpAgent = tunnel.httpsOverHttp({
+                    proxy: mainProxyOptions,
+                    ca: certificateAuthority ? [certificateAuthority] : undefined
```
tsmaeder, line 57: If we defined `getCertificateAuthority()` to return `Buffer[] | undefined`, we wouldn't have to write the conditional expression at each use.

plugins/task-plugin/src/machine/websocket.ts
```diff
@@ -28,22 +31,102 @@ export class ReconnectingWebSocket {
...
+                    ca: certificateAuthority
+                })
+            };
         }
 
         return options;
     }
 
+    private getCertificateAuthority(): Buffer | undefined {
+        let certificateAuthority: Buffer | undefined;
```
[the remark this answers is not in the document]
tsmaeder: Together with the comment about the `certificateAuthority` variable above, I would rewrite this method as 
```
    private getCertificateAuthority(): Buffer[] | undefined {
        if (fs.existsSync(SS_CRT_PATH)) {
            return [fs.readFileSync(SS_CRT_PATH)];
        }
        return undefined;
}
```

plugins/task-plugin/src/machine/websocket.ts
```diff
@@ -28,22 +31,102 @@ export class ReconnectingWebSocket {
...
+        if (fs.existsSync(SS_CRT_PATH)) {
+            certificateAuthority = fs.readFileSync(SS_CRT_PATH);
+        }
+        return certificateAuthority;
+    }
+
+    private shouldProxy(hostname: string): boolean {
+        const noProxyEnv = process.env.no_proxy || process.env.NO_PROXY;
+        const noProxy: string[] = noProxyEnv ? noProxyEnv.split(',').map(s => s.trim()) : [];
```
tsmaeder, line 95: Maybe simpler like this
```
        const noProxyEnv = process.env.no_proxy || process.env.NO_PROXY || '';
        const noProxy: string[] = noProxyEnv.split(',').map(s => s.trim());
```

tsmaeder:
If I understand this code correctly, it is using the external URL of the machine exec service. But the machine exec service is running in the same pod, so should be available locally, no? Is there a way we could shortcut going through the proxy?
Also, since we're in a plugin here, why not make executing commands available as part of the che-theia plugin API?
Furthermore, I believe this whole proxying problem would go away if we moved task execution back to being an extension, as is planned, I believe.

RomanNikitenko:
> Also, since we're in a plugin here, why not make executing commands available as part of the che-theia plugin API?
Furthermore, I believe this whole proxying problem would go away if we moved task execution back to being an extension, as is planned, I believe.

The best way from my point of view is:
>  if we moved task execution back to being an extension, as is planned
if it solves the problem.

plugins/task-plugin/src/machine/websocket.ts
```diff
@@ -28,22 +31,102 @@ export class ReconnectingWebSocket {
...
+    private getCertificateAuthority(): Buffer | undefined {
+        let certificateAuthority: Buffer | undefined;
+        if (fs.existsSync(SS_CRT_PATH)) {
+            certificateAuthority = fs.readFileSync(SS_CRT_PATH);
+        }
+        return certificateAuthority;
+    }
+
+    private shouldProxy(hostname: string): boolean {
+        const noProxyEnv = process.env.no_proxy || process.env.NO_PROXY;
```
benoitf, line 94: out of curiosity, I thought that only `NO_PROXY` is a env variable used, not `no_proxy`

azatsarynnyy:
@RomanNikitenko let's focus on testing and merging this PR ASAP if no critical comments within the reviews.
And address the review comments in the PR https://github.com/eclipse/che-theia/pull/782 for `master` branch.

plugins/task-plugin/src/machine/websocket.ts
RomanNikitenko: @benoitf I took it [here](https://github.com/eclipse/che-workspace-client/blob/dce065248ae382d7baefd570d9937befc5cdd1c3/src/index.ts#L154)

azatsarynnyy:
I've verified it on CRC without a proxy and I see no regression.

Checked:
- running a task from Command Palette
- running a task from My Workspace panel
- terminating the running task
- restarting the already running task

image

RomanNikitenko:
Tested for proxy related environment
tasks_testing_proxy

RomanNikitenko:
@tsmaeder 
The PR is for `7.14.x` branch.
I'm going to merge the PR as is and apply your comments for `master` branch https://github.com/eclipse/che-theia/pull/782

Is it OK for you?

RomanNikitenko:
Tested on https://che.openshift.io/

openshift_tests_tasks

tsmaeder:
Sure.

RomanNikitenko:
@tsmaeder 
thank you!

PR #789, opened 2020-07-02
Provide ability to call filtered workspace service methods using user token

vzhukovskii:
### What does this PR do?
This changes proposal adds an ability to call workspace service methods using user access token. Some of methods due to restriction, requires keycloak token, not machine's one.

https://github.com/eclipse/che/issues/17246

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/dockerfiles/theia/e2e/test.sh
+++ b/dockerfiles/theia/e2e/test.sh
@@ -23,6 +23,10 @@ run_test_in_docker_container() {
        -v "${base_dir}/videos":/root/cypress/videos \
        -v "${base_dir}/logs":/root/logs \
        -v /var/run/docker.sock:/var/run/docker.sock \
+       -e CHE_API_INTERNAL=http://127.0.0.1/api \
+       -e CHE_MACHINE_TOKEN=foobar \
+       -e CHE_WORKSPACE_ID=workspaceId \
+       -e CHE_WORKSPACE_TELEMETRY_BACKEND_PORT=4167 \
            $IMAGE_NAME
 }
 
--- a/extensions/che-theia-hosted-plugin-manager-extension/package.json
+++ b/extensions/che-theia-hosted-plugin-manager-extension/package.json
@@ -12,7 +12,7 @@
     "@theia/core": "next",
     "@theia/plugin-dev": "next",
     "@theia/plugin-ext": "next",
-    "@eclipse-che/workspace-client": "0.0.1-1585913592",
+    "@eclipse-che/workspace-client": "latest",
     "@eclipse-che/api": "latest"
   },
   "scripts": {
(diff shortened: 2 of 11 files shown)


extensions/eclipse-che-theia-plugin-ext/src/browser/oauth-utils.ts
```diff
@@ -65,16 +65,34 @@ export class OauthUtils {
     }
 
     async getToken(oAuthProvider: string): Promise<string | undefined> {
-        return await this.cheApiService.getOAuthToken(oAuthProvider, await this.getUserToken());
+        const token = await this.getUserToken();
+
+        if (token === undefined) {
+            return undefined;
```
vinokurig, line 71: We can't return `undefined` if token is missing. This can happen In case of the single-user mode. `che-workspace-client` handles this case by itself: https://github.com/eclipse/che-workspace-client/blob/dfaf9f1c543eb5ddb43146c7c1248187328d8030/src/rest/resources.ts#L302-L305

vzhukovskii: che-api-service handles what type of token should we provide for making requests: https://github.com/eclipse/che-theia/blob/80548d47b83a073e71128b80cd51d47009a3f1fd/extensions/eclipse-che-theia-plugin-ext/src/node/che-api-service.ts#L233-L234

And AFAIK, in single-user mode this method should return keycloak token most of the time.

vzhukovskii: And on the other hand, there are bunch of methods in che-api-service, that won't work with machine token, such as `getAll` and `getAllByNamespace`, so, anyway they should have user token. When oauth-utils doesn't provide user token, we'll know, that something happened on obtaining token step.

vzhukovskii: We can throw an Error object here, for example.

vinokurig: > And AFAIK, in single-user mode this method should return keycloak token most of the time.

This function `getToken(oAuthProvider: string)` returns the token of the specified oAuth provider (GitHub, OpenShift ...). 
> When oauth-utils doesn't provide user token, we'll know, that something happened on obtaining token step. We can throw an Error object here, for example. Then we'll be confident, that we'll have token anyway.

It should be done in the step of obtaining the user token.

vzhukovskii: So, `getToken(oAuthProvider: string)` signature can be modified to return only `string` value, w/o `undefined` value.

> It should be done in the step of obtaining the user token.

Make sense, I'll update the PR.

vinokurig: > So, `getToken(oAuthProvider: string`) signature can be modified to return only `string` value, w/o `undefined` value.

agree

vinokurig: Sorry, I forgot, we need `undefined` in case of the single-user

vzhukovskii: Code reverted.

vzhukovskii:
Checked on single and multi-user instances.

PR #794, opened 2020-07-06
Handle more exported properties from remote plugins

benoitf:
### What does this PR do?
Handle more properties from the exports
also handle constants directly (as some plugin's are checking if the returned value is a string and not a function that might return a string)

Checked with https://che.openshift.io/f?url=https://gist.githubusercontent.com/benoitf/1fcff181b08448a96ee186bec3b5222f/raw/f4bbe2febfbd89a32dcfdf6eff470f674be7010c/devfile-16589.yaml

image

Part of https://github.com/eclipse/che/issues/16589

Change-Id: Icbb026f65577422168f1acab486e598c6146c170

### DIFF
--- a/extensions/eclipse-che-theia-plugin-remote/src/common/plugin-remote-rpc.ts
+++ b/extensions/eclipse-che-theia-plugin-remote/src/common/plugin-remote-rpc.ts
@@ -11,6 +11,12 @@
 import { createProxyIdentifier } from '@theia/plugin-ext/lib/common/rpc-protocol';
 import { DeployedPlugin, ConfigStorage, PluginPackage } from '@theia/plugin-ext/lib/common';
 
+export interface ProxyNameDefinition {
+    name: string;
+    value?: string;
+    type: string;
+}
+
 export interface PluginRemoteNode {
     $initExternalPlugins(externalPlugins: DeployedPlugin[]): Promise<void>;
     $loadPlugin(pluginId: string, configStorage: ConfigStorage): Promise<void>;
@@ -19,7 +25,7 @@ export interface PluginRemoteNode {
     $callLocalMethod(callId: number, index: number, ...args: any[]): Promise<any>;
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     $callMethod(fromHostId: string, pluginId: string, callId: number, entryName: string, ...args: any[]): Promise<any>;
-    $definePluginExports(hostId: string, pluginId: string, proxyNames: string[]): Promise<void>;
+    $definePluginExports(hostId: string, pluginId: string, proxyNames: ProxyNameDefinition[]): Promise<void>;
     $definePluginPackage(pluginId: string, rawModel: PluginPackage): Promise<void>;
 }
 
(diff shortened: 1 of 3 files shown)


tsmaeder:
I think there is a problem here: I set up a debug environment according to https://github.com/eclipse/che-theia/blob/master/CONTRIBUTING.md#how-to-develop-che-theia-remote-plugin-mechanism

1. I built the sonarlint plugin from source and added that to plugins folder of the remote plugin host
2. Added a breakpoint in java.ts#installClasspathListener(...)
3. The breakpoint was hit
4. Stepped over the line containing `VSCode.extensions.getExtension('redhat.java');`
5. Observe: the result is `undefined`

So the classpath listener is never installed.

tsmaeder:
FWIW: my devfile:

```
metadata:
  name: che-theia-remote-dev
projects:
  - name: theia
    source:
      location: 'https://github.com/eclipse-theia/theia'
      type: git
      branch: master
  - name: che-theia
    source:
      location: 'https://github.com/benoitf/che-theia'
      type: git
      branch: CHE-16589
    clonePath: theia/che/che-theia
components:
  - id: eclipse/che-theia/next
    type: cheEditor
    alias: theia-editor
  - id: che-incubator/typescript/latest
    memoryLimit: 2Gi
    type: chePlugin
  - mountSources: true
    endpoints:
      - name: theia-dev-flow
        port: 3010
        attributes:
          protocol: http
          public: 'true'
    memoryLimit: 3Gi
    type: dockerimage
    image: 'quay.io/eclipse/che-theia-dev:next'
    alias: theia-dev
    env:
      - value: '2504'
        name: THEIA_PLUGIN_ENDPOINT_DISCOVERY_PORT
  - mountSources: true
    memoryLimit: 1Gi
    type: dockerimage
    image: 'quay.io/eclipse/che-theia-dev:next'
    alias: remote-1
    env:
      - value: '2504'
        name: THEIA_PLUGIN_ENDPOINT_DISCOVERY_PORT
      - value: 'local-dir:///projects/plugins-1/'
        name: THEIA_PLUGINS
  - mountSources: true
    memoryLimit: 1Gi
    type: dockerimage
    image: 'quay.io/eclipse/che-theia-dev:next'
    alias: sonar-plugin
    env:
      - value: '2504'
        name: THEIA_PLUGIN_ENDPOINT_DISCOVERY_PORT
      - value: 'local-dir:///projects/plugins-sonar/'
        name: THEIA_PLUGINS
  - type: chePlugin
    reference: >-
      https://raw.githubusercontent.com/eclipse/che-plugin-registry/master/v3/plugins/redhat/java11/0.63.0/meta.yaml
    alias: java
  - type: chePlugin
    reference: >-
      https://raw.githubusercontent.com/eclipse/che-plugin-registry/master/v3/plugins/ms-vscode/node-debug2/1.42.3/meta.yaml
    alias: node-debug
apiVersion: 1.0.0
```

tsmaeder:
@benoitf what is your procedure for verifying that the sonarlint plugin works correctly with Java stuff?

tsmaeder:
Ok, I need to add the Java plugin to the target system...stand by.

benoitf:
in that specific case, I added explicit dependencies to redhat.java in `extensionDependencies` of package.json of sonarsource

without that patch, it fails to see all stuff from `redhat.java`

tsmaeder:
In the spirit of compatiblity with vs code extensions, I think we should not rely on the "extensionDependencies" field in the package.json for "getExtension(...)" to work. Also, since it works when running in the same container, I would argue that the extension API should be as much location transparent as possible.

benoitf:
I agree that we should make it work but at some point but there is a real need to have optional dependencies marked in package.json https://github.com/Microsoft/vscode/issues/6384

Because in activation events, it needs to be ordered

BTW this PR was just fixing wrong exported elements, not addressing the optional/required extensions issue

PR #800, opened 2020-07-09
Show recent Che workspaces in Theia editor

vzhukovskii:
### What does this PR do?
This feature request provides an ability to switch to another Che workspace directly from Theia editor.

Go to `File` menu and choose `Open Recent Workspace...`

Снимок экрана 2020-07-13 в 18 50 16

It will open workspace list, sorted by modification time:

Снимок экрана 2020-07-13 в 18 49 30

Then developer will be notified if current workspace should be stopped:

Снимок экрана 2020-07-13 в 18 53 44

https://github.com/eclipse/che/issues/17034

#### Release Notes
Show recent Che workspaces in Theia editor.

### DIFF
--- a/extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
+++ b/extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
@@ -1,0 +1,203 @@
+/*********************************************************************
+ * Copyright (c) 2019 Red Hat, Inc.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ **********************************************************************/
+import { inject, injectable } from 'inversify';
+import {
+    QuickOpenGroupItem,
+    QuickOpenItem,
+    QuickOpenMode,
+    QuickOpenModel
+} from '@theia/core/lib/common/quick-open-model';
+import { LabelProvider, QuickOpenService } from '@theia/core/lib/browser';
+import { CheApiService } from '@eclipse-che/theia-plugin-ext/lib/common/che-protocol';
+import { che as cheApi } from '@eclipse-che/api';
+import { OauthUtils } from '@eclipse-che/theia-plugin-ext/lib/browser/oauth-utils';
+import { AbstractDialog } from '@theia/core/lib/browser/dialogs';
+import { MessageService } from '@theia/core/lib/common/message-service';
+import * as moment from 'moment';
+import { Message } from '@theia/core/lib/browser/widgets/index';
+import { Key } from '@theia/core/lib/browser/keyboard/keys';
+
+@injectable()
+export class QuickOpenCheWorkspace implements QuickOpenModel {
+    protected items: QuickOpenGroupItem[];
+    protected currentWorkspace: cheApi.workspace.Workspace;
+
+    @inject(QuickOpenService) protected readonly quickOpenService: QuickOpenService;
+    @inject(CheApiService) protected readonly cheApi: CheApiService;
+    @inject(OauthUtils) protected readonly oAuthUtils: OauthUtils;
+    @inject(LabelProvider) protected readonly labelProvider: LabelProvider;
+    @inject(MessageService) protected readonly messageService: MessageService;
+
+    private async open(workspaces: cheApi.workspace.Workspace[]): Promise<void> {
+        this.items = [];
+
+        if (!workspaces.length) {
+            this.items.push(new QuickOpenGroupItem({
+                label: 'No Recent Workspaces',
+                run: (mode: QuickOpenMode): boolean => false
+            }));
+            return;
+        }
+
+        for (const workspace of workspaces) {
+            const icon = this.labelProvider.folderIcon;
+            const iconClass = icon + ' file-icon';
+            this.items.push(new QuickOpenGroupItem({
+                label: this.getWorkspaceName(workspace) + (this.isCurrentWorkspace(workspace) ? ' (Current)' : ''),
+                detail: `Stack: ${this.getWorkspaceStack(workspace)}`,
+                groupLabel: `last modified ${moment(this.getWorkspaceModificationTime(workspace)).fromNow()}`,
+                iconClass,
+                run: (mode: QuickOpenMode): boolean => {
+                    if (mode !== QuickOpenMode.OPEN) {
+                        return false;
+                    }
+
+                    if (this.isCurrentWorkspace(workspace)) {
+                        return true;
+                    }
+
+                    this.openWorkspace(workspace);
+
+                    return true;
+                },
+            }));
+        }
+
+        this.quickOpenService.open(this, {
+            placeholder: 'Type the name of the Che workspace you want to open',
+            fuzzyMatchLabel: true,
+            fuzzySort: false
+        });
+    }
+
+    onType(lookFor: string, acceptor: (items: QuickOpenItem[]) => void): void {
+        acceptor(this.items);
+    }
+
+    async select(): Promise<void> {
+        this.items = [];
+
+        const token = await this.oAuthUtils.getUserToken();
+
+        if (!this.currentWorkspace) {
+            this.currentWorkspace = await this.cheApi.currentWorkspace();
+        }
+
+        if (!this.currentWorkspace.namespace) {
+            return;
+        }
+
+        const workspaces = await this.cheApi.getAllByNamespace(this.currentWorkspace.namespace, token);
+
+        workspaces.sort((a: cheApi.workspace.Workspace, b: cheApi.workspace.Workspace) => {
+            const updatedA: number = this.getWorkspaceModificationTime(a);
+            const updatedB: number = this.getWorkspaceModificationTime(b);
+
+            if (isNaN(updatedA) || isNaN(updatedB)) {
+                return 0;
+            } else {
+                return updatedB - updatedA;
+            }
+        });
+
+        await this.open(workspaces);
+    }
+
+    private getWorkspaceName(workspace: cheApi.workspace.Workspace): string | undefined {
+        if (workspace.devfile && workspace.devfile.metadata) {
+            return workspace.devfile.metadata.name;
+        }
+    }
+
+    private getWorkspaceStack(workspace: cheApi.workspace.Workspace): string | undefined {
+        return workspace.attributes && workspace.attributes.stackName ? workspace.attributes.stackName : 'Custom';
+    }
+
+    private getWorkspaceModificationTime(workspace: cheApi.workspace.Workspace): number {
+        if (workspace.attributes) {
+            if (workspace.attributes.updated) {
+                return parseInt(workspace.attributes.updated);
+            } else if (workspace.attributes.created) {
+                return parseInt(workspace.attributes.created);
+            }
+        }
+
+        return NaN;
+    }
+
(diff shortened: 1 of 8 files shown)


extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
```diff
@@ -0,0 +1,229 @@
...
+        });
+
+        await this.open(workspaces);
+    }
+
+    private getWorkspaceName(workspace: cheApi.workspace.Workspace): string | undefined {
+        if (workspace.config) {
+            return workspace.config.name;
+        } else if (workspace.devfile && workspace.devfile.metadata) {
+            return workspace.devfile.metadata.name;
```
azatsarynnyy, line 137: Can a Workspace not have a name?
Maybe just use `workspace.devfile!.metadata!.name` to make a compiler happy?

extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
```diff
@@ -0,0 +1,229 @@
...
+    @inject(QuickOpenService) protected readonly quickOpenService: QuickOpenService;
+    @inject(CommandRegistry) protected readonly commandRegistry: CommandRegistry;
+    @inject(CheApiService) protected readonly cheApi: CheApiService;
+    @inject(OauthUtils) protected readonly oAuthUtils: OauthUtils;
+    @inject(LabelProvider) protected readonly labelProvider: LabelProvider;
+    @inject(MessageService) protected readonly messageService: MessageService;
+
+    registerCommands(commands: CommandRegistry): void {
+        // VS Code command id
+        commands.unregisterCommand('workbench.action.openRecent');
```
tsmaeder, line 43: If I've followed the conversation on the issue correctly, eventually we'll end up with two sets of commands ("Open Workspace") and "Open \<whatever we rename che workspaces to\>". Why not just leave the current commands in place and have call the Che workspace "Che workspace"?
Currently, the ability to "Open Workspace" and "Add Folder to Workspace" play an important role when doing development of che-theia. Please don't take them away.

extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
```diff
@@ -0,0 +1,229 @@
...
+            }));
+            return;
+        }
+
+        for (const workspace of workspaces) {
+            const icon = this.labelProvider.folderIcon;
+            const iconClass = icon + ' file-icon';
+            this.items.push(new QuickOpenGroupItem({
+                label: this.getWorkspaceName(workspace) + (this.isCurrentWorkspace(workspace) ? ' (Current)' : ''),
+                description: this.getWorkspaceStack(workspace),
```
azatsarynnyy, line 74: Since a Stack can be a long string, in case of Devfile link for example,
maybe it worth displaying it as [`detail`](https://github.com/eclipse-theia/theia/blob/a1ea2307d50565944b42f74ba3797698697ec897/packages/core/src/common/quick-open-model.ts#L38) instead of `description`?

It should allow using the whole widget width for Stack name, e.g.:
image

extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
```diff
@@ -0,0 +1,229 @@
...
+            } else if (workspace.attributes.created) {
+                return parseInt(workspace.attributes.created);
+            }
+        }
+
+        return NaN;
+    }
+
+    private stopCurrentWorkspace(): Promise<boolean | undefined> {
+        class StopWorkspaceDialog extends AbstractDialog<boolean | undefined> {
```
azatsarynnyy, line 162: Maybe extract the class a separate file?
It can be reused for [`Close Workspace`](https://github.com/eclipse/che/issues/17238) feature.

extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
```diff
@@ -0,0 +1,229 @@
...
+                run: (mode: QuickOpenMode): boolean => false
+            }));
+            return;
+        }
+
+        for (const workspace of workspaces) {
+            const icon = this.labelProvider.folderIcon;
+            const iconClass = icon + ' file-icon';
+            this.items.push(new QuickOpenGroupItem({
+                label: this.getWorkspaceName(workspace) + (this.isCurrentWorkspace(workspace) ? ' (Current)' : ''),
```
azatsarynnyy, line 73: Does it make sense to show the current Workspace on the list?
Maybe filter it on fetching all the workspaces stage?

extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
vzhukovskii: >we'll end up with two sets of commands ("Open Workspace") and "Open <whatever we rename che workspaces to>"

That is true. According to https://github.com/eclipse/che/issues/17106#issue-631935000 and https://github.com/eclipse/che/issues/17034 the current command will be replaced with Che's one. And with next PR (for the issue: https://github.com/eclipse/che/issues/17106#issue-631935000) default Theia's command will be renamed to a new one. But this is out of scope of current PR. This is the first step.

extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
```diff
@@ -0,0 +1,229 @@
...
+                return updatedB - updatedA;
+            }
+        });
+
+        await this.open(workspaces);
+    }
+
+    private getWorkspaceName(workspace: cheApi.workspace.Workspace): string | undefined {
+        if (workspace.config) {
+            return workspace.config.name;
```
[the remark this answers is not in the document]
vzhukovskii: Removed redundant check.

extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
```diff
@@ -0,0 +1,229 @@
...
+        } else if (workspace.devfile && workspace.devfile.metadata) {
+            return workspace.devfile.metadata.name;
+        }
+    }
+
+    private getWorkspaceStack(workspace: cheApi.workspace.Workspace): string | undefined {
```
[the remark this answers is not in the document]
vzhukovskii: AFAIK, this can be reached when user creates a workspace from custom devfile, editing directly from the dashboard, not importing the last one from URL.

extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
```diff
@@ -0,0 +1,229 @@
...
+        if (workspace.attributes && workspace.attributes.stackName) {
+            return `Stack: ${workspace.attributes.stackName}`;
```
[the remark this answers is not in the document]
vzhukovskii: Make sense. Updated the PR.

extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
vzhukovskii: Done.

extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
vzhukovskii: I thought about it, I'm going to extract this dialog into separate class during the next PR followed by this issue: https://github.com/eclipse/che/issues/17106#issue-631935000

extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
vzhukovskii: Current workspace will be the first workspace in a list at anytime. We can remove label `(Current)` from the workspace list. Or, we can place `Current` label to details section, as far as this section is available and empty at this moment.

extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
vzhukovskii: Yeah, one line is less code, but according to `@eclipse-che` library, `devfile`, `metadata` and `name` can be undefined. So simple check is more understandable in this case.

extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
azatsarynnyy: I'd either don't show the current Workspace or keep it as it is now.

vzhukovskii: Let's keep it then. Because choosing the current workspace from the list simply closes the current list and do nothing.

extensions/eclipse-che-theia-workspace/src/browser/che-quick-open-workspace.ts
```diff
@@ -0,0 +1,223 @@
...
+    @inject(OauthUtils) protected readonly oAuthUtils: OauthUtils;
+    @inject(LabelProvider) protected readonly labelProvider: LabelProvider;
+    @inject(MessageService) protected readonly messageService: MessageService;
+
+    registerCommands(commands: CommandRegistry): void {
+        // VS Code command id
+        commands.unregisterCommand('workbench.action.openRecent');
+
+        // Theia command id
+        commands.unregisterCommand('workspace:openRecent');
```
tsmaeder, line 46: I don't think we should unregister the command here: if an extension tries to execute `workbench.action.openRecent` it wants the original Theia command, not the Che version.
What we should do is unregister/reregister the menu items. There is an example of how to do that in https://github.com/eclipse/che-theia/pull/408/files#diff-698c505739271662ed1414a0a9c6314fR56

vzhukovskii: >if an extension tries to execute `workbench.action.openRecent` it wants the original Theia command

Can you provide an example of existed extension, that might call this command?

I think, in a context of Che, it would be reasonable to operate with Che workspaces. I mean, in Theia (pure), workspace means something like content root, in our case it is a physical workspace (Che workspace). The same situation for `Close Workspace` command (which will be the next step), in pure Theia it simply closes the browser window, in our case it should stop the current workspace. @l0rd wdyt?

tsmaeder: The effect on the UI level of what I'm requesting is the exact same thing. Let's say we register our commands under the id `che.openRecentWorskpace` if we write some code that wants to call this command via the plugin API, we can just use that command. If we overwrite the command, plugins cannot call the original command anymore (even from a different menu), so offering "Open Recent Theia Workspace..." becomes impossible to implement. Also, we can't possibly know what the intention of third-party authors with respect to theia vs che workspaces is. My suspicion is they would want to call the original command, yours seems to be the opposite. The conservative move is to leave the original command in place and add a new one.

vzhukovskii: You can see updates in latest commits. Is this okay for you?

Just note, if we remove default Theia's menu item (`Open Recent Workspace...`), the command is still registered and it is visible in command palette among with the registered Che command and menu item. Label for the Che menu item won't be changed. The label for Theia's one will be changed, but according to https://github.com/eclipse/che/issues/17106#issue-631935000 issue, not with this PR.

PR #802, opened 2020-07-10
Implementation of testing api for testing vscode extensions language features [Part 2]

svor:
### What does this PR do?
_This PR is based on https://github.com/eclipse/che-theia/pull/597 and was created to continue work on the task about providing language's TestAPI._

It introduces testing API that allows you to talk to a specific vscode extensions language features. It gives the ability to create tests that call into a vscode language extension for two reasons:

Tests can be run against newer version of language plugins so we know if a language plugin can be updated or something needs to be fixed
Che tests can talk directly with the language server so we know that if the language server request fails then it's an actual failure, rather than a test failing for a different reason and making it look like it was the cause
Examples of such tests can be found in: https://github.com/che-incubator/Che-Java-Tests

https://github.com/eclipse/che/issues/14409
partially https://github.com/eclipse/che/issues/15797


svor:
crw-ci-test

extensions/eclipse-che-theia-plugin-ext/src/browser/che-languages-test-api.ts
```diff
@@ -0,0 +1,177 @@
...
+    ReferenceContext,
+    Location,
+    SignatureHelpContext,
+    CodeActionContext,
+    CodeAction,
+    FoldingRange,
+} from '@theia/plugin-ext/lib/common/plugin-api-rpc-model';
+import { UriComponents } from '@theia/plugin-ext/lib/common/uri-components';
+import { CancellationToken, FoldingContext } from '@theia/plugin';
+import { SymbolInformation } from 'vscode-languageserver-types';
```
benoitf, line 36: is SymbolInformation should come from vscode-languageserver-types or it could come from `@theia/plugin` as well

extensions/eclipse-che-theia-plugin/src/che-proposed.d.ts
```diff
@@ -15,6 +15,34 @@
...
+    DocumentLink,
+    CodeLensSymbol,
+    DocumentSymbol,
+    ReferenceContext,
+    Location,
+    SignatureHelpContext,
+    CodeActionContext,
+    CodeAction,
+    FoldingRange,
+} from '@theia/plugin-ext/lib/common/plugin-api-rpc-model';
```
benoitf, line 37: it can't come from `@theia/plugin`  ?

extensions/eclipse-che-theia-plugin/src/che-proposed.d.ts
```diff
@@ -15,6 +15,34 @@
...
+import { UriComponents } from '@theia/plugin-ext/lib/common/uri-components';
+import { SymbolInformation } from 'vscode-languageserver-types';
```
benoitf, line 39: `@theia/plugin` for all these one as well

extensions/eclipse-che-theia-plugin-ext/src/browser/che-languages-test-api.ts
```diff
@@ -0,0 +1,177 @@
...
+    CodeLensSymbol,
+    DocumentSymbol,
+    ReferenceContext,
+    Location,
+    SignatureHelpContext,
+    CodeActionContext,
+    CodeAction,
+    FoldingRange,
+} from '@theia/plugin-ext/lib/common/plugin-api-rpc-model';
+import { UriComponents } from '@theia/plugin-ext/lib/common/uri-components';
```
benoitf, line 34: could we use theia.Uri ?

extensions/eclipse-che-theia-plugin-ext/src/browser/che-languages-test-api.ts
svor: at the end language test api requests methods from LanguagesExt https://github.com/eclipse-theia/theia/blob/master/packages/plugin-ext/src/common/plugin-api-rpc.ts#L1187 (for example see https://github.com/eclipse/che-theia/pull/802/files#diff-173df998f3e077e39aedf6ba7d61454fR58) and all these methods return types from `@theia/plugin-ext/lib/common/plugin-api-rpc-model` or `vscode-languageserver-types`  and they are different which we have in `@theia/plugin` that's why we can't use them

extensions/eclipse-che-theia-plugin-ext/src/browser/che-languages-test-api.ts
svor: we could, fixed

extensions/eclipse-che-theia-plugin/src/che-proposed.d.ts
benoitf: it would be nice if here, we could keep only from @theia/plugin. For the other locations, it's not important as it's implementation details.

It's because by doing this, we ensure that the dependencies are very small when linking to `@eclipse-che/plugin`

extensions/eclipse-che-theia-plugin-ext/src/browser/che-languages-test-api.ts
benoitf: yes for implementation I would say why not, I'm more concerned for the `che.d.ts`

svor:
@benoitf PR was updated to re-use declarations from theia.d.ts, could you please take another look

benoitf:
@svor : do you have a testing plug-in that you're using ?

svor:
@benoitf test plugin is not ready yet, i'm working on adapting it

svor:
@benoitf 
Branch with adapted tests is https://github.com/svor/che-java-tests/tree/sv/tests but it is still in progress
here is a video about how it looks when running tests inside che-the: https://youtu.be/vT0eSgexYB0

benoitf (commented):
oups, it seems my review was not published
clicking submit now

extensions/eclipse-che-theia-plugin-ext/src/browser/type-converters.ts
```diff
@@ -0,0 +1,378 @@
...
+import * as Converter from '@theia/plugin-ext/lib/plugin/type-converters';
+import { URI } from 'vscode-uri';
+import {
+    Selection, Position, RawColorInfo, ResourceFileEditDto,
+    ResourceTextEditDto
+} from '@theia/plugin-ext/lib/common/plugin-api-rpc';
+
+export function toWorkspaceEdit(workspaceEdit: model.WorkspaceEdit): theia.WorkspaceEdit {
+    const result: theia.WorkspaceEdit = new types.WorkspaceEdit();
+    console.log(result);
```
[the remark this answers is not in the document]
svor: cleaned

extensions/eclipse-che-theia-plugin-ext/src/browser/type-converters.ts
```diff
@@ -0,0 +1,378 @@
...
+    return range;
+}
+
+export function toDocumentLink(link: model.DocumentLink): theia.DocumentLink {
+    let target = undefined;
+    if (link.url) {
+        try {
+            target = typeof link.url === 'string' ? URI.parse(link.url, true) : URI.revive(link.url);
+        } catch (err) {
+            // ignore
```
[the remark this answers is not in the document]
svor: logged by `console.error(err);`

benoitf (approved):
tested on a custom che instance

PR #803, opened 2020-07-13
Describe running and debugging plugins in sidecars

tsmaeder:

### What does this PR do?
Describes workflows for working with plugins in sidecar containers in detail.

https://github.com/eclipse/che/issues/13249

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -217,27 +217,120 @@ $ export THEIA_PLUGIN_ENDPOINT_DISCOVERY_PORT="2506"
 $ yarn theia start /projects/theia-projects-dir --hostname=0.0.0.0 --port=3010
 ```
 
-## How to develop Che Theia remote plugin mechanism
+# Working with Plugins in Sidecars
 
-_Please note, this section provides a flow how to develop remote plugin mechanism in Che Theia, but not a remote plugin._
+In a real Che workspace, many plugins are run in their own "sidecar" container. The idea is that the container furnishes any "native" dependencies the plugin needs. For example the container that runs the theia IDE back end does not contain a Java SDK, so in order to run a plugin that needs a Java installation, we'll run the plugin inside a container that has Java installed.
 
-First, create a workspace from prepared [devfile](https://github.com/eclipse/che-theia/blob/master/extensions/eclipse-che-theia-plugin-remote/devfile.yaml), which could be found in the `eclipse-che-theia-plugin-remote` extension folder.
-When workspace is ready:
- - Init Che Theia. This could be done with `che-theia init` command in `/projects/theia` folder or run the init command.
-   Che Theia sources will be awailable at `/projects/theia/che/che-theia`.
- - Now one may make changes in Che Theia remote plugin mechanism in both (Che Theia and Remote plugin) sides.
- - Build Che Theia from `theia-dev` container by executing `yarn` in `/projects/theia` folder or by running corresponding command.
- - Put binaries (*.theia or *.vsix) of your remote plugin(s) into `/projects/remote-plugins/` directory.
-   Note that the plugin(s) shouldn't have any external dependencies.
-   For example, [this sample plugin](https://github.com/eclipse/che-theia-samples/tree/master/samples/hello-world-backend-plugin) might be used.
-   Or a user may generate one using `Generate Hello World plugin package` command.
- - Run dev Che Theia and Remote plugins endpoint in `theia-dev` and `theia-remote-runtime-dev` containers correspondingly.
-   One may use predefined commands to start them.
- - Open `theia-dev` route from `My Workspace` panel and test chenges.
+As when working on developing Che-Theia itself, the approach is to run a copy of the Theia back end on a different port than that we use for the self hosting workspace. 
 
-Also it is possible to run watchers for remote plugin mechanism.
-In `theia-dev` container run `npx run watch @eclipse-che/theia-remote` from `/projects/theia` folder to recompile the extension on changes made.
-Also run `yarn watch` in `/projects/theia/examples/assembly` to bring the changes to Che Theia binaries.
-If needed one may start watchers in plugin API extension: `npx run watch @theia/plugin-ext` from `/projects/theia` directory.
-The commands for these actions are also available.
-But please note, you have to restart server to which changes is made.
+## Setting up the devfile
+
+So, starting with a workspace to develop che-theia, we'll have to add a container that will serve as the sidecar running our plugin.
+
+```yaml
+  - mountSources: true
+    memoryLimit: 2Gi
+    type: dockerimage
+    image: 'quay.io/eclipse/che-theia-dev:next'
+    alias: remote-1
+    env:
+      - value: '2504'
+        name: THEIA_PLUGIN_ENDPOINT_DISCOVERY_PORT
+      - value: 'local-dir:///projects/plugins-1/'
+        name: THEIA_PLUGINS
+      - value: /projects/java-11-openjdk
+        name: JAVA_HOME
+```
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -217,27 +217,122 @@ $ export THEIA_PLUGIN_ENDPOINT_DISCOVERY_PORT="2506"
...
+```json
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  "version": "0.2.0",
+  "configurations": [
+    {
+        "type": "node",
+        "request": "attach",
+        "name": "Java",
```
sunix, line 329: Naming it Java is confusing in IMO, maybe vscodejava is more appropriate?

CONTRIBUTING.md
```diff
@@ -217,27 +217,122 @@ $ export THEIA_PLUGIN_ENDPOINT_DISCOVERY_PORT="2506"
...
-But please note, you have to restart server to which changes is made.
+# Working with Plugins in Sidecars
+
+In a real Che workspace, many plugins are run in their own "sidecar" container. The idea is that the container furnishes any "native" dependencies the the plugins needs. For example the container that runs the theia IDE back end does not contain a Java SDK, so in order to run a plugin that needs a Java installation, we'll run the plugin inside a container that has Java installed.
+
+As when working on developing Che-Theia itself, the approach is to run a copy of the Theia back end on a different port than that we use for the self hosting workspace. 
+
+## Setting up the devfile
+
+So, starting with a workspace to develop che-theia, we'll have to add a container that will serve as the sidecar running our plugin.
```
sunix, line 228: This is very nice to show how we could build the devfile. Is the final ready-to-use devfile (for java plugin che-in-che) available somewhere? I think it would make sense to keep this somewhere so contributors would have something ready to use.

CONTRIBUTING.md
```diff
@@ -217,27 +217,122 @@ $ export THEIA_PLUGIN_ENDPOINT_DISCOVERY_PORT="2506"
...
-
-Also it is possible to run watchers for remote plugin mechanism.
-In `theia-dev` container run `npx run watch @eclipse-che/theia-remote` from `/projects/theia` folder to recompile the extension on changes made.
-Also run `yarn watch` in `/projects/theia/examples/assembly` to bring the changes to Che Theia binaries.
-If needed one may start watchers in plugin API extension: `npx run watch @theia/plugin-ext` from `/projects/theia` directory.
-The commands for these actions are also available.
-But please note, you have to restart server to which changes is made.
+# Working with Plugins in Sidecars
+
+In a real Che workspace, many plugins are run in their own "sidecar" container. The idea is that the container furnishes any "native" dependencies the the plugins needs. For example the container that runs the theia IDE back end does not contain a Java SDK, so in order to run a plugin that needs a Java installation, we'll run the plugin inside a container that has Java installed.
```
ericwill, line 222: `the the plugins needs` -> `the plugin needs`

CONTRIBUTING.md
```diff
@@ -217,27 +217,122 @@ $ export THEIA_PLUGIN_ENDPOINT_DISCOVERY_PORT="2506"
...
+    reference: >-
+      https://raw.githubusercontent.com/eclipse/che-plugin-registry/master/v3/plugins/ms-vscode/node-debug2/1.42.3/meta.yaml
+    alias: node-debug
+```
+
+Note that some earlier versions of the plugin did not work in a self-hosting setup, so make sure you are using a recent one. 
+
+## Setting up plugins to run
+
+We can now add plugins to run in the sidecar container by adding them to the folder `/projects/plugins-1`(or whaterver we define as `THEIA_PLUGINS`env variable). If we're just interested in debugging the theia plugins host process, we can download a *.vsix file to the folder: it will be unpacked to a subfolder of `/tmp`. If we need to debug the plugin, we can checkout the source code and built it according the the instructions for the plugin:
```
ericwill, line 283: `and built it` -> `and build it`

ericwill:
@tsmaeder can we get this one merged before the sprint is out?

CONTRIBUTING.md
tsmaeder: I have chosen not to provide one: I want people to understand what is happening and to be capable of creating their own devfiles. They will have to anyway if their sidecars have any native dependencies.

benoitf:
FYI, would be nice to fix vale errors (the plug-in used when working on che documentation)

image

image

CONTRIBUTING.md
```diff
@@ -217,27 +217,121 @@ $ export THEIA_PLUGIN_ENDPOINT_DISCOVERY_PORT="2506"
...
+    image: 'quay.io/eclipse/che-theia-dev:next'
+    alias: theia-dev
+    env:
+      - value: '2504'
+        name: THEIA_PLUGIN_ENDPOINT_DISCOVERY_PORT
+```
+
+This way, the IDE and the remote plugin host process can find each other.
+
+The container in question must be able to work as a plugin sidecar for any plugin you want to run in it. If you are trying to debug a plugin that already has a sidecar container, it would make sense to base `remote-1`container on the sidecar that runs the original plugin. In that case, you would have to add a copy of node to sidecar container, since they generally don't contain one and we need to run the runtime host process, which is a node application. In this case, however, I've started with a `theia-dev`image and just point it to a JDK I've copied to `/projects/java-11-openjdk`by setting `JAVA_HOME`.
```
benoitf, line 268:
```suggestion
The container in question must be able to work as a plugin sidecar for any plugin you want to run in it. If you are trying to debug a plugin that already has a sidecar container, it would make sense to base `remote-1` container on the sidecar that runs the original plugin. In that case, you would have to add a copy of node to sidecar container, since they generally don't contain one and we need to run the runtime host process, which is a node application. In this case, however, I've started with a `theia-dev` image and just point it to a JDK I've copied to `/projects/java-11-openjdk` by setting `JAVA_HOME`.
```

CONTRIBUTING.md
```diff
@@ -217,27 +217,121 @@ $ export THEIA_PLUGIN_ENDPOINT_DISCOVERY_PORT="2506"
...
+
+If we want to debug later on, we'll have to add the `node-debug2`Che plugin to our workspace:
```
benoitf, line 270:
```suggestion
If we want to debug later on, we'll have to add the `node-debug2` Che plugin to our workspace: 
```

CONTRIBUTING.md
```diff
@@ -217,27 +217,121 @@ $ export THEIA_PLUGIN_ENDPOINT_DISCOVERY_PORT="2506"
...
+    reference: >-
+      https://raw.githubusercontent.com/eclipse/che-plugin-registry/master/v3/plugins/ms-vscode/node-debug2/1.42.3/meta.yaml
+    alias: node-debug
+```
+
+Note that some earlier versions of the plugin did not work in a self-hosting setup, so make sure you are using a recent one. 
+
+## Setting up plugins to run
+
+We can now add plugins to run in the sidecar container by adding them to the folder `/projects/plugins-1`(or whaterver we define as `THEIA_PLUGINS`env variable). If we're just interested in debugging the theia plugins host process, we can download a *.vsix file to the folder: it will be unpacked to a subfolder of `/tmp`. If we need to debug the plugin, we can checkout the source code and build it according the the instructions for the plugin:
```
benoitf, line 283:
```suggestion
We can now add plugins to run in the sidecar container by adding them to the folder `/projects/plugins-1` (or whaterver we define as `THEIA_PLUGINS` env variable). If we're just interested in debugging the theia plugins host process, we can download a *.vsix file or *.theia file to the folder: it will be unpacked to a subfolder of `/tmp`. If we need to debug the plugin, we can checkout the source code and build it according the the instructions for the plugin:
```

tsmaeder: Long term, we're trying to de-emphasize the theia API. Plus, the only theia-format plugins I'm aware of are our own, which run in the theia container.

tsmaeder:
crw-ci-test

benoitf:
I think you need to rebase to current master first.

PR #815, opened 2020-07-27
Add a warning about blocked oauth popup

vinokurig:

### What does this PR do?
If the `getUserToken()` function works longer then 10 sec. warning notification about blocked oauth popup appears.

https://github.com/eclipse/che/issues/16930

### Hapy Path Channel

HAPPY_PATH_CHANNEL=stable


benoitf:
Hello Igor,

if the user is taking more than 10s to complete oAuth it will also display this popup ?

vinokurig:
Hello Florent,
It is expected that user is already logged in to Che when this function is called, so the popup window which is a web-application from che-server should just open and close immediately after sending the user token back.

PR #821, opened 2020-08-03
Freeze Theia version 1.4.0

RomanNikitenko:

### What does this PR do?
Some changes on Theia side were merged today, so we have to freeze Theia version and provide some adaptation on our side.


RomanNikitenko:
> Please, create a corresponding issue for adaptation to upstream Theia. I guess it depends on the dropped monaco-language client.

I tested with commit before monaco-language client changes and tests are failed: https://github.com/eclipse/che-theia/pull/695
So maybe we have breaking changes before monaco-language client changes as well...
I'm going to create a common issue and we will investigate the cause of the problem.

PR #824, opened 2020-08-04
Support custom public certificates

vzhukovskii:
### What does this PR do?
This changes proposal adds an ability to provide custom certificates.
Custom certificates should be located in `/public-certs/*.crt`

https://github.com/eclipse/che/issues/17440

### Hapy Path Channel

HAPPY_PATH_CHANNEL=stable

### DIFF
--- a/extensions/eclipse-che-theia-plugin-ext/src/node/che-plugin-service.ts
+++ b/extensions/eclipse-che-theia-plugin-ext/src/node/che-plugin-service.ts
@@ -24,7 +24,8 @@ import URI from '@theia/core/lib/common/uri';
 import { PluginFilter } from '../common/plugin/plugin-filter';
 import * as fs from 'fs-extra';
 import * as https from 'https';
-import { SS_CRT_PATH } from './che-https';
+import { PUBLIC_CRT_PATH, SS_CRT_PATH } from './che-https';
+import * as path from 'path';
 
 const yaml = require('js-yaml');
 
@@ -127,11 +128,11 @@ export class ChePluginServiceImpl implements ChePluginService {
                 const httpOverHttpsAgent = tunnel.httpOverHttps({ proxy: httpsProxyOptions });
                 const httpsOverHttpAgent = tunnel.httpsOverHttp({
                     proxy: mainProxyOptions,
-                    ca: certificateAuthority ? [certificateAuthority] : undefined
+                    ca: certificateAuthority ? certificateAuthority : undefined
                 });
                 const httpsOverHttpsAgent = tunnel.httpsOverHttps({
                     proxy: httpsProxyOptions,
-                    ca: certificateAuthority ? [certificateAuthority] : undefined
+                    ca: certificateAuthority ? certificateAuthority : undefined
                 });
                 const urlIsHttps = (parsedBaseUrl.protocol || 'http:').startsWith('https:');
                 const proxyIsHttps = (parsedProxyUrl.protocol || 'http:').startsWith('https:');
@@ -155,13 +156,13 @@ export class ChePluginServiceImpl implements ChePluginService {
         return axios;
     }
 
-    private getHttpsProxyOptions(mainProxyOptions: tunnel.ProxyOptions, servername: string | undefined, certificateAuthority: Buffer | undefined): tunnel.HttpsProxyOptions {
+    private getHttpsProxyOptions(mainProxyOptions: tunnel.ProxyOptions, servername: string | undefined, certificateAuthority: Buffer[] | undefined): tunnel.HttpsProxyOptions {
         return {
             host: mainProxyOptions.host,
             port: mainProxyOptions.port,
             proxyAuth: mainProxyOptions.proxyAuth,
             servername,
-            ca: certificateAuthority ? [certificateAuthority] : undefined
+            ca: certificateAuthority ? certificateAuthority : undefined
         };
     }
 
@@ -196,12 +197,23 @@ export class ChePluginServiceImpl implements ChePluginService {
         return (typeof process !== 'undefined') && (typeof process.versions.node !== 'undefined');
     }
 
-    private getCertificateAuthority(): Buffer | undefined {
-        let certificateAuthority: Buffer | undefined;
+    private getCertificateAuthority(): Array<Buffer> | undefined {
+        const certificateAuthority: Buffer[] = [];
         if (fs.existsSync(SS_CRT_PATH)) {
-            certificateAuthority = fs.readFileSync(SS_CRT_PATH);
+            certificateAuthority.push(fs.readFileSync(SS_CRT_PATH));
         }
-        return certificateAuthority;
+
+        if (fs.existsSync(PUBLIC_CRT_PATH)) {
+            const publicCertificates = fs.readdirSync(PUBLIC_CRT_PATH);
+            for (const publicCertificate of publicCertificates) {
+                if (publicCertificate.endsWith('.crt')) {
+                    const certPath = path.join(PUBLIC_CRT_PATH, publicCertificate);
+                    certificateAuthority.push(fs.readFileSync(certPath));
+                }
+            }
+        }
+
+        return certificateAuthority.length > 0 ? certificateAuthority : undefined;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


vzhukovskii:
PR's build can be failed due to separate PR in workspace client library: https://github.com/eclipse/che-workspace-client/pull/34

I'll fork workspace-client library with changes and provide custom image for che-theia with changes from current PR and PR from workspace-client repository.

tolusha:
If I understand correctly `theia` will use '/tmp/che/secret/ca.crt' plus certificates located in `/public-certs` folder.
It is ok for me.

vzhukovskii:
>If I understand correctly theia will use '/tmp/che/secret/ca.crt' plus certificates located in /public-certs folder.

That's correct, '/tmp/che/secret/ca.crt' is using in conjunction with public certificates.

vzhukovskii:
> I see that `/tmp/che/secret/ca.crt` is used here
> 
> https://github.com/eclipse/che-theia/blob/5155d757a7228f55975f53609d8b8a43668787f0/plugins/workspace-plugin/src/theia-commands.ts#L21
> 
> Should we have some changes here?

Nice catch! I'll add parameters to the curl call. Thanks!

vzhukovskii:
> I see that `/tmp/che/secret/ca.crt` is used here
> 
> https://github.com/eclipse/che-theia/blob/5155d757a7228f55975f53609d8b8a43668787f0/plugins/workspace-plugin/src/theia-commands.ts#L21
> 
> Should we have some changes here?

According to `curl` documentation it is enough to pass only parameter `-k` if any certificate is found:

```
-k, --insecure

(TLS) By default, every SSL connection curl makes is verified to be secure. This option allows curl to proceed and operate even for server connections otherwise considered insecure.

The server connection is verified by making sure the server's certificate contains the right name and verifies successfully using the cert store.

See this online resource for further details:  https://curl.haxx.se/docs/sslcerts.html

See also --proxy-insecure and --cacert.
```

not sure we should pass every certificate as parameters in this case, so, leave this code as is.

RomanNikitenko:
> According to curl documentation it is enough to pass only parameter -k if any certificate is found
> not sure we should pass every certificate as parameters in this case, so, leave this code as is.

is it possible that `/tmp/che/secret/ca.crt` is NOT found, 
but
`/public-certs` is found?

If it's possible, should we pass `parameter -k` for such case?
https://github.com/eclipse/che-theia/blob/5155d757a7228f55975f53609d8b8a43668787f0/plugins/workspace-plugin/src/theia-commands.ts#L209

vzhukovskii:
>is it possible that /tmp/che/secret/ca.crt is NOT found

I suppose no, it's impossible. @tolusha is it possible to run che on http environment?

vzhukovskii:
Here is an image for testing purposes: `vzhukovs/che-theia:17440`. It was built based on https://github.com/eclipse/che-workspace-client/pull/34 and https://github.com/eclipse/che-theia/pull/824 changes.

azatsarynnyy:
I believe we can merge this part and the whole flow will be checked once all the other parts are finished within [the downstream issue](https://issues.redhat.com/browse/CRW-1054).
@tolusha WDYT?

tolusha:
@azatsarynnyy 
Pls merge. If we find any issues we will let you know.

vzhukovskii:
[ci-build]

vzhukovskii:
crw-ci-test

vzhukovskii:
crw-ci-test

vzhukovskii:
crw-ci-test

vzhukovskii:
crw-ci-test

vzhukovskii:
ci-build

vzhukovskii:
[ci-build]

vzhukovskii:
[ci-build]

vzhukovskii:
crw-ci-test

mmorhun (approved):
I tested this changes on Openshift 4 with institutional (self-signed) certificate and cluster wide proxy which is configured to intercept TLS. Everything works as expected.

azatsarynnyy:
Thank you very much @mmorhun for testing it!

@vzhukovskii now we can proceed with getting it merged:
- get the HappyPath tests passed
- create a corresponding PR for workspace-client

tolusha:
@azatsarynnyy 
Could you merge changes in 7.18.x branch as well ?
In this case the fix will be available in CRW 2.4

azatsarynnyy:
> @azatsarynnyy
> Could you merge changes in 7.18.x branch as well ?
> In this case the fix will be available in CRW 2.4

Sure. That's the plan.
@tolusha thanks for reminding.

vzhukovskii:
crw-ci-test

davidfestal:
> is it possible that `/tmp/che/secret/ca.crt` is NOT found, but `/public-certs` is found?

@tolusha According to our today discussion, I would say it is possible, no ?

If the router CA extracted by the Che operator through a dummy route is not seen as self-signed, then the `self-signed-certificate` secret will not be created, and no `/tmp/che/secret/ca.crt` will be found, right ?
But I assume this should not prevent an administrator to add custom public certificates to the `ca-certs` config map and expect to have them propagated up to `che-theia`, so that the `/public-certs` would be found ?

extensions/eclipse-che-theia-plugin-ext/src/node/che-plugin-service.ts
```diff
@@ -127,11 +128,11 @@ export class ChePluginServiceImpl implements ChePluginService {
                 const httpOverHttpsAgent = tunnel.httpOverHttps({ proxy: httpsProxyOptions });
```
davidfestal, line 128: It seems to me that the `httpsProxyOptions` might need to have the `public-certs` as well, since this is the bundle that contains the intercepting proxy CA (coming from the cluster-wide-proxy).

Does it make sense @tolusha ?

The same change would have to be done on the corresponding changes in the `che-workspace-client` library.

extensions/che-theia-hosted-plugin-manager-extension/package.json
```diff
@@ -12,7 +12,7 @@
     "@theia/core": "next",
     "@theia/plugin-dev": "next",
     "@theia/plugin-ext": "next",
-    "@eclipse-che/workspace-client": "latest",
+    "@vzhukovs/workspace-client": "0.0.9",
```
vitaliy-guliy, line 15: Should it be replaced on `@eclipse-che/workspace-client` before merging?

tolusha:
@davidfestal 
> is it possible that /tmp/che/secret/ca.crt is NOT found, but /public-certs is found?
Yes, it is possible accordingly to our discution.

tolusha:
@davidfestal 
> It seems to me that the httpsProxyOptions might need to have the public-certs as well, since this is the bundle that contains the > > intercepting proxy CA (coming from the cluster-wide-proxy).
> Does it make sense @tolusha ?

It is hard to answer,  I am not familiar with that how it works in theia.

extensions/che-theia-hosted-plugin-manager-extension/package.json
vzhukovskii: Yes, it will be replaced before merge

vzhukovskii:
> @davidfestal
> 
> > is it possible that /tmp/che/secret/ca.crt is NOT found, but /public-certs is found?
> 
> Yes, it is possible accordingly to our discussion.

I suppose, it would be better to implement this in separate issue/pr. This is out of scope of current issue. @davidfestal, @tolusha wdyt?

azatsarynnyy:
> I suppose, it would be better to implement this in separate issue/pr. This is out of scope of current issue

I'm +1 for making a separate issue for that case.

tolusha:
@azatsarynnyy 
> is it possible that /tmp/che/secret/ca.crt is NOT found, but /public-certs is found?

Accordingly to `private getCertificateAuthority():` you already handle the this case. Am I right?

vzhukovskii:
> @azatsarynnyy
> 
> > is it possible that /tmp/che/secret/ca.crt is NOT found, but /public-certs is found?
> 
> Accordingly to `private getCertificateAuthority():` you already handle the this case. Am I right?

You're right. This logic is in master branch now and in current PR.

vzhukovskii:
Are there any objections, which prevent merging this PR to master? @tolusha @davidfestal @azatsarynnyy

tolusha:
I don't have any.

azatsarynnyy:
> Are there any objections, which prevent merging this PR to master?

I've already approved it.

PR #826, opened 2020-08-10
feat(terminal): Use workspace root as default path in containers

benoitf:
### What does this PR do?
Open by default terminal in workspace root.
In case of multiple workspace roots, prompt user.

Fix https://github.com/eclipse/che/issues/17548

### Happy Path Channel
HAPPY_PATH_CHANNEL=stable

Change-Id: I8ad33cd6d85c5582c347bcc5390c25daacf825ef

### DIFF
--- a/extensions/eclipse-che-theia-terminal/src/browser/contribution/exec-terminal-contribution.ts
+++ b/extensions/eclipse-che-theia-terminal/src/browser/contribution/exec-terminal-contribution.ts
@@ -186,13 +186,7 @@ export class ExecTerminalFrontendContribution extends TerminalFrontendContributi
     }
 
     async openTerminalByContainerName(containerName: string): Promise<void> {
-        const editorContainer = await this.getEditorContainerName();
-        let cwd: string | undefined;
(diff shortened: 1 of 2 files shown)


benoitf:
@azatsarynnyy with multiple workspace root ?

azatsarynnyy:
> @azatsarynnyy with multiple workspace root ?

@benoitf yes

benoitf:
@azatsarynnyy it seems there is a bug with the two calls with 2 quick prompts (we go in the same code but second prompt is not shown)

benoitf:
@azatsarynnyy it looks like there is a bug when there are two prompts called in a row. (one for choosing container and one for the path)

benoitf:
ok, I had to debug quick open service which was not directly related to the original change as we were supposed to handle it correctly for the theia-editor container.

There is an issue when quick open service was called from itself and then the new prompt was not displayed but we got undefined result.

I added a setTimeout/0 to call from another tick

benoitf:
@azatsarynnyy could you please test again ?

azatsarynnyy:
> @azatsarynnyy could you please test again ?

works well now
thanks @benoitf

benoitf:
thanks @azatsarynnyy

PR #840, opened 2020-08-22
Use CHE_API_INTERNAL instead of CHE_API_EXTERNAL.

monaka:
### What does this PR do?

Uses CHE_API_INTERNAL instead of CHE_API_EXTERNAL.
As che-theia backend is behind of Ingress, it can be connected to the Che API server with the K8s DNS name.

None.


azatsarynnyy:
crw-ci-test

azatsarynnyy (approved):
Makes sense. Thank you for your contribution!

PR #842, opened 2020-08-27
Make plugin-host drive startup sequence

tsmaeder:
### What does this PR do?
This is a companion PR to https://github.com/eclipse-theia/theia/pull/8437. While debugging plugin startup, I noticed some messages to the front end were ignored silently, because "setClient()" had not been called before the message were sent.

The work is happening in the context of https://github.com/eclipse/che/issues/16502. I was trying to register a custom file system at startup of the plugin host and noticed that the client side was not ready for business at that time.
#### Docs PR
n/a 

### Happy Path Channel

HAPPY_PATH_CHANNEL=next


nickboldt:
This PR is 13 months old and has conflicts. closing.

tsmaeder:
Some of the improvements in the companion PR (Multiplexer an connection refactorings) have been introduced into the Theia code base via other PR's, but the underlying timing problem is still present. It's hard to drum up support for fixing architectural problems that don't explode into user's faces 🤷

PR #870, opened 2020-10-01
feat(cypress): Remove cypress tests as now we have happy path tests

benoitf:
### What does this PR do?
Remove cypress e2e tests
They were introduced before happy path tests and QE team is not using cypress but selenium so no need to invest time on it.
But now for example typescript LS is not integrated in the image so it tests almost nothing except to see if command palette is there.

N/A

### Happy Path Channel

HAPPY_PATH_CHANNEL=stable

Change-Id: I6b61ae2e3ebd921928609640941b46211f739725


azatsarynnyy (approved):
+1

vzhukovskii (approved):

PR #885, opened 2020-10-20
Open Explorer on start

azatsarynnyy:

### What does this PR do?
This PR adds a couple of UX enhancements to a new user flow:
- each time when a new Workspace started, Project Explorer is opened automatically
- fixed opening `README.md` file - Che Theia starts looking for a file only after finishing cloning the project's sources
- project folder that contains a `README.md` file is auto-expanded

Peek 2020-10-27 10-07

closes https://github.com/eclipse/che/issues/17969

### Happy Path Channel

HAPPY_PATH_CHANNEL=stable

### DIFF
--- a/extensions/eclipse-che-theia-workspace/src/browser/explorer-contribution.ts
+++ b/extensions/eclipse-che-theia-workspace/src/browser/explorer-contribution.ts
@@ -1,0 +1,25 @@
+/*********************************************************************
+ * Copyright (c) 2020 Red Hat, Inc.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ **********************************************************************/
+
+import { inject, injectable } from 'inversify';
+import { FrontendApplicationContribution, FrontendApplication, ApplicationShell } from '@theia/core/lib/browser';
+import { FILE_NAVIGATOR_ID } from '@theia/navigator/lib/browser';
+
+@injectable()
+export class ExplorerContribution implements FrontendApplicationContribution {
+
+    @inject(ApplicationShell)
+    protected readonly shell: ApplicationShell;
+
+    // Note, it's called only when there is no previous workbench layout state is stored.
+    async initializeLayout(app: FrontendApplication): Promise<void> {
+        this.shell.revealWidget(FILE_NAVIGATOR_ID);
+    }
+}
--- a/plugins/welcome-plugin/package.json
+++ b/plugins/welcome-plugin/package.json
@@ -35,6 +35,9 @@
     "@theia/plugin-packager": "latest",
     "@eclipse-che/plugin": "0.0.1"
   },
+  "extensionDependencies": [
+    "Eclipse Che.@eclipse-che/workspace-plugin"
+  ],
   "scripts": {
     "prepare": "yarn run clean && yarn run build",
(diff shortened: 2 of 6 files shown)


extensions/eclipse-che-theia-workspace/src/browser/explorer-contribution.ts
```diff
@@ -0,0 +1,31 @@
...
+@injectable()
+export class ExplorerContribution implements FrontendApplicationContribution {
+
+    @inject(FrontendApplicationStateService)
+    protected readonly stateService: FrontendApplicationStateService;
+
+    @inject(CommandService)
+    protected readonly commandService: CommandService;
+
+    async onStart(app: FrontendApplication): Promise<void> {
```
benoitf, line 26: it seems you can just bind on initializeLayout instead of start and use https://github.com/eclipse-theia/theia/issues/2005#issuecomment-393787523
 navigator.openView({
        activate: true
    })

benoitf: if you use the command (doing the same stuff), maybe just need to move from onStart to initializeLayout

benoitf: what would be nice is to have folders expanded after cloning the projects

benoitf (approved):

benoitf:
maybe it should be part of the welcome plug-in (as it's just executing a command) to not use extension mechanism where we don't need it.

https://github.com/eclipse/che-theia/tree/master/plugins/welcome-plugin

azatsarynnyy:
> if you use the command (doing the same stuff), maybe just need to move from onStart to initializeLayout

It's a good idea! Thx @benoitf  I'll try it.
In that case, the previous workbench layout state shouldn't be ignored as it's when called from onStart

> what would be nice is to have folders expanded after cloning the projects

That's exactly what I'm working on next )

> maybe it should be part of the welcome plug-in

Yeah, it could be. But I want to expand a folder as well, and there's no Plugin API to do it. That's why I've chosen an extension for both actions: opening Explorer + expanding a folder.

azatsarynnyy:
> what would be nice is to have folders expanded after cloning the projects

> maybe it should be part of the welcome plug-in

@benoitf actually, I've achieved both by reworking a bit opening a `README.md` file. :smile:

plugins/welcome-plugin/package.json
```diff
@@ -35,6 +35,9 @@
     "@theia/plugin-packager": "latest",
     "@eclipse-che/plugin": "0.0.1"
   },
+  "extensionDependencies": [
+    "Eclipse Che.@eclipse-che/workspace-plugin"
```
benoitf, line 39: should we fix our plug-ins ? so that extension dependencies looks better (without space) ?

benoitf (approved):
image

works as expected 👍  🎉

plugins/welcome-plugin/package.json
azatsarynnyy: It's not a problem but to comply with the VS Code's rule for extensions IDs
> The id of an extension is always ${publisher}.${name}. For example: vscode.csharp.

we can consider renaming our plug-ins. I'll create an issue today for renaming and testing it.

azatsarynnyy: sorry, I meant matching the pattern
Screenshot from 2020-10-27 23-03-39

azatsarynnyy: created the issue https://github.com/eclipse/che/issues/18220

PR #946, opened 2020-12-07
Implement simple side car file system which fetches file content

vzhukovs:
### What does this PR do?
This changes proposal adds a registration for sidecar file scheme: `file-sidecar-${machineName}`. As the first iteration across [dependent issue #16870](https://github.com/eclipse/che/issues/16870) only two methods from file system have been implemented: `stat` and `readFile`.

Signed-off-by: Vladyslav Zhukovskyi <vzhukovs@redhat.com>

### Screenshot/screencast of this PR
This changes doesn't provide any UI feedback. Just internal feature.

### What issues does this PR fix or reference?
https://github.com/eclipse/che/issues/18405

### How to test this PR?
Here is the devfile for the workspace, based on this PR:
```yaml
apiVersion: 1.0.0
metadata:
  name: che-18405
components:
  - memoryLimit: 256Mi
    type: chePlugin
    reference: 'https://gist.githubusercontent.com/vzhukovs/29fccda72fbbd63e0c8aa8f751c9745a/raw/713d0b9ab47576ddb25c5a1a423346dd8f645268/meta.yaml'
  - memoryLimit: 1024Mi
    type: cheEditor
    reference: 'https://gist.githubusercontent.com/vzhukovs/e341154d8cbd33d5e56b1cbd719238ae/raw/c7ff20bbeb8a36c97559987cb2a4c79beba56c46/meta.yaml'
```

I've created a simple VS Code plugin that located here: [sample-0.0.1.vsix](https://github.com/vzhukovs/vscode-webview-sample/blob/sidecar/sample-0.0.1.vsix). Source code of extension located here: [extension.ts](https://github.com/vzhukovs/vscode-webview-sample/blob/sidecar/src/extension.ts)

There is a sample command in command palette called: **Side Car: Show Webview**

Expected behavior - webview will be opened and resources, that are locaed on the dedicate container will be successfully loaded (css, js and font).

### PR Checklist

[As the author of this Pull Request I made sure that:](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#pull-request-template-and-its-checklist)

- [x] [The Eclipse Contributor Agreement is valid](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-eclipse-contributor-agreement-is-valid)
- [x] [Code produced is complete](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-produced-is-complete)
[...]

PR #948, opened 2020-12-08
Load snippets via plugin resource

tsmaeder:

### What does this PR do?
Load snippet resources via the plugin resource loading mechanism instead of file uris (like the are in Theia)

https://github.com/eclipse/che/issues/16502

### How to test this PR?
Use the Che-Theia image built for the happy-path test in any workspace including the Java plugin and try using snippets in any Java file. 

### PR Checklist

### Reviewers

### Happy Path Channel

HAPPY_PATH_CHANNEL=next


tsmaeder:
[crw-ci-test --rebuild]

extensions/eclipse-che-theia-plugin-remote/src/browser/plugin-resource.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  dispose(): void {}
+}
+
+@injectable()
+export class PluginResourceResolver implements ResourceResolver {
+  resolve(uri: URI): MaybePromise<Resource> {
+    if (uri.scheme !== PluginUri.SCHEME) {
+      throw new Error('Not a plugin resource');
```
benoitf, line 62: Provides the uri that is invalid ?

extensions/eclipse-che-theia-plugin-remote/src/common/plugin-uri.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ ***********************************************************************/
+
+/**
+ * Share the definition of plugin resource uris between back end and
+ * front end.
+ */
+export namespace PluginUri {
+  export const SCHEME = 'pluginresource';
```
benoitf, line 16: if it's specific to che, probably add che into the SCHEME (in case upstream provides pluginresource as well) to avoid any mismatch

extensions/eclipse-che-theia-plugin-remote/src/node/che-plugin-scanner.ts
```diff
@@ -0,0 +1,53 @@
...
+  const result: SnippetContribution[] = [];
+  for (const contribution of pck.contributes.snippets) {
+    if (contribution.path) {
+      const absolutePath = path.join(pck.packagePath, contribution.path);
+      const normalizedPath = path.normalize(absolutePath);
+      const relativePath = path.relative(pck.packagePath, normalizedPath);
+      result.push({
+        language: contribution.language,
+        source: pck.displayName || pck.name,
+        uri: PluginUri.createUri(getPluginId(pck), relativePath),
```
benoitf, line 48: we should just simply make a SnippetContributionProviderURI or SnippetContributionProvider singleton upstream and then just overrides how we create URI downstream
Here there are too many overrides

extensions/eclipse-che-theia-plugin-remote/src/browser/plugin-resource.ts
```diff
@@ -0,0 +1,66 @@
...
+import URI from '@theia/core/lib/common/uri';
+import { injectable } from 'inversify';
+
+/**
+ * A resource for loading plugin resources. Uris will be of the form
+ * pluginresource://<plugin id>/path/relative/to/plugin/location
+ * The path is resolve relative to the directory designated by the
+ * plugins package path.
+ */
+export class PluginResource implements Resource {
```
benoitf, line 23: Should be a ChePluginResource

benoitf (changes requested):

tsmaeder:
[crw-ci-test]

dmytro-ndp:
[crw-ci-test --rebuild]

extensions/eclipse-che-theia-plugin-remote/src/common/plugin-uri.ts
tsmaeder: It's not really specific to che: the plugin resources at <che host>/hostedPlugin/... are also available upstream.

extensions/eclipse-che-theia-plugin-remote/src/node/che-plugin-scanner.ts
tsmaeder: What do you mean, there are too many overrides? It's a single method. If we extract the various contribution providers upstream, we would logically have to to it for all kinds of contributions. That may or may not be desirable, but one of the reasons we're doing this downstream is that the work was deemed urgent.

extensions/eclipse-che-theia-plugin-remote/src/browser/plugin-resource.ts
tsmaeder: No, it's not Che specific, even if it lives inside che-theia: same reasoning as above. I'm open to proposing a PR upstream with the plugin resource stuff.

extensions/eclipse-che-theia-plugin-remote/src/browser/plugin-resource.ts
tsmaeder: True. Not that throwing an exception is just the way for the plugin resolver to indicate it cannot handle this scheme. The error is swallowed silently.

extensions/eclipse-che-theia-plugin-remote/src/common/plugin-uri.ts
benoitf: it's tied to che because this protocol is not available upstream

extensions/eclipse-che-theia-plugin-remote/src/node/che-plugin-scanner.ts
benoitf: I mean here that there are too many classes being extended (and we may miss some of them)
while we're only interested on delegating the contribution uri.

> That may or may not be desirable, but one of the reasons we're doing this downstream is that the work was deemed urgent.

https://github.com/eclipse/che/issues/16502 is opened since march
I think that adding a new simple method in upstream theia to handle the contribution can be accepted in a couple of days as it's a very minor change.

extensions/eclipse-che-theia-plugin-remote/src/browser/plugin-resource.ts
benoitf: if it's becoming an upstream yes it'll be a pluginResource else it'll be a Che plugin resource.
Proposing it upstream will be better if you want to stick to 'pluginResource'

benoitf: But I think it's only useful upstream if this protocol is used somewhere

azatsarynnyy:
[crw-ci-test --rebuild]

extensions/eclipse-che-theia-plugin-remote/src/node/che-plugin-scanner.ts
ericwill: As far as I understand it, this issue is not reproducible in upstream theia. In the interest of saving time, I suggest we keep the fix downstream: not only because the benefit for theia is low, but also because the holidays are coming and getting the PR merged on time is not guaranteed. We can always iterate later if the need for such methods in theia arises.

benoitf: 'sdon't think it would be a problem to have something merge upstream if it's only an helper to get an URI for a plugin resource. Upstream theia having implementation with File and downstream with something else.
Discussing it is taking more time than having proposed a simple tiny PR url upstream.

Also I see two different ways being implemented for getting resources for a plug-in this one and https://github.com/eclipse/che-theia/pull/946 which is kind of colliding.

tsmaeder: I asked explicitly on a Theia dev call whether separation of plugin host and back end on different machines was something that should be supported in Theia and the answer was no. Hence the downstream effort instead of pursuing https://github.com/eclipse-theia/theia/pull/8468

tsmaeder: And again, we can push for this refactoring upstream any time later, but we already have all the necessary parts to fix the issue.

benoitf: It's different. It's not to have a plug-in-resource: protocol upstream, just to have something than convert a plug-in path to an URI by using File and that can be replaced in Che (with protocols, etc)

dmytro-ndp:
[crw-ci-test --rebuild]

tsmaeder:
I've tried using the sidecar file system to load snippet files, but that does not work because monaco tries to fetch the snippet files before the sidecar files system is registered.

benoitf:
@tsmaeder we can't register the filesystem earlier ?

tsmaeder:
@benoitf not easily: the file system is initialized when the plugin manager for a particular front-end is initialized. However, the initialization of connections to plugin hosts is done in parallel to opening the editor, it seems. I believe we would have to rework the lifecycle of plugin host connections. T.b.h, I think it would be beneficial to the project, but might introduce bugs and will delay startup.

tsmaeder:
@benoitf could we move forward with this PR? It works and your suggestion of introducing a `SnippetContributionProviderURI ` will not work until we make the plugin host file system provider work earlier as I described, because the upstream still uses the file system to load the snippets.

benoitf:
>  will not work until we make the plugin host file system provider work earlier as I described, because the upstream still uses the file system to load the snippets.

as compromise, I'm fine to keep the logic of custom URI on Che-Theia, but not having to override multiple scanners / copy almost all content of upstream readSnippets

tsmaeder:
> as compromise, I'm fine to keep the logic of custom URI on Che-Theia
@benoitf What do you mean by "logic of custom URI"? Do you mean using a resource instead of the file service to read the contents of 'snippets.json'? Or do you mean the generation of uri's from extension contributions like 'contributes.snippets'?

benoitf:
@tsmaeder to keep ChePluginResource and CheSnippetSuggestProvider, but remove the need of CheTheiaPluginScanner, CheVsCodePluginScanner by implementing upstream the delegation of readSnippet URI (SnippetContributionProviderURI)

extensions/eclipse-che-theia-plugin-remote/src/node/che-plugin-scanner.ts
sunix: I agree with @benoitf. Here using inheritance and overrides could be quite hard to maintain in the future. It is overriding a complete method and copying some parts. So somehow, upstream someone change that line https://github.com/eclipse-theia/theia/blob/bd25d737742a0974b911da5d3a5c978d950fc64e/packages/monaco/src/browser/monaco-snippet-suggest-provider.ts#L146, it would not be reflected downstream and could cause bugs that are hard to track.

azatsarynnyy (commented):
I tested the changes - snippets work well:

try

extensions/eclipse-che-theia-plugin-remote/src/browser/che-plugin-resource.ts
```diff
@@ -0,0 +1,66 @@
+/**********************************************************************
+ * Copyright (c) 2020 Red Hat, Inc.
```
benoitf, line 2:
```suggestion
 * Copyright (c) 2021 Red Hat, Inc.
```

extensions/eclipse-che-theia-plugin-remote/src/browser/che-snippet-suggest-provider.ts
```diff
@@ -0,0 +1,47 @@
+/**********************************************************************
+ * Copyright (c) 2020 Red Hat, Inc.
```
benoitf, line 2:
```suggestion
 * Copyright (c) 2021 Red Hat, Inc.
```

extensions/eclipse-che-theia-plugin-remote/src/common/che-plugin-uri.ts
```diff
@@ -0,0 +1,23 @@
+/**********************************************************************
+ * Copyright (c) 2020 Red Hat, Inc.
```
benoitf, line 2:
```suggestion
 * Copyright (c) 2021 Red Hat, Inc.
```

benoitf (approved):
works as expected. nice 👍 

https://user-images.githubusercontent.com/436777/107777859-aac1dc00-6d43-11eb-82e4-9a8776385650.mp4

azatsarynnyy (dismissed):
The code looks good to me. The functionality I've tested earlier https://github.com/eclipse/che-theia/pull/948#pullrequestreview-589490006
Thanks!

sunix (approved):
Good to merge :)

tsmaeder:
[crw-ci-test]

benoitf:
@tsmaeder as Happy path running in github action succeed:  `happy-path (alpine, stable)` you're free to merge (happy path is not stable enough to have all mode being green :-/ )

azatsarynnyy:
This patch breaks the contributed themes support.

There're only three themes provided by the Theia extensions:
image
The themes contributed by the built-ins are not loaded.

There are errors in the browser's console:
image

tsmaeder:
@azatsarynnyy good catch. I hadn't thought of replacing the files system stuff in themes, etc.

tsmaeder:
I think the theming (and others) probably was broken for anything that does not run in the theia container before. Now it's also broken for built-in plugins.

azatsarynnyy:
> @azatsarynnyy good catch. I hadn't thought of replacing the files system stuff in themes, etc.

I just thought about how it would load the contributed resources from the local file system via `chepluginresource://` scheme.
And had noticed something interesting :slightly_smiling_face:

ericwill:
If we want this in Che 7.26, we'll need to backport this PR and also fork theia at [this commit](https://github.com/eclipse/che-theia/blob/7.26.x/build.include#L17) and cherry pick the upstream fix onto our fork.

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-backend-module.ts
```diff
@@ -43,7 +47,8 @@ export default new ContainerModule((bind, unbind, isBound, rebind) => {
   bind(HostedPluginMapping).toSelf().inSingletonScope();
   bind(MetadataProcessor).to(RemoteMetadataProcessor).inSingletonScope();
   bind(PluginReaderExtension).toSelf().inSingletonScope();
-
+  rebind(HostedPluginReader).to(PluginReaderExtension).inSingletonScope();
```
benoitf, line 50: shouldn't it be toService(PluginReaderExtension) to use the same instance ? (and then we don't need to cast pluginReaderExtension as any type to call setRemotePluginConnection on it ?
else it looks odd

extensions/eclipse-che-theia-plugin-remote/package.json
```diff
@@ -11,6 +11,7 @@
     "@theia/core": "next",
     "@theia/plugin-ext": "next",
     "@theia/plugin-ext-vscode": "next",
+    "@theia/monaco": "next",
```
benoitf, line 14: do we still need this dependency ?

extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts
```diff
@@ -171,14 +183,39 @@ export class HostedPluginRemote {
...
-      const resourceBase64 = jsonMessage.data;
-      const resource = resourceBase64 ? Buffer.from(resourceBase64, 'base64') : undefined;
-      this.onGetResourceResponse(jsonMessage['pluginId'], jsonMessage['path'], resource);
+    if (InternalRequestResponsePayload.is(jsonMessage)) {
+      this.onInternalRequestResponse(jsonMessage);
       return;
     }
   }
 
+  requestPluginResourceStat(pluginId: string, resourcePath: string): Promise<Stat> {
```
benoitf, line 192: for readability maybe it would be easier to flag it as async method (and use await instead of chaining promises)

extensions/eclipse-che-theia-plugin-remote/src/node/plugin-remote-backend-module.ts
tsmaeder: I tried that and it did didn't work. T.b.h: I have no idea what "toService()" really does and inversify documentation didn't enlighten me.

benoitf: it's to bind the same instance under different interfaces (and not having two different singletons intances)
Here it looks like code smell, we have two different instances of the same class bound on different interfaces.

extensions/eclipse-che-theia-plugin-remote/src/node/hosted-plugin-remote.ts
tsmaeder: I used to like the async/await syntax, but t.b.h these days I find it makes following the control flow rather confusing, in particular when debugging. I tend to only use async/await where it really improves readability. Here it does not, IMO:

benoitf:
as it's a bugfix, would be nice to have unit tests
```
- [ ]:  Tests are covering the bugfix
```
all modules having now jest setup.

tsmaeder:
@benoitf since this is a bug that rises from the integration of multiple components (containers, even), I don't see an easy way to write a unit test that would be meaningful.
On a more general level, requiring regression tests for every bugfix does not yield sufficient quality gains to justify the expense in development time, IMO. Investing in reducing the complexity or our codebase and better manual testing possibilities would have a better return on investment. (That and making everyone use Che to develop Che).

PR #955, opened 2020-12-17
linkless build

tsmaeder:

### What does this PR do?
This is a proof of concept of a build process for che-theia that does not involve linking source folders: linking source folders introduces some problems because tools that manipulate paths will either see /projects/theia as a parent of a che-theia package or not, depending on whether the tools uses resolved file system paths or just manipulates the path string.

The idea of the PR is to move the `assembly` folder as a yarn workspace folder inside the che-theia repo and to yarn link all @theia packages at the level of the che-theia repo root.

https://github.com/eclipse/che/issues/16852

### How to test this PR?
1. In a self-hosting workspace, checkout the PR branch of che-theia into /projects
2. Checkout master of theia into /projects
3. Run `yarn` inside theia; wait for the build to finish
4. cd into /projects/che-theia
5. run the script `linktheia.sh`
6. run `yarn`
7. To run the back-end: `cd assembly && yarn run start --hostname=0.0.0.0 --port=3010 --plugins=local-dir:/projects/theia/plugins,local-dir:/projects/che-theia/plugins`
8. To run the plugin host: 
   * run the `linktheia.sh` script in the plugin host container 
   * `cd /projects/che-theia/extensions/eclipse-che-theia-plugin-remote/lib/node && node plugin-remote.js`
9. Open the `theia-dev-flow` endpoint

if you want to build against the latest Theia release (instead of against master), you can just remove the `@theia` folder from `/projects/che-theia/node_modules` and rerun `yarn`.

### PR Checklist

### Reviewers

### Happy Path Channel

HAPPY_PATH_CHANNEL=next


tsmaeder:
@benoitf I would be interested if you see any major problems with my approach.

benoitf:
@tsmaeder I think you've done two steps in this PR (two birds, one stone) and usually when we mix some concerns it's hard to get it merge easily.

There are linkless and 'generator-less' as well (BTW there is still a script to run:  `linktheia.sh` script)

so maybe target first linkless (as for example an option in generator) so everyone can try and there is no such breakage on all existing jobs.
if committers like it, option becomes default

and if committers don't like generator, we could remove it as well (another PR/step/issue)

tsmaeder:
@benoitf I'm looking for a validation of the concept before I update the generator. I was able to come up with a PoC in 1-2 days: fixing the generator to work would take me longer (I'd have to dive into it first).

benoitf:
@tsmaeder if it's introduced as a new flag I don't see any issue at all

sunix:
@tsmaeder do you have a devfile that would setup that POC ?

tsmaeder:
@sunix the one [here](https://github.com/eclipse/che/issues/18763#issuecomment-758510591) works fine

sunix:
> @sunix the one [here](https://github.com/eclipse/che/issues/18763#issuecomment-758510591) works fine

So that one with `theia` and `che-theia (to this fork/branch)` side by side

tsmaeder:
@tsmaeder just rm -rf theia and che-theia.

sunix:
step 8. maybe having a volume for HOME would help not to link twice.

assembly/package.json
```diff
@@ -70,7 +70,7 @@
   "scripts": {
     "prepare": "yarn run clean && yarn build",
     "clean": "theia clean && rimraf errorShots",
-    "build": "theia build --mode development --config cdn/webpack.config.js --env.cdn=./cdn.json --env.monacopkg=@theia/monaco-editor-core@0.20.0 && yarn run override-vs-loader",
+    "build": "NODE_OPTIONS='--max-old-space-size=4096' theia build --mode development --config cdn/webpack.config.js --env.cdn=./cdn.json --env.monacopkg=@theia/monaco-editor-core@0.20.0 && yarn run override-vs-loader",
```
benoitf, line 73: I would move it to `.github/workflows/pr.yml` ENV variable

tsmaeder: I've thought about this a bit more, and I think it's correct to set this only for the theia build: the size of memory needed depends on the task being executed, not the environment. Other tasks don't need that much max memory, and giving it to them can make them grab more than they need and take the memory away from other processes, thus reducing potential concurrency.

generator/src/templates/assembly-compile-new.tsconfig.json
```diff
@@ -0,0 +1,55 @@
+{
```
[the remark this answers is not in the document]
tsmaeder: Yes, but since we're not manipulating the theia/package.json, we don't have to. This PR is a first step that should allow folks to use the linkless build process in day-to-day development. Once we've changed the docker build scripts to use the new process, the init-sources files will go away.

plugins/github-auth-plugin/yarn.lock
```diff
@@ -0,0 +1,1494 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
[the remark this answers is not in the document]
tsmaeder: Not sure how that got here.

generator/src/templates/assembly-package-new.mst.json
```diff
@@ -0,0 +1,166 @@
+{
```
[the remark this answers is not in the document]
tsmaeder: I tried going that way, since the manipulation of the theia/package.json requires much more complicated manipulation that is not needed in the linkless case, I decided it would be simplest to introduce new templates.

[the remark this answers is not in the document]
tsmaeder: I don't think we need to generate the assembly as part of the build process at all, t.b.h. But I decided leave that for a later PR to keep this smaller.

generator/src/templates/assembly-package-new.mst.json
```diff
@@ -0,0 +1,166 @@
...
+    "coverage": "yarn coverage:compile && yarn test && yarn coverage:remap && yarn coverage:report:lcov && yarn coverage:report:html",
+    "download:plugins": "theia download:plugins"
+  },
+  "devDependencies": {
+    "@theia/cli": "next",
+    "html-webpack-plugin": "^3.2.0",
+    "async-limiter": "^2.0.0"
+  },
+  "theiaPluginsDir": "../plugins",
+  "theiaPlugins": {
```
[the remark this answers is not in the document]
tsmaeder: Again, this is not meant as a long term solution: theiaPlugins.json should go away in the medium term.

generator/src/templates/assembly-compile-new.tsconfig.json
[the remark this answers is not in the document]
benoitf: having two definition for the same stuff is always breaking. People only modify at a single place.
I would not reproduce the nightmare it was on plug-in registry when there was both all meta.yaml and vscode-extensions.json in parallel.

Also I would not opt-in for `-new` as there is no '-old' assembly, just that they're different

generator/src/templates/assembly-package-new.mst.json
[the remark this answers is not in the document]
benoitf: we should not introduce complete new files, we should reuse at maxium the templates.

If assembly folder needs to be committed one day, then generated files would just have to be committed, but during that time we don't have several builds in parallel

It was the goal of using templates and not using plain files (where we do copy of copy for every usecase)

generator/src/templates/assembly-package-new.mst.json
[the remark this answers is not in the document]
benoitf: to me it does not look a priority at all to have it by default

tsmaeder:
[crw-ci-test --rebuild]

generator/src/json-utils.ts
```diff
@@ -0,0 +1,38 @@
+/**********************************************************************
+ * Copyright (c) 2018-2020 Red Hat, Inc.
```
benoitf, line 2:
```suggestion
 * Copyright (c) 2021 Red Hat, Inc.
```

generator/src/link.ts
```diff
@@ -0,0 +1,78 @@
+/**********************************************************************
+ * Copyright (c) 2018-2020 Red Hat, Inc.
```
benoitf, line 2:
```suggestion
 * Copyright (c) 2021 Red Hat, Inc.
```

generator/src/json-utils.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Simple code to rewrite parts of JSON files
+ * @author Thomas Mäder
+ */
+
+import * as fs from 'fs-extra';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export async function rewriteJson(packageJSONPath: string, rewriteFunction: (json: any) => void) {
+    const json = fs.readJSONSync(packageJSONPath);
```
benoitf, line 20:
```suggestion
    const json = await fs.readJSON(packageJSONPath);
```

generator/src/link.ts
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ ***********************************************************************/
+
+/**
+ * A command to yarn link theia dependencies into che-theia
+ * @author Thomas Mäder
+ */
+
+import * as fsextra from 'fs-extra';
```
benoitf, line 16: why not only `fs` ?

generator/src/link.ts
```diff
@@ -0,0 +1,78 @@
...
+    const cfg = await new Command(cheTheiaDir).exec('yarn --silent --json --non-interactive config current');
+
+    try {
+        const yarnConfig = JSON.parse(JSON.parse(cfg).data);
+        const linkDir = yarnConfig['linkFolder'] || path.resolve(os.homedir(), '.yarn/link');
+
+        await link(cheTheiaDir, theiaDir, linkDir);
+    } catch (e) {
+        console.info('first char is ' + cfg.charCodeAt(0));
+        console.error('thsi is the start' + cfg + 'stop');
```
benoitf, line 52:
```suggestion
        console.error('this is the start' + cfg + 'stop');
```

generator/src/link.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    const cfg = await new Command(cheTheiaDir).exec('yarn --silent --json --non-interactive config current');
+
+    try {
+        const yarnConfig = JSON.parse(JSON.parse(cfg).data);
+        const linkDir = yarnConfig['linkFolder'] || path.resolve(os.homedir(), '.yarn/link');
+
+        await link(cheTheiaDir, theiaDir, linkDir);
+    } catch (e) {
+        console.info('first char is ' + cfg.charCodeAt(0));
```
benoitf, line 51: not sure about this trace ?

generator/src/update-dependencies.ts
```diff
@@ -0,0 +1,88 @@
+/**********************************************************************
+ * Copyright (c) 2018-2020 Red Hat, Inc.
```
benoitf, line 2:
```suggestion
 * Copyright (c) 2021 Red Hat, Inc.
```

generator/src/update-dependencies.ts
```diff
@@ -0,0 +1,88 @@
...
+    replaceVersion: (packageName: string) => string | undefined
+) {
+    try {
+        await rewriteJson(packageJSONPath, pkgJson => {
+            // we're assuming the package.json is well formed
+            replaceInSection(pkgJson.dependencies, replaceVersion);
+            replaceInSection(pkgJson.devDependencies, replaceVersion);
+        });
+    } catch (e) {
+        console.warn('could not find package json: ' + packageJSONPath);
```
benoitf, line 71:
```suggestion
        console.warn(`could not find package json: ${packageJSONPath}`);
```

generator/tests/link.spec.ts
```diff
@@ -0,0 +1,52 @@
+/**********************************************************************
+ * Copyright (c) 2018-2020 Red Hat, Inc.
```
benoitf, line 2:
```suggestion
 * Copyright (c) 2021 Red Hat, Inc.
```

package.json
```diff
@@ -31,7 +31,7 @@
     "typescript": "~4.0.5"
   },
   "scripts": {
-    "prepare": "lerna run prepare",
+    "prepare": "lerna --concurrency=2 run prepare",
```
benoitf, line 34: maybe move out to a separate PR/issue ?
or have a way to provide parameters/override it

plugins/github-auth-plugin/yarn.lock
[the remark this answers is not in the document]
benoitf: this file should be removed then

linkless-build.md
```diff
@@ -0,0 +1,46 @@
...
+3. Link into che-theia
+
+        cd /projects/che-theia
+        che-theia link
+   This will `yarn link` all necessary modules from the Theia source into the node modules folder inside the 
+   che-theia project. At this time, it's a good idea to rebuild the che-theia project:
+4. Rebuild che-theia
+
+        rm yarn.lock
+        yarn
```
benoitf, line 46: I think the best option is to provide a devfile.yaml using linkless build commands there so it's easy to anyone to try it out. (but can be a follow-up PR)

generator/src/link.ts
tsmaeder: Conversely, why import fs-extra as 'fs'?

benoitf: because it's a drop-in replacement (and all over theia source code it's currently imported as 'fs')

tsmaeder: As I understand it, it has more API than fs, so I wouldn't call it a drop-in replacement, but the "convention" argument is fair enough.

benoitf: it's how it is advertised
https://github.com/jprichardson/node-fs-extra#usage

> fs-extra is a drop in replacement for native fs

generator/src/update-dependencies.ts
tsmaeder: It really doesn't matter. What makes you think string interpolation is intrinsically better in this case?

benoitf: https://rules.sonarsource.com/typescript/RSPEC-3512

tsmaeder: So there's no real reason, it's just the opinion of some person at sonarsource that it's clearer? I disagree 🤷

benoitf: if people run analysis tools on repositories, they'll look only at number of "code smell" reports and won't look deeply if it is related to Thomas agreement or not.

tsmaeder: As far as I know, there is no requirement to heed sonarsource warnings on the Eclipse che-theia project. Was there a decision by the development team I'm not aware of? @ericwill maybe you have more info?

ericwill: For a print statement/warning I would say it's a question of code style as nothing is concretely impacted by doing it one or the other (in this case). I personally would have opted for the template string but it's not a hill I think we need to die on.

@benoitf if you think we should have sonarlint run on PRs then let's open an issue for that and discuss it there

benoitf:
FYI I've an error on the link test:

```
 FAIL  tests/link.spec.ts
  ● Console

    console.error
      Error: Unable to execute the command yarn link @theia/bar: Error: Command failed: yarn link @theia/bar
error No registered package found called "@theia/bar".

          at new CliError (/tmp/che-theia/generator/src/cli-error.ts:17:9)
          at /tmp/che-theia/generator/src/command.ts:34:32
          at ChildProcess.exithandler (child_process.js:315:5)
          at ChildProcess.emit (events.js:314:20)
          at maybeClose (internal/child_process.js:1022:16)
          at Process.ChildProcess._handle.onexit (internal/child_process.js:287:5)

      49 |         await link(cheTheiaDir, theiaDir, linkDir);
      50 |     } catch (e) {
    > 51 |         console.error(e);
         |                 ^
      52 |     }
      53 | }
      54 |

      at Object.<anonymous> (src/link.ts:51:17)
      at step (src/link.ts:5576:19)
      at Object.throw (src/link.ts:5306:14)
      at rejected (src/link.ts:5209:32)

  ● Test link command › test link

    expect(received).toBeTruthy()

    Received: false

      46 |         const zozPackage = path.resolve(srcFolder, 'node_modules/@theia/zoz/package.json');
      47 |
    > 48 |         expect(fs.existsSync(fooPackage)).toBeTruthy();
         |                                           ^
      49 |         expect(fs.existsSync(barPackage)).toBeTruthy();
      50 |         expect(fs.existsSync(zozPackage)).toBeTruthy();

```

tsmaeder:
@benoitf I can't reproduce the test failure. How are you running the tests?

benoitf:
@tsmaeder I think it's related to symlinks where theia is cloned
I've created https://github.com/tsmaeder/che-theia/pull/2 and it works for me with that configuration

tsmaeder:
So are you saying your link folder is itself in a symlinked location?

benoitf:
> So are you saying your link folder is itself in a symlinked location?

@tsmaeder yes

benoitf:
it's a bug in yarn but unfortunately it's not yet solved (https://github.com/yarnpkg/yarn/pull/7028 and https://github.com/yarnpkg/yarn/issues/1297)

benoitf:
@tsmaeder could you please rebase/resolve the conflict ? so we can have PR checks running

benoitf (approved):
👍

PR #959, opened 2021-01-05
fix(workpace-plugin): Importing several private projects in a single devfile

sunix:
### What does this PR do?
- Fix workspace start initial git cloning when multiple private projects are defined in the devfile.
- Only ask for credentials once if cloning from the same host.
- Adding a `devfile.yaml` for the workspace plugin (simplify review).
- ~Fix jest import errors in Che~
- Include unit tests

Fix https://github.com/eclipse/che/issues/18494

### How to test this PR?
1. create 2 private projects in Github with README files
2. start a workspace from the given devfile.
3. build the plugin with the predifined tasks
4. run che-theia with the predefined task
5. go to the dashboard and add the private projects to the devfile of the running workspace. For instance (adapt the location with the ones you have created):
    ```
      - name: private-1
        source:
          location: 'https://github.com/sunix/test-private-1'
          type: github
      - name: private-2
        source:
          location: 'https://github.com/sunix/test-private-2'
          type: github
    ```
    SAVE the devfile DO NOT APPLY, DO NOT RESTART THE WORKSPACE!
6. back to che-theia, open the preview Url in a new tab, try to fill the username and password, observe the bug
7. stop the run task and close the preview tab, remove the content of `/tmp/theiadev_projects` from a terminal of che-theia-next-dev cotainer.
8. check the workspace devfile (dashboard) that the private projects are still there.
9. from the Github PR panel, checkout this PR
10. rebuild with the predifined task
11. rerun che-theia dev with the predifined task
12. open the Preview Url in a new tab.
13. fill your Github.com username and password, it should ask for it only once.
14. observe that the projects are cloned with the README files.

### PR Checklist

- [x] [Tests are covering the bugfix](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#tests-are-covering-the-bugfix)
- [x] [The repository devfile is up to date and works](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-repository-devfile-is-up-to-date-and-works)
- [x] [Relevant user documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
- [x] [Relevant contributing documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
- [x] [CI/CD changes implemented, documented and communicated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#cicd-changes-implemented-documented-and-communicated)

### Reviewers

### Happy Path Channel

HAPPY_PATH_CHANNEL=next

### DIFF
--- a/plugins/workspace-plugin/devfile.yaml
+++ b/plugins/workspace-plugin/devfile.yaml
@@ -1,0 +1,72 @@
+apiVersion: 1.0.0
+
+metadata:
+  generateName: che-theia-workspace-plugin-
+
+projects:
+  - name: che-theia
+    source:
+      location: 'https://github.com/eclipse/che-theia.git'
+      type: git
+
+components:
+  - mountSources: true
+    command:
+      - tail
+      - '-f'
+      - /dev/null
+    memoryLimit: 512Gi
+    type: dockerimage
+    image: 'quay.io/eclipse/che-theia:next'
+    alias: che-theia-next-dev
+    env:
+      - value: 0.0.0.0
+        name: THEIA_HOST
+      - value: '3130'
+        name: THEIA_PORT
+      - value: '0'
+        name: NODE_TLS_REJECT_UNAUTHORIZED
+
+  - mountSources: true
+    memoryLimit: 3Gi
+    type: dockerimage
+    image: 'quay.io/eclipse/che-theia-dev:next'
+    alias: che-dev
+
+  - id: redhat/vscode-yaml/latest
+    type: chePlugin
+
+  - id: che-incubator/typescript/latest
+    memoryLimit: 2048M
+    type: chePlugin
+
+  - id: ms-vscode/vscode-github-pullrequest/latest
+    type: chePlugin
+
+commands:
+
+  - name: build ... workspace-plugin
+    actions:
+      - workdir: /projects/che-theia/plugins/workspace-plugin
+        type: exec
+        command: |
+          killall node; yarn || (yarn lint:fix && yarn format:fix && yarn) && echo -e "\e[32mDone.\e[0m build ... workspace-plugin"
+        component: che-dev
+
+  - name: test-watch ... workspace-plugin
+    actions:
+      - workdir: /projects/che-theia/plugins/workspace-plugin
+        type: exec
+        command: |
+          killall node; yarn test:watch
+        component: che-dev
+
+  - name: run ... che-theia + workspace-plugin
+    previewUrl:
+      port: 3130
+    actions:
+      - workdir: /home/theia
+        type: exec
+        command: |
+          rm /default-theia-plugins/eclipse_che_workspace_plugin.theia; mkdir -p /tmp/theiadev_projects && export CHE_PROJECTS_ROOT=/tmp/theiadev_projects && cp /projects/che-theia/plugins/workspace-plugin/eclipse_che_workspace_plugin.theia /default-theia-plugins/ && /entrypoint.sh
+        component: che-theia-next-dev
--- a/plugins/workspace-plugin/src/askpass-prompt-manager.ts
+++ b/plugins/workspace-plugin/src/askpass-prompt-manager.ts
@@ -1,0 +1,44 @@
+/**********************************************************************
+ * Copyright (c) 2019-2021 Red Hat, Inc.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ ***********************************************************************/
+
+export class PromptManager {
+  private askPassResults: Map<HostPlaceHolderKey, PromiseLike<string | undefined>> = new Map();
+
+  constructor(promptLauncher: (host: string, placeHolder: string) => PromiseLike<string | undefined>) {
+    this.askPassPromptLauncher = promptLauncher;
+  }
+
+  async askPass(host: string, placeHolder: string): Promise<string> {
+    let existingPromise: PromiseLike<string | undefined> | undefined;
+
+    // wait for all promises to be finished sequencially
+    this.askPassResults.forEach(async (promise, key) => {
+      if (key.host === host && key.placeHolder === placeHolder) {
+        existingPromise = promise;
+      }
+      await promise;
+    });
+
+    if (existingPromise) {
+      return (await existingPromise) || '';
+    }
+
+    const askPassPromise = this.askPassPromptLauncher(host, placeHolder);
+    this.askPassResults.set({ host: host, placeHolder: placeHolder }, askPassPromise);
+    return (await askPassPromise) || '';
+  }
+
+  askPassPromptLauncher: (host: string, placeHolder: string) => PromiseLike<string | undefined>;
+}
+
+class HostPlaceHolderKey {
+  host: string;
+  placeHolder: string;
+}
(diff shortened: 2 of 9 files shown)


sunix:
[crw-ci-test]

benoitf:
[crw-ci-test]

benoitf:
FYI I switched happy path channel to `next`

plugins/workspace-plugin/devfile.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: 1.0.0
```
benoitf, line 1: I'm not sure it's a good practice to introduce devfiles in each directory ?

plugins/workspace-plugin/tests/file-uri.spec.ts
```diff
@@ -8,6 +8,8 @@
  * SPDX-License-Identifier: EPL-2.0
  ***********************************************************************/
 
+import 'jest';
```
benoitf, line 11: looks odd to import `jest` ?

sunix: This was the only way I could get rid of the errors reported by the typescript plugin.

plugins/workspace-plugin/devfile.yaml
sunix: This a devfile to work on the workspace plugin, as this project lives in a subdirectory of the git repo, it looks good to have it there, next to the package.json, etc ...

plugins/workspace-plugin/tests/file-uri.spec.ts
benoitf: but jest is added on tsconfig types

plugins/workspace-plugin/devfile.yaml
benoitf: but it's part of a yarn workspace so you should use top-level folder/definitions/devfile anyway to inherits from all workspace definition

sunix: ok but isn't it overkill when a dev just want to work on 1 plugin ?

ericwill: It might be useful to have the devfile here, even just for reference purposes. It could be useful to point for new contributors in the future.

benoitf: On a maven project you won't have devfile all over the places in every modules as well
Also we do need parent package.json
It would be nightmare to maintain all these devfiles

ericwill: > It would be nightmare to maintain all these devfiles

That's a good point, better leave it out then @sunix

plugins/workspace-plugin/src/askpass-prompt-manager.ts
```diff
@@ -0,0 +1,44 @@
...
+    });
+
+    if (existingPromise) {
+      return (await existingPromise) || '';
+    }
+
+    const askPassPromise = this.askPassPromptLauncher(host, placeHolder);
+    this.askPassResults.set({ host: host, placeHolder: placeHolder }, askPassPromise);
+    return (await askPassPromise) || '';
+  }
```
benoitf, line 36: it looks like code smell

I'm not sure what it's trying to solve there 

Also it looks like there are too many await: for example in the forEach loop, we await all promises 
and then if there is an existing promise we do await again on it

plugins/workspace-plugin/devfile.yaml
sunix: no but the devfiles are maintained as their are used. And for maven submodules, it really depends if a submodule is to be used independently. Take the old che6 maven structure, we had a plugin folder. each plugin could be built independently.

plugins/workspace-plugin/src/askpass-prompt-manager.ts
sunix: What we are trying to solve: making sure the promises are done and executed in sequence. OK to try to remove the await of existing promise.

plugins/workspace-plugin/devfile.yaml
tsmaeder: I'll have to add my voice to the detractors here: I don't think there is a benefit to having a devfile for every single che-theia plugin: you can already rebuild single plugins and I don't really see any other upside, but a maintenance burden: each time we change something in the recommended plugin development process, we'll have to go over all those devfiles. We're not doing a good job keeping the devfiles in sync with best practices.
As for extensions: you have to build all of those together anyway, as changes in one may break others (compile-time):

benoitf: it looks like: hey the current devfiles are not so good so let's add a new one. I mean, let's take a contributor. What devfile pick up if I find there are 30 devfiles across different folders ?

Depending on working on a che-theia task, I could work only on one extension and one plug-in, so it would result in a new submitted devfile to work with these two modules ?

devfile can use generic commands based on the current selected folder, so if you work on plug-ins basically they all have same commands: lint, compile, format, test, etc.

sunix: This devfile is for coding the workspace plugin without having to rebuild che-theia and all the other plugins. It doesn't make sense to have it elsewhere.

sunix: The maintenance issue is not related to the location of the devfiles. but how a devfile could be extended: so the devfile for the workspace plugin (that would be in the workspace plugin folder) could extend a more generic devfile for built-in plugins but adding its own specifics (for instance, custom commands, etc ...).
The devfile for the workspace plugin has to stay here because it would setup the right workspace for the workspace-plugin developer flow.

sunix: and then if the suggestion is to move that devfile to `devfiles` I don't see how easier it would be to maintain. Or is the suggestion not to have devfile at all and let all the developers loosing time to setup an dev environment to work on each individual plugins?

tsmaeder: No, the suggestion is to have a a single devfile for developing che-theia.

sunix: there is that devfile https://github.com/eclipse/che-theia/blob/master/devfiles/che-theia-all.devfile.yaml but it is doing way too many things when it is to work a a particular plugin.

plugins/workspace-plugin/src/askpass-prompt-manager.ts
sunix: I have simplified ... using mutex.

plugins/workspace-plugin/tests/file-uri.spec.ts
sunix: no idea ... 
Selection_142 maybe a bug in typescript plugin.

Do you want to remove these imports?

benoitf: could you add
```
      "types": ["reflect-metadata", "jest"],
```

in https://github.com/eclipse/che-theia/blob/master/plugins/workspace-plugin/tsconfig.json ?

like it's done there: https://github.com/eclipse/che-theia/blob/master/plugins/ports-plugin/tsconfig.json#L12-L14

sunix: I tried that already (see the M in tsconfig.ts in the screenshoft above). And just retried but it doesn't get rid of the error message. Maybe a bug in the plugin.

sunix: @benoitf I rebased and removed the imports

benoitf:
@sunix there is still the devfile

sunix:
> @sunix there is still the devfile

I thought we were ok

1. to keep it for now
2. work on https://github.com/eclipse/che/issues/18763
3. remove it later on when we would  https://github.com/eclipse/che/issues/18763 as well as other devfiles located in devfiles.

benoitf (approved):

PR #995, opened 2021-02-16
Clone Git projects with SSH URI

vitaliy-guliy:
### What does this PR do?
Allows the workspace plugin to handle events when cloning Git projects by SSH uri
When the devfile contain GIT SSH project, the workspace plugin
- performs the secure login to Git repository
- in case of GitHub repository, proposes the user to upload public key
- proposes the user to add auto-generated key to the Git repository by hands
- allows the user to use its custom certificate

### Screenshot/screencast of this PR

Screenshot from 2021-02-22 14-19-56

Screenshot from 2021-02-22 14-20-04

https://github.com/eclipse/che/issues/13494

### HappyPath tests related on SSH
- https://github.com/eclipse/che/blob/master/tests/e2e/tests/e2e/GitSsh.spec.ts

### How to test this PR?

Devfile to test
```
---
apiVersion: 1.0.0
metadata:
  generateName: clone-ssh-
projects:
  -
    name: che-theia
    source:
      type: git
      location: "git@github.com:eclipse/che-theia.git"
  -
    name: atlascode
    source:
      type: git
      location: "git@bitbucket.org:atlassianlabs/atlascode.git"
components:
  - type: cheEditor
    alias: che-theia
    reference: https://raw.githubusercontent.com/chepullreq4/pr-check-files/master/che-theia/pr-995/che_theia_meta.yaml
    memoryLimit: 2G
```

### PR Checklist

### Reviewers

### Happy Path Channel

HAPPY_PATH_CHANNEL=stable


plugins/workspace-plugin/src/theia-commands.ts
```diff
@@ -116,28 +119,100 @@ export class TheiaGitCloneCommand implements TheiaImportCommand {
...
+
+      // unable to login
+      // Give the user possible actions
+      // - retry the login
+      // - show SSH options
+
+      const RETRY = 'Retry';
+      const CONFIGURE_SSH = 'Configure SSH';
+
+      let message = `It is failing to clone Git project ${this.locationURI}`;
```
ericwill, line 172:
```suggestion
      let message = `Failure to clone Git project ${this.locationURI}`;
```

plugins/workspace-plugin/src/git.ts
```diff
@@ -100,3 +100,44 @@ export function getGitRootFolder(uri: string): string {
...
+      return error.message;
+    } else {
+      throw error;
+    }
+  }
+}
+
+export function getErrorReason(message: string): string | undefined {
+  if (message.indexOf('Permission denied (publickey)') >= 0) {
+    return 'It may be missing SSH key';
```
ericwill, line 139:
```suggestion
    return 'A valid SSH key may be missing';
```

sunix:
> Upload Private Key

It is not clear that it is to upload a private key to the che workspace. Isn't that `upload Key Pair` ? to be consistant with the other commands ?

> Upload existing key to Github

What about  `Add existing key to Github` ?

vitaliy-guliy:
> > Upload Private Key
> 
> It is not clear that it is to upload a private key to the che workspace. What about `Upload Key Pair` ? to be consistant with the other commands.

Probably yes, but as I understand, the command must upload a private key from your local machine and register it in the system(your workspace) with `ssh-add`.
I don't remember, have we discussed that or not, but you could want to generate a keypair on your machine and then use it in Che. You will be able to upload your private key to your workpspace and add correspodring public key to your Git account.

> 
> > Upload existing key to Github
> 
> What about `Add existing key to Github` ?

Renamed, thanks.

sunix:
> Probably yes, but as I understand, the command must upload a private key from your local machine and register it in the system(your workspace) with `ssh-add`.
> I don't remember, have we discussed that or not, but you could want to generate a keypair on your machine and then use it in Che. You will be able to upload your private key to your workpspace and add correspodring public key to your Git account.
> 

Usually, you generate a private key and a public key. upload the public key to github. But they generally come together in the generation phase. 
```
Generating public/private ed25519 key pair.
```
I would recommend to use the same terminology here https://docs.github.com/en/github/authenticating-to-github/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent

sunix:
BTW what is the difference between `Generate Key pair`, `Create key pair`, and `upload private key` ?

vitaliy-guliy:
> BTW what is the difference between `Generate Key pair`, `Create key pair`, and `upload private key` ?

It seems `Create key pair` creates key pair from the content, provided by user. You have to enter content of public and private keys. @vinokurig am I right? It's pretty similar to `Upload Private Key..`, but creates a pair instead of uploading and registering one file.

`Generate Key pair` creates a new SSH key using `ssh-keygen`.

`Upload Private Key` uploads and registers your private key (without creating a key pair).

vitaliy-guliy:
> I would recommend to use the same terminology here https://docs.github.com/en/github/authenticating-to-github/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent.

Then to be more clear, we need to rename a couple of commands (remove Pair from the command name)

`SSH: Generate Key Pair For Particular Host...` -> `SSH: Generate Key For Particular Host...`
`SSH: Generate Key Pair...` -> `SSH: Generate Key...`
`SSH: Create Key Pair...` -> `SSH: Create Key...` (do we really need this command?)
`SSH: Delete Key Pair...` -> `SSH: Delete Key...`

And I would rework `SSH: Upload Private Key...` to be able to upload both public and private keys.

sunix:
I don't remember, maybe it can regenerate a public key from the private key. So uploading private key would be enough. But give it a try with the complete flow and see if it works: what does it do when you upload a private key ? is this generating the public key automatically ?

vitaliy-guliy:
> I don't remember, maybe it can regenerate a public key from the private key. So uploading private key would be enough. But give it a try with the complete flow and see if it works: what does it do when you upload a private key ? is this generating the public key automatically ?

Based on source code, uploading of private key doesn't create a public key.

https://github.com/eclipse/che-theia/blob/8e61ae25ca90bdab2bba99a40fe48b6fa97c62ef/plugins/ssh-plugin/src/ssh-plugin-backend.ts#L288-L330

plugins/workspace-plugin/src/git.ts
```diff
@@ -100,3 +100,44 @@ export function getGitRootFolder(uri: string): string {
 export async function execGit(directory: string, ...args: string[]): Promise<string | undefined> {
   return execute('git', args, { cwd: directory });
 }
+
+export function isSecureGitURI(uri: string): boolean {
+  return uri.startsWith('git@');
+}
+
+export function isSecureGitGubURI(uri: string): boolean {
```
vinokurig, line 108: `isSecureGitHubURI` ?

vinokurig: or `isSecureGithubURI`

sunix:
@vitaliy-guliy I expect that you give a try with these different flow:

User starts a workspace with a devfile and a ssh github url
Flow#1: User wants Che to generate a new key pair and upload the public key to github
Flow#2: User wants to add his private key from his machine: upload the file(s). He has already uploaded the public key to github.
Flow#3: User wants to add his private key from his machine: copy paste to an editor. He has already uploaded the public key to github.

For flow#2 and #3, these are pretty similar. We have to see if keys with passphrase working in Che or not. If not, create a gh-issue to fix that.

plugins/workspace-plugin/src/git.ts
vitaliy-guliy: Fixed, thanks!

vinokurig:
@vitaliy-guliy
* If user has more than one SSH key what key will be used for adding to GitHub?
* When a workspace starts it has a predefined SSH key even if user hasn't created any so why not to propose user to add a key to GitHub rather than showing the full list of ssh commands?
* When I used `SSH: Add Existing Key To GitHub...` it doesn't show any notification about the success result of it's work.

vitaliy-guliy:
> @vitaliy-guliy
> 
> * If user has more than one SSH key what key will be used for adding to GitHub?

The first key from the list. It needs to be improved.

> * When a workspace starts it has a predefined SSH key even if user hasn't created any so why not to propose user to add a key to GitHub rather than showing the full list of ssh commands?

User may want to use his personal key, not the default one.

> * When I used `SSH: Add Existing Key To GitHub...` it doesn't show any notification about the success result of it's work.

I don't think we need to display extra notifications here. If the key is added and the user is authorized, why not just clone the project?
But this could be discussed.

vinokurig:
> User may want to use his personal key, not the default one.

In this case user can use the SSH commands from the command palette. I mean if user want's to use a more specific SSH key it means that the user has some experience in SSH in Che, so he knows how to do it. But if a new user see's a list of such SSH commands he might be confused.

> I don't think we need to display extra notifications here. If the key is added and the user is authorized, why not just clone the project?

I am talking about the `SSH: Add Existing Key To GitHub...` command itself, not as a part of the `workspace plugin` flow.

vitaliy-guliy:
> > User may want to use his personal key, not the default one.
> 
> In this case user can use the SSH commands from the command palette. I mean if user want's to use a more specific SSH key it means that the user has some experience in SSH in Che, so he knows how to do it. But if a new user see's a list of such SSH commands he might be confused.

I think it's even better that we are displaying all the options.
Here is another question, how the list of possible SSH commands make a bit understandable?
Having `Generate Key..` and `Create Key..` confuses even me.

> > I don't think we need to display extra notifications here. If the key is added and the user is authorized, why not just clone the project?
> 
> I am talking about the `SSH: Add Existing Key To GitHub...` command itself, not as a part of the `workspace plugin` flow.

This make sense. I will add a notification for that.

sunix:
I think in general, we have that menu when there is an error. So displaying all the possible options is what we should do because it could be anything.

vinokurig:
>I think it's even better that we are displaying all the options.

I think that such notifications should propose only a quick solution, but that's only my point of view :)
>Here is another question, how the list of possible SSH commands make a bit understandable?
Having Generate Key.. and Create Key.. confuses even me.

I don't think that we have to invent new names, it would be better to build a nice UI widget instead of having just a list of commands.

vinokurig:
@sunix 
> I think in general, we have that menu when there is an error. So displaying all the possible options is what we should do because it could be anything.

What do you mean by `it could be anything`? We can perform an action that will 100% work without opening anything.

sunix:
As far as I know it is not necessary a flow for github. What if it fails for bitbucket ?

vinokurig:
@sunix 
> As far as I know it is not necessary a flow for github. What if it fails for bitbucket ?

Yes you are right but we can check if the clone url is a GitHub url, otherwise we can show the command list. @vitaliy-guliy WDYT?

vitaliy-guliy:
> @sunix
> 
> > As far as I know it is not necessary a flow for github. What if it fails for bitbucket ?
> 
> Yes you are right but we can check if the clone url is a GitHub url, otherwise we can show the command list. @vitaliy-guliy WDYT?

For GitHub we display an item to add public key. But the user may want to use his private key. So we give here a bit more freedom.

vinokurig:
@vitaliy-guliy 
> For GitHub we display an item to add public key.

screenshot-che-eclipse-che 192 168 99 254 nip io-2021 02 18-18_06_55
I don't see it in the notification.

vitaliy-guliy:
> @vitaliy-guliy
> 
> > For GitHub we display an item to add public key.
> 
> screenshot-che-eclipse-che 192 168 99 254 nip io-2021 02 18-18_06_55
> I don't see it in the notification.

It will appear if you click `Configure SSH`. It did so intentionally to not overload the popup.

vinokurig:
> It will appear if you click Configure SSH. It did so intentionally to not overload the popup.

I know, I am talking about having an action to add the key to GitHub *IN* the notification, if the url is GitHub, instead of the list of commands. My point is that we already have a predefined SSH key so why not to use it? BTW when we added this predefined key the idea was to help unexperienced users and propose them to add the ssh key in the notification when they clone projects via SSH.

vitaliy-guliy:
> > It will appear if you click Configure SSH. It did so intentionally to not overload the popup.
> 
> I know, I am talking about having an action to add the key to GitHub _IN_ the notification, if the url is GitHub, instead of the list of commands. My point is that we already have a predefined SSH key so why not to use it? BTW when we added this predefined key the idea was to help unexperienced users and propose them to add the ssh key in the notification when they clone projects via SSH.

I think inexperienced user will retry cloning and then click on `Configure SSH` as there are no more buttons (except button to close the popup).
He could want to play a bit with SSH before uploading the key to GH, isn't he?

vitaliy-guliy:
@l0rd since you requested the SSH plugin some time ago, it would be nice to see your opinion.
WDYT about having a button to upload an SSH key to the GitHub next to `Configure SSH` button?

We are discussing here two ways
- not to overload the notification with dozen of buttons and use menu/command palette to manage SSH keys

configure-ssh

- do the same as above, but add 'Add Key To GitHub' when cloning from GH

configure-ssh

vitaliy-guliy:
> @vitaliy-guliy I expect that you give a try with these different flow:
> 
> User starts a workspace with a devfile and a ssh github url
> Flow#1: User wants Che to generate a new key pair and upload the public key to github
> Flow#2: User wants to add his private key from his machine: upload the file(s). He has already uploaded the public key to github.
> Flow#3: User wants to add his private key from his machine: copy paste to an editor. He has already uploaded the public key to github.
> 
> For flow#2 and flow#3, these are pretty similar. We have to see if keys with passphrase working in Che or not. If not, create a gh-issue to fix that.
> Also flow#2 and flow#3, check that upload/add private key is enough (don't need to upload a public key?)

Uploading of private keyfile does not work as expected. In fact, the file is uploaded, but need to investigate what's going when the plugins tries to register it with `ssh-add`.
I see that a keypair is created, but seems it has a wrong content.

See the issue https://github.com/eclipse/che/issues/19109 where I explained the problem and steps to reproduce.
(need additional time to fix that ^^)

BTW, when I uploaded and registered the file by hand, I was able to clone the repository. So, the flow is working. Just need to fix the bug.

`SSH: Create Key...` command is also not working. It creates a key from the content, provided by user, but it's not possible to clone the git repository. I think we will fix that withing fixing the bug with uploading of private key.

l0rd:
@vitaliy-guliy I like the option with the "Add Key to GitHub" cc @sympatheticmoose

sunix:
If someone has a git ssh url in the devfile, we expect him to understand these kind of doc: https://docs.github.com/en/github/authenticating-to-github/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent

For those people, giving the list of actions to perform in case of failure is the way to go.

sunix:
Also, in most of the case, Github users would get through `https`, not `ssh` and they would be already logged in thanks to Github Oauth.
Ssh could be widely used in private git repos with onprem github/bitbucket/gitlab. So let's not hard coding this.

vitaliy-guliy:
[crw-ci-test]

vitaliy-guliy:
[crw-ci-test --rebuild]

vinokurig (approved):

vinokurig:
I confirm that the ssh notifications buttons work as expected.

plugins/workspace-plugin/src/theia-commands.ts
```diff
@@ -116,28 +119,106 @@ export class TheiaGitCloneCommand implements TheiaImportCommand {
...
+        if (tryAgain === TRY_AGAIN) {
+          // continue the loop to try again
+          continue;
+        }
+        // skip
+        return;
+      }
+
+      // pause will be removed after debugging this method
+      await this.pause(100);
```
RomanNikitenko, line 210: @vitaliy-guliy should we remove `pause` related code?

vitaliy-guliy: [PR](https://github.com/eclipse/che-theia/pull/1006) removing the delay.
Thanks

PR #963, opened 2021-01-12
Implement authority plugin API

vitaliy-guliy:
Signed-off-by: Vitaliy Gulyy <vgulyy@redhat.com>

### What does this PR do?
- Introduces authority plugin API to be able the plugin to get the an array of all available CA certificates.

Prevents duplicating code
 - https://github.com/eclipse/che-theia/blob/e165f775c878f87a124d1ecf1b2bdad5c1ed4bc9/plugins/task-plugin/src/machine/websocket.ts#L197-L214

 - https://github.com/eclipse/che-theia/blob/e165f775c878f87a124d1ecf1b2bdad5c1ed4bc9/extensions/eclipse-che-theia-remote-api/src/node/che-server-certificate-service-impl.ts#L21-L41

 - https://github.com/eclipse/che-theia/blob/eb34d11ca58b21279169acc320ed1d9bb2739a72/plugins/workspace-plugin/src/ca-cert.ts#L19-L34

### How to test this PR?

See https://github.com/eclipse/che-theia/pull/960

### PR Checklist

[As the author of this Pull Request I made sure that:](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#pull-request-template-and-its-checklist)

- [ ] [The Eclipse Contributor Agreement is valid](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-eclipse-contributor-agreement-is-valid)
- [ ] [Code produced is complete](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-produced-is-complete)
- [ ] [Code builds without errors](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-builds-without-errors)
- [ ] [Tests are covering the bugfix](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#tests-are-covering-the-bugfix)
- [ ] [The repository devfile is up to date and works](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-repository-devfile-is-up-to-date-and-works)
- [ ] [Sections `What issues does this PR fix or reference` and `How to test this PR` completed](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#sections-what-issues-does-this-pr-fix-or-reference-and-how-to-test-this-pr-completed)
- [ ] [Relevant user documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
[...]

PR #1014, opened 2021-03-04
Apply default certificate when downloading ZIP projects

vitaliy-guliy:

### What does this PR do?
Fixes downloading of zipped projects when creating a workspace.
Now the default system certificate is used when downloading zip with curl.

https://github.com/eclipse/che/issues/19120

### How to test this PR?

Devfile
```
---
apiVersion: 1.0.0
metadata:
  name: test-import-zip
projects:
  -
    name: theia-master
    source:
      type: zip
      location: "https://github.com/eclipse-theia/theia/archive/master.zip"
components:
  - type: cheEditor
    alias: che-theia
    reference: https://raw.githubusercontent.com/chepullreq4/pr-check-files/master/che-theia/pr-1014/che_theia_meta.yaml
    memoryLimit: 1G
```

### PR Checklist

### Reviewers

### Happy Path Channel

HAPPY_PATH_CHANNEL=stable

### DIFF
--- a/plugins/workspace-plugin/src/ca-cert.ts
+++ b/plugins/workspace-plugin/src/ca-cert.ts
@@ -16,8 +16,36 @@ const SS_CRT_PATH = '/tmp/che/secret/ca.crt';
 
 const CA_BUNDLE_PATH = '/tmp/ca-bundle.crt';
 
+/**
+ * Possible locations of default system certificates.
+ */
+const systemCerts = [
+  // Debian, Ubuntu, Gentoo
+  '/etc/ssl/certs/ca-certificates.crt',
+  // Fedora, RHEL 6
+  '/etc/pki/tls/certs/ca-bundle.crt',
+  // OpenSUSE
+  '/etc/ssl/ca-bundle.pem',
+  // OpenELEC
+  '/etc/pki/tls/cacert.pem',
+  // CentOS, RHEL 7
+  '/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem',
+  // Alpine
+  '/etc/ssl/cert.pem',
+];
+
 export const getCertificate = new Promise<string | undefined>(async resolve => {
   const certificates: Buffer[] = [];
+
+  // Look for default certificate.
+  // Stop iterating when finding one.
+  for (const cert of systemCerts) {
+    if (await fs.pathExists(cert)) {
+      // read the certificate
+      certificates.push(await fs.readFile(cert));
+      break;
+    }
+  }
 
   if (await fs.pathExists(SS_CRT_PATH)) {
     certificates.push(await fs.readFile(SS_CRT_PATH));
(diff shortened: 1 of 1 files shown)


plugins/workspace-plugin/src/ca-cert.ts
```diff
@@ -16,9 +16,37 @@ const SS_CRT_PATH = '/tmp/che/secret/ca.crt';
 
 const CA_BUNDLE_PATH = '/tmp/ca-bundle.crt';
 
+/**
+ * Possible locations of default system certificates.
```
azatsarynnyy, line 20: These are possible locations in general, but not in the case of Che/CRW.
As we run Che on Alpine and CRW on UBI only.

vitaliy-guliy: What's your proposal?
Remove all except the first and the last?

vitaliy-guliy: I would like to keep all of those file to be sure it is working.
If you want, I can remove the comments, to make it less confusing.

vitaliy-guliy: So, simplifying the list and leave only for Alpine and for RHEL/Fedora (as I understand UBI is RHEL)
But in fact, the default certificate has the same location for RHEL and Fedora.

vitaliy-guliy: BTW, do you remember our discussion about having a common code for getting a certificate?
This will be an improvement in my next PR.

azatsarynnyy (approved):

sunix:
Would that be possible to use the env variable NODE_EXTRA_CA_CERTS with the right certificates and see if it works ? if it does, it may fix this issue as well:
https://github.com/eclipse/che/issues/19168#issuecomment-789531114

vitaliy-guliy:
> Would that be possible to use the env variable NODE_EXTRA_CA_CERTS (in theia editor meta.yaml) with the right certificates path and see if it works ? if it does, it may fix this issue as well:
> [eclipse/che#19168 (comment)](https://github.com/eclipse/che/issues/19168#issuecomment-789531114)

It's a bit different.
The problem with downloading of ZIP packages was because `curl` was not using default certificates. Default certificate (it's a bundle of certificates) is necessary for https://github..., https://someotherserver, and for other server, that are normally working and for which we don't care which certificate is used. Each system has more that 200 certificates by default.

sunix:
for that use case, it is trying to download a file from the devfile registry, so it may be related. But if we are calling curl ... yeah it won't work. But why are we calling curl and not directly downloading the files ... :/

vitaliy-guliy:
> for that use case, it is trying to download a file from the devfile registry, so it may be related. But if we are calling curl ... yeah it won't work. But why are we calling curl and not directly downloading the files ... :/

@benoitf some time ago proposed to rework it a bit on using `axios` to download zips. I'm making a note about that improvement, next to https://github.com/eclipse/che-theia/pull/1014#discussion_r587587610

benoitf:
Hello, is the fix should be part of https://github.com/eclipse/che-theia/pull/963 as well ?

sunix:
> Hello, is the fix should be part of #963 as well ?

To me, #963 should not be done in Che theia, but editor agnostic.

PR #1019, opened 2021-03-05
fix(task-plugin): task plugin should not have to connect to an extern…

sunix:
…al route and should avoid dealing with https, tls issues.

Signed-off-by: Sun Tan <sutan@redhat.com>

### What does this PR do?
Replacing connection to machine-exec through external route with a connection to localhost. Should work as these are all running in the same pod. Also fixing certificate issues, simplifying code, avoid any other issues with proxy and certificates.

https://github.com/eclipse/che/issues/18957

### How to test this PR?
Make sure that che commands are working with the provided maven or nodejs generated devfile.

(optional) Setup https://che-incubator.github.io/2021/02/01/@mario.loriedo-using-mkcert-to-locally-trust-eclipse-che-tls-certificates-ffaafe76e5d0.html, make sure to have the problem described in https://github.com/eclipse/che/issues/18957. Then run the maven or nodejs devfile provided and make sure that tasks are working.

### PR Checklist

- [x] [The repository devfile is up to date and works](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-repository-devfile-is-up-to-date-and-works)

### Reviewers

### Happy Path Channel

HAPPY_PATH_CHANNEL=stable

### DIFF
--- a/plugins/task-plugin/src/che-workspace-client.ts
+++ b/plugins/task-plugin/src/che-workspace-client.ts
@@ -109,7 +109,10 @@ export class CheWorkspaceClient {
     if (!machineExecServer) {
       throw new Error(`No server with type ${TERMINAL_SERVER_TYPE} found.`);
     }
-    return machineExecServer.url!;
+    if (!machineExecServer.attributes!.port) {
+      throw new Error('No machine-exec-server attributes.port found');
+    }
+    return `ws://127.0.0.1:${machineExecServer.attributes!.port}`;
   }
 
   protected async getMachineExecServer(): Promise<cheApi.workspace.Server | undefined> {
--- a/plugins/task-plugin/src/machine/websocket.ts
+++ b/plugins/task-plugin/src/machine/websocket.ts
@@ -9,17 +9,8 @@
  ***********************************************************************/
 
 import * as WS from 'ws';
-import * as fs from 'fs';
-import * as http from 'http';
-import * as https from 'https';
-import * as path from 'path';
-import * as tunnel from 'tunnel';
-import * as url from 'url';
 
 import { ConsoleLogger, IWebSocket, Logger, MessageConnection, createWebSocketConnection } from 'vscode-ws-jsonrpc';
-
-const SS_CRT_PATH = '/tmp/che/secret/ca.crt';
-const PUBLIC_CRT_PATH = '/public-certs';
 
 /** Websocket wrapper allows to reconnect in case of failures */
 export class ReconnectingWebSocket {
@@ -46,7 +37,7 @@ export class ReconnectingWebSocket {
 
   /** Open the websocket. If error, try to reconnect. */
   open(): void {
-    this.ws = new WS(this.url, this.getOptions());
+    this.ws = new WS(this.url);
 
     this.ws.on('open', () => {
       this.onOpen(this.url);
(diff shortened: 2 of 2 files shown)


sunix:
@amisevsk it may break devworkspace operator as we should not use tls anymore https://github.com/eclipse/che-machine-exec/pull/107

plugins/task-plugin/src/che-workspace-client.ts
```diff
@@ -109,7 +109,8 @@ export class CheWorkspaceClient {
     if (!machineExecServer) {
       throw new Error(`No server with type ${TERMINAL_SERVER_TYPE} found.`);
     }
-    return machineExecServer.url!;
+    const machineExecPort = machineExecServer.attributes!.port || '4444';
```
benoitf, line 112: I think it should not use hardcoded 4444 if the port is not there, just goes into error/invalid state

sunix: if for some reasons it's not define in the workspace config, it is worth trying. At worst case, it would fail connecting.

sunix: I can throw an error if you prefer, if port is undefined

benoitf: yes throw an error
as for example with devworkspaces attribute is targetPort and we may change the port value

sunix: @benoitf done throwing an error

benoitf (approved):

amisevsk:
> it may break devworkspace operator as we should not use tls anymore eclipse/che-machine-exec#107

I'm not sure about this -- serving content over TLS is required in some cases for machine-exec (though my requirements come from the Web Terminal Operator). It makes sense to talk to machine-exec on localhost if possible (for a multitude of reasons) though.

cc: @sleshchenko

plugins/task-plugin/src/che-workspace-client.ts
```diff
@@ -109,7 +109,10 @@ export class CheWorkspaceClient {
     if (!machineExecServer) {
       throw new Error(`No server with type ${TERMINAL_SERVER_TYPE} found.`);
     }
-    return machineExecServer.url!;
+    if (!machineExecServer.attributes!.port) {
+      throw new Error('No machine-exec-server attributes.port found');
+    }
+    return `ws://127.0.0.1:${machineExecServer.attributes!.port}`;
```
sleshchenko, line 115: I wonder if terminal does not have the same issues you faced with tasks.

sleshchenko: Ah, after taking a deeper look I think terminal does not have such issue because it's a browser (not theia backend) who does certificates validation.

And I see some benefit in using localhost for tasks, but I don't see it as a solution for https://che-incubator.github.io/2021/02/01/@mario.loriedo-using-mkcert-to-locally-trust-eclipse-che-tls-certificates-ffaafe76e5d0.html
I assume that the issues caused by fact that Theia Backend needs CA certificate. See https://github.com/eclipse/che-theia/pull/1019/files#diff-fdc678251442b2044dd33d56c0a5ff4498f17c42b005cd63a696eb46f5703c81L129
But there may be some bug and end-certificate is propagated to Theia backend as well.
https://github.com/eclipse/che-operator/blob/master/pkg/deploy/server/deployment_che.go#L63

If you're able to share what's is propagate inside Che Theia - we would be able to check the theory above.

sleshchenko (approved):
I don't see any downsides in using localhost, while we run machine exec and theia in one pod, so +1 to merge it

But I think the way how https://che-incubator.github.io/2021/02/01/@mario.loriedo-using-mkcert-to-locally-trust-eclipse-che-tls-certificates-ffaafe76e5d0.html should be investigated more and probably fixed, to avoid confusion and misusing custom certificates.

sunix:
@sleshchenko I think it should be clarify in our advanced documentation:
- how the certificates should be configured so it works as expected
- how these certificates should be consumed (by a plugin, an extension or a che editor)

Does it make sense to create a dedicated issue for that ?

RomanNikitenko (approved):

PR #1036, opened 2021-03-17
Don't use cached url in the local storage to connect to the terminal server

AndrienkoAleksandr:
### What does this PR do?
Don't use cached url in the local storage to connect to the terminal server

### Referenced issues:
https://github.com/eclipse/che/issues/16593
https://github.com/eclipse/che/issues/19201

### How to test this PR?
On the minikube:
  - start any workspace
  - open new terminal
  - stop workspace
  - start workspace again
Expected result: che-theia should restore previously opened terminal.
Fixed bug: che-theia opened empty terminal and refresh browser page doesn't help. Helps only clean up local storage of the browser page...

### PR Checklist

### Reviewers

### Happy Path Channel

HAPPY_PATH_CHANNEL=stable

### DIFF
--- a/extensions/eclipse-che-theia-terminal/src/browser/server-definition/terminal-proxy-creator.ts
+++ b/extensions/eclipse-che-theia-terminal/src/browser/server-definition/terminal-proxy-creator.ts
@@ -43,4 +43,8 @@ export class TerminalProxyCreator {
     }
     return this.remoteTermServer;
   }
+
+  getApiEndPointUrl(): URI | undefined {
+    return this.apiEndPoint;
+  }
 }
--- a/extensions/eclipse-che-theia-terminal/src/browser/terminal-widget/remote-terminal-widget.ts
+++ b/extensions/eclipse-che-theia-terminal/src/browser/terminal-widget/remote-terminal-widget.ts
@@ -35,7 +35,6 @@ export interface RemoteTerminalWidgetOptions extends Partial<TerminalWidgetOptio
   machineName: string;
   workspaceId: string;
   closeWidgetOnExitOrError: boolean;
-  endpoint: string;
 }
 
 export interface RemoteTerminalWidgetFactoryOptions extends Partial<TerminalWidgetOptions> {
@@ -227,7 +226,9 @@ export class RemoteTerminalWidget extends TerminalWidgetImpl {
       this.resolveRemoteConnection();
       return Promise.resolve();
     }
-    this.socket = this.createWebSocket(id.toString());
+
+    const termProxyCreator = await this.termProxyCreatorProvider();
+    this.socket = this.createWebSocket(id.toString(), termProxyCreator.getApiEndPointUrl()!);
 
     const sendListener = (data: string) => this.socket.send(data);
 
@@ -252,6 +253,7 @@ export class RemoteTerminalWidget extends TerminalWidgetImpl {
       if (onDataDisposeHandler) {
         onDataDisposeHandler.dispose();
       }
+      this.messageService.error(`Terminal failed to connect. ${err.message}`);
       return Promise.resolve();
     };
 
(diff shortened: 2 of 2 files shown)


azatsarynnyy (commented):
I can't reproduce the issue with che-theia:next.
I let other reviewers check it.

azatsarynnyy:
I believe it fixes the issue https://github.com/eclipse/che/issues/19201

RomanNikitenko:
I tried to reproduce the issue on 3 different instances and couldn't.
So, it would be nice to get it tested by users who can reproduce the problem without these changes.

@tsmaeder could you check it when you have a chance

tsmaeder:
@RomanNikitenko that we can't reproduce it does not mean this PR fixes the "no terminal" issue (https://github.com/eclipse/che/issues/19201). It might just be different timing. @AndrienkoAleksandr does your analysis confirm that the issue this PR fixes would also cause https://github.com/eclipse/che/issues/19201?

RomanNikitenko:
@tsmaeder 
To be sure that we fix a bug - we should be able to reproduce it before some changes and then check it after the provided changes - I don't see another way. That's why I asked you to check it - based on https://github.com/eclipse/che/issues/19124 I thought you quite often face with it...

tsmaeder:
@RomanNikitenko that's the problem: it happens less often these days. That's why I'm saying just that it doesn't happen does not mean the bug is gone. If we can explain the bug through what's in the code though, we should be able to derive a scenario to reliably reproduce it and demonstrate it's gone after fixing.

RomanNikitenko:
@AndrienkoAleksandr 
 If I understand correctly the current changes should fix the problem related to **restoring** a terminal after workspace restarting?
 So, the use case is:
 - a user has a running workspace
 - create a terminal, do something
 - stop the workspace
 - start the workspace
 As result - the terminal should be restored.
 
 But the current changes doesn't fix the problem for a **new terminal creation** .
 So, the use case is:
 - a user has a running workspace
 - stop the workspace
 - start the workspace
 - create a new terminal
 The terminal should work correctly.
 
 Just for my understanding - the changes fix the first use case and it's not about the second use case. As for the second use case we don't store anything to the local storage.

AndrienkoAleksandr:
Hello, I recorded video about bug reproduction: https://youtu.be/YSW-r35XMYI

@RomanNikitenko  About current fix: after restart workspase should be fixed bug with "restore" previosly created terminals("restore" from point of view UI widgets. Under the hood actually these widgets, will be connected to the new execs, because after restart workspace previously execs were killed by k8s api). And yes, if user will create new terminals in this restored workspace, they should work fine too, not only restored terminals.

@azatsarynnyy I agree that should be fix for https://github.com/eclipse/che/issues/19201 , but I'm not sure about https://github.com/eclipse/che/issues/19124 .

Also I notify that I did this fix actually in the frame of the issue: https://github.com/eclipse/che/issues/16593 ...

RomanNikitenko:
>  yes, if user will create new terminals in this restored workspace, they should work fine too, not only restored terminals.

I thought local storage has influence on restoring terminal process only.
AFAIK a new terminal creation doesn't use cached options in local storage - it gets a new value using API - isn't it?
So should be no influence of local storage on a new terminal creation.

RomanNikitenko:
I had a meeting with @AndrienkoAleksandr, within the meeting we found the reliable way to reproduce the issue.
Steps to reproduce:
- run a workspace
- create a terminal
- wait few seconds for storing the layout
- restart the workspace
The terminal is restarted, but doesn't work.
Try to create a new terminal - id doesn't work. 

Using the steps I'm able to reproduce the issue permanently on minikube.

RomanNikitenko (approved):
Tested using the steps from https://github.com/eclipse/che-theia/pull/1036#issuecomment-801978227

**Before:**

terminal_before

**After:**

terminal_after

The changes fix the problem for me, thank you, Oleksandr!

PR #1090, opened 2021-04-22
Release images for ARM64

odidev:
Added **.github/workflows/build-publish-next-arm64.yml** to build and release the image for ARM46 platform.

Updated .github/workflows/build-publish-next.yml, build.include, .github/workflows/happy-path-tests.yaml, .github/workflows/pr.yml, dockerfiles/build.include and dockerfiles/theia/build.sh to build amd64 and arm64 images.


benoitf:
Hello @odidev and thanks for the contribution.

it seems we may not need to add a separate workflow for every build as the full idea with multi-arch is to submit as well multi-arch images on the OCI registry (instead of having some images with a suffix like `-arm64`, another `-ppc64le`, etc.

in some che repositories we're using as well the PLATFORMS file to add the list of platforms
like
https://github.com/eclipse-che/che-devfile-registry/blob/master/PLATFORMS

so probably if you could tweak the build.include to use buildx command if some flag is given

Also, with buildx, do you the time it takes to build the image for arm64 emulator ?

odidev:
Hi,
As per [your suggestion](https://github.com/eclipse-che/che-theia/pull/1090#issuecomment-824944362), I have amended my changes for the development of docker images for multiple platforms.

benoitf:
hello, I think that it'll be handle as part of the effort to build also Power and Z images like in https://github.com/eclipse-che/che-theia/pull/1127

PR #1114, opened 2021-05-21
Update ubi8 images to the latest nodejs and minimal base images from RHEL8.4

nickboldt:
This simple PR updates a bunch of base images used for building on RHEL.
Related to https://issues.redhat.com/browse/CRW-1844
image

### DIFF
--- a/dockerfiles/theia-dev/docker/ubi8/builder-image-from.dockerfile
+++ b/dockerfiles/theia-dev/docker/ubi8/builder-image-from.dockerfile
@@ -1,2 +1,2 @@
 # https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8/nodejs-12
-FROM registry.access.redhat.com/ubi8/nodejs-12:1-70.1614929178
+FROM registry.access.redhat.com/ubi8/nodejs-12:1-87
--- a/dockerfiles/theia-endpoint-runtime-binary/docker/ubi8/runtime-from.dockerfile
+++ b/dockerfiles/theia-endpoint-runtime-binary/docker/ubi8/runtime-from.dockerfile
@@ -1,2 +1,2 @@
 # https://access.redhat.com/containers/?tab=tags#/registry.access.redhat.com/ubi8-minimal
-FROM registry.access.redhat.com/ubi8-minimal:8.3-291 as runtime
+FROM registry.access.redhat.com/ubi8-minimal:8.4-200 as runtime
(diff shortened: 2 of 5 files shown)


benoitf:
hi, could you please not remove the PR template ?

nickboldt:
this is generated by script. script does not include template.

benoitf:
scripts should follow templates then

Also title says that it's bumping ubi8 minimal while in fact it's also bumping nodejs images

benoitf:
and I don't understand the meaning of `[base]` prefix.
it's not even a semantic commit

nickboldt:
it's updating the *base* image in a bunch of docker files. 

Sorta like dependabot, but it actually works. 

Are you intentionally being obtuse, or are you genuinely trying to understand how my automated process has been working for the last 2 years ?

benoitf (changes requested):
please format the body (human could understand what it does, not 2 changelogId) and title about what is really done by the PR

benoitf:
if it's automated, it should be issued by `che-bot` (also I think dependabot should handle these updates ? )

when we review weekly PRs, it's easier to see what is being automated if it's issued by a bot instead of done by a user just by looking at the user-id. Because all your PRs are automatic then it requires manual edit/review to analyze the content.


And the title is still not accurate (it's not about only ubi8-minimal)

nickboldt:
We don't yet have full automation - see https://github.com/eclipse/che/issues/18831 Dependabot for dockerfiles doesn't work. That's why I'm using my tried-and-true script, which I've been using for a couple years now to manage Che versions and CRW releases. 

This was a bug discovered in https://issues.redhat.com/browse/CRW-1844 -- if we had full automation, you'd have seen a che-bot PR instead of a nickboldt PR. Either way, the template would not have been used.

But blocking a simple base image update for pedantic/semantic reasons seems fruitless. Why are we fighting over this? it's a trivial change.

benoitf (approved):

benoitf:
> But blocking a simple base image update for pedantic/semantic reasons seems fruitless. Why are we fighting over this? it's a trivial change.

While you find it fruitless, Changelog can list all PRs being made so it needs to say what it does. 
So for example in that case it would be preferable to do like dependabot: one PR per image being updated
```
 -> Update ubi8 minimal image to ...
 -> Update nodejs ubi8 image to ...
```


Then as we review all PRs, having semantic information like chore, fix, feat, ease the review
If the title is self-descriptive as well (no ... at the end of the PR title)
And then automation should use bots token (so it's not performed by Nick but by che-bot) so it's trivial to see it's automated (not something a user decide)

PR #1161, opened 2021-07-07
If internal links are not specified, the external one should be used

AndrienkoAleksandr:
### What does this PR do?
- If internal links are not specified, the external one should be used
- Fix hang loading che-theia when internal network disabled
- Don't use deprecated CHE_API env variable.

https://github.com/eclipse/che/issues/18643

### Howc to test this PR?
1. Install Che using operator installer
2. Execute any workspace. Check that webview(welcome page), containers panel, list plugins works.
3. Disable internal network in the custom Che operator resource(set up to false useInternalClusterSVCNames field in the server section). Check that webview(welcome page), containers panel, list plugins works. che-theia should not hangs with loader.

### PR Checklist

### Reviewers

### Happy Path Channel

HAPPY_PATH_CHANNEL=stable


dmytro-ndp:
[crw-ci-test --rebuild]

extensions/eclipse-che-theia-remote-impl-che-server/src/node/che-server-remote-api-impl.ts
```diff
@@ -24,14 +24,13 @@ export class CheServerRemoteApiImpl {
...
+  private readonly baseAPIUrl: string | undefined;
   private readonly machineToken: string;
 
   constructor() {
-    if (process.env.CHE_API_INTERNAL === undefined) {
-      console.error('Unable to create Che API REST Client: "CHE_API_INTERNAL" is not set.');
-    } else {
-      this.baseAPIUrl = process.env.CHE_API_INTERNAL;
+    this.baseAPIUrl = process.env.CHE_API_INTERNAL || process.env.CHE_API_EXTERNAL;
+    if (this.baseAPIUrl === undefined || this.baseAPIUrl === '') {
```
azatsarynnyy, line 32: As I remember, it can be checked a bit simpler - `if (!this.baseAPIUrl) {`

extensions/eclipse-che-theia-remote-impl-che-server/src/node/che-server-remote-api-impl.ts
```diff
@@ -52,6 +51,6 @@ export class CheServerRemoteApiImpl {
   }
 
   getCheApiURI(): string {
-    return this.baseAPIUrl;
+    return this.baseAPIUrl!;
```
azatsarynnyy, line 54: Why forcing non-null assertion here?
Let's declare it in the return types - `getCheApiURI(): string | undefined {`

extensions/eclipse-che-theia-plugin-ext/src/node/che-plugin-service.ts
```diff
@@ -98,9 +98,9 @@ export class ChePluginServiceImpl implements ChePluginService {
 
     try {
       const workspaceSettings: WorkspaceSettings = await this.workspaceService.getWorkspaceSettings();
-      if (workspaceSettings && workspaceSettings[PLUGIN_REGISTRY_INTERNAL_URL]) {
-        const uri = workspaceSettings[PLUGIN_REGISTRY_INTERNAL_URL];
-        const publicUri = workspaceSettings[PLUGIN_REGISTRY_URL] || uri;
+      if (workspaceSettings) {
+        const uri = workspaceSettings[PLUGIN_REGISTRY_INTERNAL_URL] || workspaceSettings[PLUGIN_REGISTRY_URL];
+        const publicUri = workspaceSettings[PLUGIN_REGISTRY_URL];
```
azatsarynnyy, line 103: Let's reuse `publicUri`:
```js
const publicUri = workspaceSettings[PLUGIN_REGISTRY_URL];
const uri = workspaceSettings[PLUGIN_REGISTRY_INTERNAL_URL] || publicUri;
```

extensions/eclipse-che-theia-remote-impl-che-server/src/node/che-server-remote-api-impl.ts
AndrienkoAleksandr: I think `process.env.CHE_API_EXTERNAL` never should be null. So from api retrieving logic we always return value and undefined it something not expected. Maybe instead of console.error we should throw unexpected exception... And check that this error should be displayed on the che-theia ui. What do you think? Otherwise if we will change return type from `string` to `string | undefined` we have to apply undefined check in every usage place. And it will be not fair: actually code will be not working without che-api url. Container plugin, task plugin, factory plugin and so on can't work properly without interaction with che-server or dev controller analog.

extensions/eclipse-che-theia-remote-impl-che-server/src/node/che-server-remote-api-impl.ts
AndrienkoAleksandr: Yes, I agree.

extensions/eclipse-che-theia-plugin-ext/src/node/che-plugin-service.ts
AndrienkoAleksandr: Resolved

extensions/eclipse-che-theia-remote-impl-che-server/src/node/che-server-remote-api-impl.ts
azatsarynnyy: Actually, there's only one client of this method that already checks it and throws :slightly_smiling_face: 
https://github.com/eclipse-che/che-theia/blob/6fe4d5611400ec7a91ca959c692394d27d0091f0/extensions/eclipse-che-theia-remote-impl-che-server/src/node/che-server-factory-service-impl.ts#L22-L26
So, we can throw a bit earlier, in this constructor. But I'd say it's not mandatory for this PR.

azatsarynnyy (approved):

PR #1177, opened 2021-07-27
fix(workspace-plugins): fix custom workspace folders

zWingz:
fixed https://github.com/eclipse/che/issues/20196

### What does this PR do?

Fix the bug that all of projects would be added to custom workspace

### How to test this PR?

- create a che workspace
- create folder `a` and `b`
image
- create a custom theia workspace config file `custom.theia-workspace`
image

  
```json
{
   "folders": [
      {
         "path": "a"
      },
      {
         "path": "b"
      },
   ]
}
```
- press `F1` and select `Open Workspace Root`
image
- select `custom.theia-workspace`
image

Only the folder which defined on workspace config would be added to workspace 

image

image

### DIFF
--- a/plugins/workspace-plugin/src/workspace-projects-manager.ts
+++ b/plugins/workspace-plugin/src/workspace-projects-manager.ts
@@ -19,6 +19,12 @@ import { TheiaImportCommand, buildProjectImportCommand } from './theia-commands'
 
 import { WorkspaceFolderUpdaterImpl } from './workspace-folder-updater';
 
+/**
+ * che default workspaces name
+ * see extensions/eclipse-che-theia-workspace/src/node/che-workspace-server.ts
+ */
+const CHE_DEFAULT_WORKSPACE_ROOT_NAME = 'che.theia-workspace';
+
 const onDidCloneSourcesEmitter = new theia.EventEmitter<void>();
 
 export const onDidCloneSources = onDidCloneSourcesEmitter.event;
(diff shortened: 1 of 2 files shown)


plugins/workspace-plugin/src/workspace-projects-manager.ts
```diff
@@ -60,12 +66,22 @@ export class WorkspaceProjectsManager {
...
+    const isCustomWorkspaceRoot =
+      workspaceFile && path.basename(workspaceFile.path) !== CHE_DEFAULT_WORKSPACE_ROOT_NAME;
     if (isMultiRoot) {
       // Backward compatibility for single-root workspaces
       // we need it to support workspaces which were created before switching multi-root mode to ON by default
       for (const project of projects) {
         const projectPath = this.getProjectPath(project);
+        // All projects should be added to the default workspace
+        // But only the project which defined on workspace config should be added to current workspace
+        // see issues https://github.com/eclipse/che/issues/20196
```
svor, line 79: I don't think that we need the information about issue in the code

plugins/workspace-plugin/src/workspace-projects-manager.ts
```diff
@@ -60,12 +66,22 @@ export class WorkspaceProjectsManager {
...
+        const shouldAddedToWorkspaced =
```
svor, line 80: should it be `shouldAddedToWorkspace` ?

svor (commented):
@zWingz could you please add some tests related to your changes

svor:
@RomanNikitenko could you take a look please

zWingz:
@svor I have fixed code style and added tests for it

RomanNikitenko:
[crw-ci-test --rebuild]

svor (approved):
LGTM

screenshot-nimbus-capture-2021 07 28-18_39_37

RomanNikitenko (approved):
Works well for me, thank you for your contribution!

I noticed that `.theia/launch.json` is added to a folder even for a custom workspace (if a devfile contains debug configs).
I don't know if it's expected behaviour for your use case...
 
custom_workspace

zWingz:
It is an unexpected behavior，I have never used the debug config

But the launch.json will be added to a folder when I press F5

I don't know if they're connected

RomanNikitenko:
Exporting launch configs logic is placed [here](https://github.com/eclipse-che/che-theia/blob/6fe4d5611400ec7a91ca959c692394d27d0091f0/plugins/task-plugin/src/export/launch-configs-exporter.ts#L37). 
Maybe it makes sense to export configs only for `che` workspace, not custom...

RomanNikitenko:
> I have never used the debug config

I think you don't have such problem for your workspaces if your devfiles don't contain debug configurations

